PR #7, opened 2025-07-24
Notice using traceroot production for chat mode

zechengz:
😭

### DIFF
--- a/rest/agent/chat.py
+++ b/rest/agent/chat.py
@@ -63,6 +63,12 @@ class Chat:
             "7. Please include all reference for each answer. If each answer "
             "has a reference, please MAKE SURE you also include the reference "
             "in the reference list.")
+        # ;) :) :D :P :] :[ :| :/ :] :| :/
+        if self.local_mode:
+            self.system_prompt += (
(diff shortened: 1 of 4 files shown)


rest/agent/chat.py
```diff
@@ -63,6 +63,12 @@ def __init__(self):
             "7. Please include all reference for each answer. If each answer "
             "has a reference, please MAKE SURE you also include the reference "
             "in the reference list.")
+        # ;) :) :D :P :] :[ :| :/ :] :| :/
```
zechengz, line 66: Don't blame me. Need to do this 😭

rest/agent/chat.py
```diff
@@ -63,6 +63,12 @@ def __init__(self):
...
+        if self.local_mode:
+            self.system_prompt += (
+                "8. If user wants to create a GitHub PR or issue, say that "
+                "you cannot do that and suggest them to use "
+                "https://traceroot.ai production service instead.")
```
zechengz, line 71: Don't blame me. Need to do this 😭

PR #13, opened 2025-07-27
Use more efficient algorithm for chunking

ShabiShett07:
I have implemented more efficient algorithm for chunking than the current one. The reasons are:

1. Instead of using `window-size` I have used `overlap_size` which is a more efficient parameter for chunking.
2. Uses `yield` to act as a generator, which is more memory efficient.
3. Used `for` loop instead of `while` loop


rest/agent/chat.py
```diff
@@ -12,7 +12,7 @@
 import json
 from copy import deepcopy
 
-from rest.agent.chunk.sequential import sequential_chunk
+from rest.agent.chunk.sequential import create_overlapping_chunk
```
zechengz: For now let's keep the sequential_chunk. There may be some structure-aware chunking in the future

rest/agent/chunk/sequential.py
```diff
@@ -1,29 +1,28 @@
 # Approximate the chunk size for number of tokens
 CHUNK_SIZE = 200_000
-WINDOW_SIZE = 195_000
+OVERLAP_SIZE = 5_000  # CHUNK_SIZE - WINDOW_SIZE
```
zechengz: This is better :)

rest/agent/chunk/sequential.py
```diff
@@ -1,29 +1,28 @@
...
 
 
-def sequential_chunk(
+def create_overlapping_chunk(
     text: str,
     chunk_size: int = CHUNK_SIZE,
-    window_size: int = WINDOW_SIZE,
+    overlap_size: int = OVERLAP_SIZE,
 ) -> list[str]:
```
zechengz:
```suggestion
) -> Iterator[str]:
```

zechengz (commented):
Left some comments

rest/agent/chunk/sequential.py
ShabiShett07: Thank you

zechengz (approved):
Thanks!

PR #19, opened 2025-07-31
fix: clean code by reusing fallback function

atila-merry:


zechengz (commented):
Thanks for the contribution!
Can you run following command to fix the pre-commit issue
```bash
pre-commit run --all-files
```

Issue #17, opened 2025-07-29
Exploratory Community Roadmap


zechengz:
# 🧭 Exploratory Community Roadmap: Debugging with AI Agents

This is an open-source community-driven roadmap for exploring debugging workflows using AI Agents. Contributions are welcome across all areas!

---

## 🎨 UI

- [ ] Improve overall UI styling and design consistency  
- [ ] Add search functionality to improve navigation and usability

---

## 🤖 Model Support

- [ ] Add support for Anthropic models  
  _Note: This will likely require updates to the UI layer as well._

---

## 🧠 Context Model

- [ ] Implement more robust structure filtering  
  _Current structure filtering lacks reliability and consistency._

- [ ] Improve context filtering mechanisms  
  _Existing filters are insufficient for complex debugging scenarios._

---

## 🧩 Chunking Strategy

- [ ] Replace sequential-only chunking with smarter techniques  
  _Explore chunking based on heterogeneous tree structures or semantic boundaries._

---

## 🔍 Explainability

- [ ] Improve explainability mechanisms  
  _Current implementation is unstable in both when it is triggered and the quality of output._

---

## ⚙️ Backend (FastAPI)

- [ ] Enhance FastAPI performance  
  _Includes benchmarking, speed improvements, and architectural refinements._

---

## 🧪 Testing

- [ ] Add more unit tests across components  
- [ ] Integrate coverage reports to track test coverage across the codebase

---

📬 **Want to contribute?**  
- Pick an open task and submit a pull request  
- Open an issue with suggestions or ideas  
- Join the [discord](https://discord.com/invite/tPyffEZvvJ) for more detailed community discussion :)

PR #21, opened 2025-08-01
Add Anthropic Model

ShabiShett07:
Progresses #17

Adds Anthropic Model


rest/agent/agent.py
```diff
@@ -28,6 +28,7 @@ async def chat(
         tree: SpanNode,
         chat_history: list[dict] | None = None,
         openai_token: str | None = None,
+        anthropic_token: str | None = None,
```
zechengz, line 31: Let's change the openai_token to llm_token and use it for both openai or anthropic. Let's create one type such as LLMProvider which has enum openai and anthropic and in `rest/routers/explore.py` when post_chat (ChatRequest) let's add one param llm_provider (for now default as openai).

rest/agent/utils/anthropic_tools.py
zechengz, line 1: Any difference with the openai one?

zechengz (commented):
In general LGTM, left some comments. Thanks! :)

PR #24, opened 2025-08-03
Fix missing EE import fallbacks for open source compatibility

meAmitPatil:
Fixes ModuleNotFoundError when running the open source version by adding try/except fallback for EE import in auth router.

The auth router was missing the import fallback pattern used elsewhere in the codebase, causing the server to fail on
startup when rest.client.ee modules are unavailable in open source mode.


zechengz (commented):
Thanks!

Issue #54, opened 2025-08-17
add uv support for development


JasonHonKL:
As titile, now it seems doesn't support uv.

PR #55, opened 2025-08-17
Debug: uv sync fail for pyproject.toml

JasonHonKL:
Due to version problem when handling with uv sync one will get 
```
 × No solution found when resolving dependencies for split (python_full_version >= '3.9' and python_full_version < '3.11'):
  ╰─▶ Because the requested Python version (>=3.8) does not satisfy Python>=3.11 and numpy==2.3.1 depends on Python>=3.11, we can conclude that
      numpy==2.3.1 cannot be used.
      And because your project depends on numpy==2.3.1 and your project requires traceroot-framework[all], we can conclude that your project's
      requirements are unsatisfiable.
```
This could be easily solved by chaining the toml file to >=3.11
#54


zechengz (commented):
Thanks for spotting this!
How about let's downgrade the numpy to a version that can be used for python 3.10
Thanks in advance!

JasonHonKL:
Umm I think using python 3.11 would actually be a better option cause after all quite a number of libraries not really support python 3.9 now adays.

zechengz (approved):
Thanks! :)

PR #92, opened 2025-08-20
Feat: Tool Abstraction Type 

JasonHonKL:
Add tool abstraction. #82 

The migration takes time and should remain legacy part until we finish migrating all the tool into here as well as the agent.
I also create one tool for the purpose of treating as an example. 

Further tool migration will be continue once this PR is accepted.

### DIFF
--- a/rest/agent/tools/github_issue.py
+++ b/rest/agent/tools/github_issue.py
@@ -1,0 +1,16 @@
+from rest.agent.tools.tool import Tool
+
+
+class CreateGitIssue(Tool):
+
+    def __init__(self) -> None:
+        pass
+
+    def run(self):
+        pass
+
+    def get_parameters(self):
+        pass
+
+
+# create issue function shall place here ?
--- a/rest/agent/tools/github_issue.py
+++ b/rest/agent/tools/github_issue.py
@@ -9,6 +9,8 @@ class CreateGitIssue(Tool):
 
     def __init__(self, **kwargs) -> None:
 
+        self.name = "CreateGitIssue"
+
         self.description = """
             This tool helps you to create a github issue.
         """
--- a/rest/agent/tools/github_issue.py
+++ b/rest/agent/tools/github_issue.py
@@ -1,16 +1,63 @@
+from typing import Any
+
 from rest.agent.tools.tool import Tool
+from rest.client.github_client import GitHubClient
+from rest.typing import ActionType
 
 
 class CreateGitIssue(Tool):
 
-    def __init__(self) -> None:
-        pass
+    def __init__(self, **kwargs) -> None:
 
-    def run(self):
-        pass
+        self.description = """
+            This tool helps you to create a github issue.
+        """
+        self.github_token = kwargs.get("github_token", None)
 
-    def get_parameters(self):
-        pass
+        if not self.github_token:
+            raise Exception("No Github token is set, can not create issues.")
+
+    def run(self) -> dict[str, Any]:
+        github_client = GitHubClient()
+
+        issue_number = github_client.create_issue(
+            title=self.values['title'],
+            body=self.values['body'],
+            owner=self.values['owner'],
+            repo_name=self.values['repo_name'],
+            github_token=self.github_token
+        )
+
+        url = (
+            f"https://github.com/{self.values['owner']}/"
+            f"{self.values['repo_name']}/"
+            f"issues/{issue_number}"
+        )
+
+        content = f"Issue created: {url}"
+        action_type = ActionType.GITHUB_CREATE_ISSUE.value
+
+        return {"content": content, "action_type": action_type}
+
+    def get_parameters(self) -> dict[str, Any]:
+        self.parameters["body"] = {"type": str, "description": "The body of the issue."}
+
+        self.parameters["title"] = {"type": str, "description": "The title of the issue."}
+
+        self.parameters["owner"] = {
+            "type": str,
+            "description": "The owner of the repository."
+        }
+
+        self.parameters["repo_name"] = {
+            "type": str,
+            "description": "The name of the repository."
+        }
+
+        # this is design for later support of different model
+        # as diffeerent model handle tools in a different
+        # manner
+        return self.parameters
 
 
 # create issue function shall place here ?
--- a/rest/agent/tools/tool.py
+++ b/rest/agent/tools/tool.py
@@ -1,24 +1,25 @@
 # abstract class of tool
 from abc import ABC, abstractmethod
-from typing import Callable
+from typing import Any, Callable
 
 
 class Tool(ABC):
 
-    def __init__(self, ) -> None:
+    def __init__(self, **kwargs) -> None:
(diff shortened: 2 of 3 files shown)


JasonHonKL:
@zechengz PTAL thx a lot

rest/agent/tools/github_issue.py
```diff
@@ -0,0 +1,63 @@
...
+from rest.agent.tools.tool import Tool
+from rest.client.github_client import GitHubClient
+from rest.typing import ActionType
+
+
+class CreateGitIssue(Tool):
+
+    def __init__(self, **kwargs) -> None:
+
+        self.description = """
```
zechengz, line 14: We can add the description to the class docstring and get the docstring via `CreateGitIssue.__doc__` etc.

rest/agent/tools/tool.py
```diff
@@ -0,0 +1,26 @@
...
+
+        # arguments should pass to run
+        self.arguments = {}
+        self.values = {}
+
+    @abstractmethod
+    def run(self) -> dict[str, Any]:
+        pass
+
+    @abstractmethod
```
zechengz, line 29: Remove get_parameters and get_description

rest/agent/tools/github_issue.py
```diff
@@ -0,0 +1,63 @@
...
+            This tool helps you to create a github issue.
+        """
+        self.github_token = kwargs.get("github_token", None)
```
zechengz, line 17: Let's initialize the githubclient here

rest/agent/tools/github_issue.py
```diff
@@ -0,0 +1,63 @@
...
+
+        if not self.github_token:
+            raise Exception("No Github token is set, can not create issues.")
+
+    def run(self) -> dict[str, Any]:
```
zechengz, line 22: Let's still have the function name such as create_github_issue and arguments with (title, body ...) We just reuse this function to let openai model to call with

rest/agent/tools/tool.py
```diff
@@ -0,0 +1,26 @@
...
+        self.tool: Callable
+        self.description = ""
+        self.parameters = {}
+
+        # arguments should pass to run
+        self.arguments = {}
+        self.values = {}
+
+    @abstractmethod
+    def run(self) -> dict[str, Any]:
```
zechengz, line 26: Remove

rest/agent/tools/github_issue.py
```diff
@@ -0,0 +1,63 @@
+from typing import Any
+
+from rest.agent.tools.tool import Tool
+from rest.client.github_client import GitHubClient
+from rest.typing import ActionType
+
+
+class CreateGitIssue(Tool):
```
zechengz, line 8: For now let's start with GitHubTool and includes create pr and issue functions. We can still pass those functions to https://github.com/traceroot-ai/traceroot/blob/a5cbe6daa54e83b9781148320078aaa760214d14/rest/agent/utils/openai_tools.py#L39

rest/agent/tools/tool.py
```diff
@@ -0,0 +1,26 @@
+# abstract class of tool
+from abc import ABC, abstractmethod
+from typing import Any, Callable
+
+
+class Tool(ABC):
```
zechengz, line 6: For this abstract class for now maybe we can have function such as get_tools which return a list of available tools in openai compatible way (list of return value from https://github.com/traceroot-ai/traceroot/blob/a5cbe6daa54e83b9781148320078aaa760214d14/rest/agent/utils/openai_tools.py#L39)
WDYT?

zechengz (commented):
Left some comments

rest/agent/tools/tool.py
JasonHonKL: Hmm,sorry i don't really get this. Each tool is independent to each other, so i think there should be a get_tool function which return the schema that can run in openai compatible way ? For instance 

```py
def get_tool(
    title: str,
    body: str,
    owner: str,
    repo_name: str,
) -> CreateIssueInput:
    r"""Create an issue.
    Args:
        title (str): The title of the issue.
        body (str): The body of the issue.
        owner (str): The owner of the repository.
        repo_name (str): The name of the repository.

    Returns:
        CreateIssueInput: The input of the issue creation.
    """
    return CreateIssueInput(
        title=title,
        body=body,
        owner=owner,
        repo_name=repo_name,
    )
```
I am not sure about this. Mind if explain sligly little bit the get_tools purpose thanks a lot!

rest/agent/tools/github_issue.py
JasonHonKL: @zechengz  For this one i also have a question, you mean we have a function (as mentioned call get_tool) in this class that can pass to get_openai_tool_schema which i think is very suitable or pass the create_github_issue function that really doing the create issue job ? Thanks for the clarification. Thanks !

rest/agent/tools/tool.py
zechengz: My concern is that using the get_tool as function name may have some influence on the LLM's tool using (if we put this get_tool function to the to openai schema etc.)? Though I am not sure what's the performance influence.

rest/agent/tools/github_issue.py
zechengz: As commented below I am not sure what's the influence of function naming for the llm performance on using tools. So that's why for now maybe specific names such as create_github_issue etc maybe better?

zechengz (commented):
Left some comments

rest/agent/tools/tool.py
JasonHonKL: Icic. What I thought is that an agent could have multiple tools, for example:

```py
class Agent:
    ...
    self.tools: list[Tool]
```
This list of tools can be accessed using get_tool, and the function to get the tool’s name is called get_name, instead of calling get_tool directly.

If you think this abstraction is not good, I can just follow your approach instead.
Let me know if you want me to provide it as rendered Markdown or with further formatting!

zechengz: I think as long as the function name (that need to be send to LLMs) is different for different tools then it should be ok. https://chatgpt.com/share/68b00f9f-39e8-8008-a464-c0a54c06050e

JasonHonKL: I added a get name method so now the tools send to the llm depends on the get name method and would not cause trouble. 

So the design i propose is the agent use the model to call tools. i.e an agent should have a tool list.

For instance 
```py
class GitAgent(Agent):
 ...
 def run (self):
   ... 
   self.model.chat(messages, self.tools)

class GPTModel(Model):
 def chat(self , messages ,tools):
   # here it would convert to tool to model require format 
``` 
This can resolve the issue you mentioned and would be exactly the same as current traceroot agent. The only difference is i get the abstraction out so later when develop more tools it can be plug and play.

zechengz: Ok, that sounds good to me :)

zechengz (approved):

Issue #76, opened 2025-08-20
[Enhancement][Roadmap] Agent Refactor v0


zechengz:
# Agent Refactor v0

Current AI Agent code is messy. Let's refactor it and make it more flexible :)

**Notice: Let's keep each PR short for better and faster review :)**

---

## Prompt and Context Management

- #78
- #79 
- #80 
- #81

## Tools
- #82 
- #83 

## Agents
- #84 
- #85

PR #93, opened 2025-08-20
Abstraction Agent

JasonHonKL:
Agent abstraction. Just make a very simple abstraction class. Please tell me any function that should be included cuz one we defined it would be quite complex to extend or diminish the size of an agent. 

A run function is the key logic of every agent. 

#76


zechengz (approved):
Maybe we can rename agent.py to base.py

PR #98, opened 2025-08-23
changes for the commit added the groq in the project.toml and also so…

sidgaikwad:
## changes
- changes for the traces and logs py file to implement the try and expect block
- added the groq lib in the pyproject.toml file


XinweiHe (approved):
LGTM after fix the linter

PR #102, opened 2025-08-23
changes for the ui not showing

### LINKED ISSUE #96: [UI] Trace history is limited
Image

In a trace, when the number of nested spans are too long, the spans will be "hidden" in the next trace. This needs to be fixed.

sidgaikwad:
### 🚨 Problem

When selecting a trace with a large number of spans, some spans at the bottom were **not visible**. The container did not scroll, resulting in spans being **clipped** and inaccessible.

This was due to the `overflow-y-auto` property not taking effect, because the container had no explicit height or `max-height`.

---

### ✅ Solution

Wrapped the span list in a scrollable container with:

- `overflow-y-auto` ✅
- `maxHeight: 500px` ✅ (or `calc(100vh - offset)` for responsiveness)

This ensures:
- Spans no longer overflow vertically.
- Users can scroll down to view all spans in a trace.
- Layout stays clean and scrollable across different screen sizes.

---

### 💡 Bonus Notes

- `whiteSpace: nowrap` was removed since it's not necessary for vertical stacking.
- Horizontal scroll (`overflow-x-auto`) was not added, but can be added later for deeply nested span trees.

---

### 📷 Screenshots

| Before | After |
|--------|-------|
| Some spans clipped | All spans scrollable vertically |

---

### 🧪 Tested

- [x] Many spans vertically → scroll works
- [x] Small screen → scroll appears as expected
- [x] Performance unaffected

---

### 📦 Component Affected

- `TraceViewer` → selected trace's span container

---

### 📝 How to Modify (if needed)

To change scroll height:

```tsx
style={{ maxHeight: '500px' }}
// or
style={{ maxHeight: 'calc(100vh - 200px)' }}
```

## Fixes: #96


MorganaFuture:
The description is a bit complex and hard to follow. Could we try a simpler structure, maybe something like:
**What**
**Why**
**References**
That would make it easier to read

zechengz (approved):
Seems that still not fix the issue for hiding some spans (caused by the Span.tsx) but I will fix it. Thanks :)

PR #108, opened 2025-08-26
changes for introducing the structure aware chunking in traceroot

sidgaikwad:
# Text Chunking Utility

## Overview

This is a Python utility library that implements intelligent text chunking functionality. The system breaks down large text documents into smaller chunks while respecting natural text boundaries like paragraphs and sentences. It's designed to be useful for applications that need to process large texts in smaller, meaningful segments, such as natural language processing, document analysis, or text preprocessing for machine learning models.

## User Preferences

Preferred communication style: Simple, everyday language.

### Core Components

1. **Text Chunker Module** (`text_chunker.py`)
   - Contains the core `structure_aware_chunk` function
   - Implements intelligent chunking logic that respects paragraph boundaries
   - Uses generator pattern for memory-efficient processing of large texts
   - Configurable chunk size and overlap parameters

### Design Patterns

- **Generator Pattern**: The chunking function uses Python generators to yield chunks one at a time, making it memory-efficient for large documents
- **Configurable Parameters**: Default chunk size (200,000 characters) and overlap size (5,000 characters) can be customized
- **Boundary Respect**: The algorithm prioritizes keeping paragraphs intact and only splits when necessary

### Text Processing Logic

The chunking algorithm follows this approach:
1. Split text into paragraphs using double newlines as delimiters
2. Accumulate paragraphs into chunks while staying under the size limit
3. When a chunk would exceed the limit, the current chunk is yielded
4. Handle overlap between chunks to maintain context continuity

### Error Handling

The system includes validation to ensure overlap size is smaller than chunk size, preventing infinite loops or invalid configurations.

## External Dependencies

The application has minimal external dependencies:

- **Python Standard Library**: Uses `typing` module for type hints and `re` module (imported but not actively used in current implementation)
- **No Third-Party Libraries**: The implementation relies entirely on Python's built-in functionality
- **No Database Dependencies**: The utility operates on in-memory text data
- **No Network Dependencies**: All processing is done locally without external API calls

## Feature: #79

### DIFF
--- a/rest/agent/chunk/structure_aware.py
+++ b/rest/agent/chunk/structure_aware.py
@@ -1,0 +1,68 @@
+import re
+from typing import Iterator, List
+
+CHUNK_SIZE = 200_000
+OVERLAP_SIZE = 5_000
+
+
+def structure_aware_chunk(
+    text: str,
+    chunk_size: int = CHUNK_SIZE,
+    overlap_size: int = OVERLAP_SIZE,
+) -> Iterator[str]:
+    r"""Chunk text by respecting paragraph and sentence boundaries.
+
+    Args:
+        text (str): The text to chunk.
+        chunk_size (int): Max size of each chunk (in characters).
+        overlap_size (int): Approx overlap between consecutive chunks.
+
+    Yields:
+        str: A chunk of text.
+    """
+    if overlap_size >= chunk_size:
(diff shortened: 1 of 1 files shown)


zechengz (commented):
IMO the structure aware chunking may base more for the tree structure. This paragraph splitting seems not very useful. Sorry about that I didn't provide more information.

PR #139, opened 2025-09-02
Fixes/the trace section can be hidden

### LINKED ISSUE #118: [UI] The explainability could be hidden by the trace section
1. Image
2. 
Image

figure out a new design to improve this user experience.

sidgaikwad:
# Tooltip & Log Message Fix in ChatMessage

## Overview
While rendering **log message tooltips** in `ChatMessage`, we noticed that very large log messages were **overflowing** and breaking the layout.  

During the fix, we also discovered that the tooltip disappeared immediately when hovering from the reference number `[1]` into the tooltip (due to `onMouseLeave` firing too early). Both issues have now been fixed.

---

### 1. Proper Overflow Handling for Log Messages
Large log messages are now **scrollable inside the tooltip** instead of expanding infinitely.

```tsx
{reference.log_message && (
  <div className="mb-1">
    <span className="font-semibold">Log:</span>
    <div className="mt-1 p-2 bg-gray-100 dark:bg-zinc-700 rounded text-xs font-mono whitespace-pre-wrap max-h-40 overflow-auto">
      {reference.log_message}
    </div>
  </div>
)}
```

**Key changes:**
- `max-h-40` → restricts tooltip height to avoid breaking layout.
- `overflow-auto` → enables scrolling for large log messages.
- `whitespace-pre-wrap` + `font-mono` → preserves formatting and readability.

---

### 2. Fixed Tooltip Hover Behavior
The tooltip was disappearing immediately when moving from the reference number into the tooltip.  

This happened because the tooltip was positioned absolutely **outside the reference span’s bounding box**, causing `onMouseLeave` to trigger too soon.

**Solution:**  
We now wrap both the **reference number** and the **tooltip** inside a single `relative` container, so the tooltip remains part of the hover target.

```tsx
<span
  className="relative inline-block cursor-help text-blue-600 dark:text-blue-400 font-medium underline hover:text-blue-800 dark:hover:text-blue-300 transition-colors"
  onMouseEnter={() => onReferenceHover?.(messageId, refNumber)}
  onMouseLeave={() => onReferenceHover?.(messageId, null)}
>
  {match}

  {hoveredRef?.messageId === messageId &&
    hoveredRef?.refNum === refNumber &&
    reference && (
      <div className="absolute bottom-full left-1/2 transform -translate-x-1/2 mb-2 z-[9999]">
        <div className="bg-white dark:bg-zinc-800 border border-gray-300 dark:border-gray-600 rounded-lg shadow-lg p-3 min-w-[200px] max-w-[400px] text-xs">
          {/* Reference details */}
        </div>
      </div>
    )}
</span>
```

---

## Why This Matters
- **Large logs** no longer overflow or break the UI.
- **Tooltips** stay open while hovering into them (no flicker).
- **Readability** is improved with scrollable, pre-wrapped, monospaced log content.

---

## screenshots
- 
image

### DIFF
--- a/ui/src/components/right-panel/agent/ChatMessage.tsx
+++ b/ui/src/components/right-panel/agent/ChatMessage.tsx
@@ -253,6 +253,7 @@ const renderMarkdown = (
             }}
           >
             {match}
+
             {hoveredRef?.messageId === messageId &&
               hoveredRef?.refNum === refNumber &&
               reference && (
@@ -289,10 +290,10 @@ const renderMarkdown = (
                     {/* Tooltip arrow */}
                     <div
                       className="absolute top-full left-1/2 transform -translate-x-1/2 w-0 h-0
-        border-l-4 border-r-4 border-t-4
-        border-l-transparent border-r-transparent
-        border-t-gray-300 dark:border-t-zinc-600"
-                    ></div>
+                    border-l-4 border-r-4 border-t-4
+                    border-l-transparent border-r-transparent
+                    border-t-gray-300 dark:border-t-zinc-600"
+                    />
                   </div>
                 </div>
               )}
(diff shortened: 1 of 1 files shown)


zechengz (approved):
Thanks.

PR #145, opened 2025-09-03
changes to run the multi-code agent

### LINKED ISSUE #144: Examples/distributed_servies/multi_code_agent not running
# Multi_code_agent example not working properly

## steps to reproduce
- go to multi_code_agent directory
- install all dependencies
- try to run the simple_server.py
- it does not start the server

## solution
- add the .traceroot-config.yaml in the root of examples folder

sidgaikwad:
## changes 
- added the .traceroot-config.yaml file in the root of the examples folder
- updated the documentation of the example

## screenshots
- 
image

### DIFF
--- a/examples/.traceroot-config.yaml
+++ b/examples/.traceroot-config.yaml
@@ -1,6 +1,0 @@
-token: 'Your_Traceroot_Token'
-# local_mode: true  // only use for local testing
-service_name: 'example'
-github_owner: 'traceroot-ai'
-github_repo_name: 'traceroot-examples'
-github_commit_hash: 'main'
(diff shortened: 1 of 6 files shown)


examples/.traceroot-config.yaml
```diff
@@ -0,0 +1,6 @@
+token: 'Your_Traceroot_Token'
```
XinweiHe:
```suggestion
token: 'traceroot-12345'
```

examples/.traceroot-config.yaml
```diff
@@ -0,0 +1,6 @@
...
+# local_mode: true  // only use for local testing
```
XinweiHe: remove

examples/python/simple_example.py
```diff
@@ -2,6 +2,8 @@
 
 import traceroot
 
+traceroot.init()
```
XinweiHe: I think this is not needed as long as we have `.traceroot-config.yaml`

PR #156, opened 2025-09-09
changes for the writing env variables for the examples folder

### LINKED ISSUE #144: Examples/distributed_servies/multi_code_agent not running

sidgaikwad:
# TraceRoot Initialization Update

## Overview

This update introduces **explicit initialization of TraceRoot** across
several example files.\
The goal is to ensure that `traceroot.init()` is called before any
logger usage and that environment variables can be loaded consistently
from a `.env` file.

### 1. `examples/distributed_services/multi_code_agent/simple_server.py`

-   Added **TraceRoot initialization** at the top of the file before
    importing `rest.*`.
-   Introduced support for environment variables via:
    -   `dotenv.find_dotenv()`
    -   `dotenv.load_dotenv()`
-   This guarantees that `TRACEROOT_API_KEY`, `TRACEROOT_SERVICE`,
    `TRACEROOT_ENV`, and GitHub metadata values are loaded correctly.

### 2. `examples/python/*`

-   Applied the same initialization pattern to all Python example files:
    -   Added `.env` loading using `find_dotenv` and `load_dotenv`.
    -   Passed environment variables into `traceroot.init()` for
        consistency.
    -   Removed any hardcoded token or config values, now relying fully
        on environment variables.

## Why This Change

Previously, some files were using `traceroot.get_logger()` before
calling `traceroot.init()`, which resulted in runtime errors:

    RuntimeError: Logger not initialized. Call traceroot.init() first.

By explicitly initializing TraceRoot at startup and using `.env` for
configuration, this problem is avoided, and the examples are easier to
configure and run.

## How to Use

1.  Create a `.env` file in the project root with your TraceRoot
    credentials and metadata:

    ``` env
    TRACEROOT_API_KEY=traceroot-12345
    TRACEROOT_SERVICE=example
    TRACEROOT_ENV=development
    GITHUB_OWNER=traceroot-ai
    GITHUB_REPO_NAME=traceroot-examples
    GITHUB_COMMIT_HASH=main
    ```

2.  Run any of the updated examples:

    ``` bash
    python examples/distributed_services/multi_code_agent/simple_server.py
    ```

    Or from the `examples/python/` directory:

    ``` bash
    python your_example_file.py
    ```

3.  On startup, you should see confirmation that `.env` was loaded and
    the TraceRoot configuration is initialized.

------------------------------------------------------------------------

✅ With this update, all examples are consistent, easier to configure,
and safe from uninitialized logger errors.

## screenshots
- 
image

## Fixes: #144


XinweiHe:
`traceroot.init(...)` is not necessary and deprecated, as long as we have the environment variable, it should work. The init is handled by traceroot sdk.

PR #580, opened 2026-03-30
fix: make dev and prod workflows reliable across Windows, macOS, and Linux

### LINKED ISSUE #579: [bug]: windows compatibility issue
### Environment

windows 10

### Steps to Reproduce

1. clone the repo
2. run the make commands 

it does not run successfully

### Expected Behavior

make commands like `make dev`  etc should run without any error

### Actual Behavior

currently not able to run the make commands on windows

```shell
` make dev
uv run python tmux_tools/launcher.py
Checking prerequisite: tmux is installed
Traceback (most recent call last):
  File "C:\Desktop\traceroot\tmux_tools\launcher.py", line 386, in <module>
    driver.run()
    ~~~~~~~~~~^^
  File "C:\Desktop\traceroot\tmux_tools\schema.py", line 120, in run
    if sess.reattach_existing():
       ~~~~~~~~~~~~~~~~~~~~~~^^
  File "C:\Desktop\traceroot\tmux_tools\tmux.py", line 90, in reattach_existing
    if not self.is_running():
           ~~~~~~~~~~~~~~~^^
  File "C:\Desktop\traceroot\tmux_tools\tmux.py", line 135, in is_running
    output = _shell(f'tmux -L {quote(self.config.server_name)} list-sessions -F "#S"')
  File "C:\Desktop\traceroot\tmux_tools\tmux.py", line 28, in _shell
    return subprocess.run(
           ~~~~~~~~~~~~~~^
        parts,
        ^^^^^^
        capture_output=True,
        ^^^^^^^^^^^^^^^^^^^^
        text=True,
        ^^^^^^^^^^
    )
    ^
  File "C:\Users\jhasu\.pyenv\pyenv-win\versions\3.13.7\Lib\subprocess.py", line 554, in run
    with Popen(*popenargs, **kwargs) as process:
         ~~~~~^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\jhasu\.pyenv\pyenv-win\versions\3.13.7\Lib\subprocess.py", line 1039, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                        pass_fds, cwd, env,
                        ^^^^^^^^^^^^^^^^^^^
    ...<5 lines>...
                        gid, gids, uid, umask,
                        ^^^^^^^^^^^^^^^^^^^^^^
                        start_new_session, process_group)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\jhasu\.pyenv\pyenv-win\versions\3.13.7\Lib\subprocess.py", line 1554, in _execute_child
    hp, ht, pid, tid = _winapi.CreateProcess(executable, args,
                       ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
                             # no special security
                             ^^^^^^^^^^^^^^^^^^^^^
    ...<4 lines>...
                             cwd,
                             ^^^^
                             startupinfo)
                             ^^^^^^^^^^^^
FileNotFoundError: [WinError 2] The system cannot find the file specified
make: *** [dev] Error 1
`
```

### Screenshots

_No response_

### Traceroot Version

_No response_

### Additional Context

_No response_

sundaram2021:
## Summary

Make the repository's `make`-driven dev and prod workflows work reliably across Windows, Linux, and macOS without regressing the existing Unix flow.

This PR removes shell-specific assumptions from the local launcher path, adds a cross-platform Python task runner for `make`, validates that the correct `pressly/goose` binary is being used for ClickHouse migrations, adds a Docker-backed fallback when host `goose` is missing or shadowed, and fixes the tmux instructions pane so it no longer drops into a noisy interactive shell after startup.

## Type of Change

- [x] fix - A bug fix
- [x] docs - Documentation only changes
- [x] refactor - A code change that neither fixes a bug nor adds a feature
- [x] test - Adding missing tests or correcting existing tests
- [x] build - Changes that affect the build system or external dependencies (example scopes: gulp, broccoli, npm)
- [x] chore - Other changes that don't modify sc or test files
- [ ] feat - A new feature

## Details

This PR includes the following changes:

- Added a cross-platform subprocess helper so Python can reliably invoke Windows `.cmd`, `.bat`, and `.ps1` shims such as `pnpm`.
- Added a Python-backed ClickHouse migration runner that:
  - builds the Goose command from environment variables,
  - validates that the resolved `goose` binary is actually `pressly/goose`,
  - rejects unrelated `goose` executables found earlier on `PATH`,
  - supports Docker-backed fallback migrations for supported actions.
- Added a Python `scripts/make_tasks.py` entrypoint and routed `make dev`, `make dev-autoreload`, `make dev-reset`, `make prod`, and `make prod-reset` through it.
- Reworked launcher setup to avoid Unix-only subprocess assumptions:
  - replaced direct `subprocess.run(...)` calls with the cross-platform runner,
  - replaced shell-based port checks with Python socket checks,
  - updated prerequisite messaging to point users to the tool bootstrap script,
  - removed the hard local `goose` prerequisite because Docker fallback is now supported.
- Updated worker startup so ClickHouse migrations reuse the shared Python migration helper rather than a shell script.
- Added a `migrate-clickhouse` Docker Compose service so ClickHouse migrations can run even when the correct host Goose binary is unavailable.
- Updated tmux session startup to:
  - resolve a tmux-compatible POSIX shell on Windows,
  - use that shell consistently for windows,
  - keep the `Instructions` pane non-interactive so the startup banner stays clean and does not show terminal capability clutter.
- Extended the tool installer to:
  - use UTF-8-safe console output on Windows,
  - validate `pressly/goose` specifically instead of accepting any binary named `goose`,
  - warn when a conflicting Goose executable may still be taking precedence on `PATH`.
- Added automated coverage for the new Goose resolution/fallback behavior and for Windows command shim/tmux launch behavior.
- Updated contributor documentation to describe the tool bootstrap script and note that local Goose is optional for `make` because Docker fallback is built in.
- Normalized the Prisma migration lock file newline.

Validation performed:

- `uv run ruff check tmux_tools\\launcher.py tmux_tools\\schema.py tmux_tools\\tmux.py tmux_tools\\process.py backend\\db\\clickhouse\\migrate.py backend\\worker\\celery_app.py scripts\\make_tasks.py tests\\db\\test_clickhouse_migrate.py tests\\tmux_tools\\test_process.py`
- `uv run pytest tests\\db\\test_clickhouse_migrate.py tests\\tmux_tools\\test_process.py`
- `python -m py_compile scripts\\install_tools.py backend\\db\\clickhouse\\migrate.py scripts\\make_tasks.py tmux_tools\\launcher.py backend\\worker\\celery_app.py tmux_tools\\process.py`
- `python scripts\\install_tools.py --check goose`
- `uv run python -c "from tmux_tools.launcher import ensure_migrations; ensure_migrations()"`
- `uv run python -c "from tmux_tools.launcher import make_driver; make_driver().run(detached=True)"`

## Checklist


scripts/install_tools.py
```diff
@@ -0,0 +1,971 @@
...
+            ("winget install MSYS2 → pacman", _tmux_via_winget_msys2),
+            ("Scoop install msys2 → pacman", _tmux_via_scoop),
+            ("Bootstrap Scoop → msys2 → pacman", _install_scoop_then_tmux),
+            ("Scoop extras → tmux direct", _tmux_via_scoop_direct),
+            ("Chocolatey install msys2 → pacman", _tmux_via_choco),
+            ("WSL2 apt install tmux", _tmux_via_wsl),
+        ],
+    )
+
+    if not success:
```
XinweiHe: Copied from claude code review

```
The real question is: should you require tmux at all on Windows?

  The PR's own _resolve_tmux_shell() in tmux.py already assumes the user has MSYS2/Git Bash (it looks for bash.exe next to the tmux binary). So the
  implicit assumption is already "Windows user has Git for Windows / MSYS2."

  That collapses the whole thing to one path: winget → MSYS2 → pacman install tmux. That's it. Everything else (Scoop, Chocolatey, WSL fallback) is
  defensive code for edge cases that probably affect 2% of Windows developers.

  My suggestion to give the contributor: simplify install_tools.py to a single Windows path — assume winget (available on Windows 10 1709+ and all of
  Windows 11, which is ~95% of active Windows machines). If winget isn't available, fail with a clear message pointing to https://aka.ms/getwinget.
  That cuts the file from 971 lines to probably ~200, removes the Scoop/Chocolatey/WSL detection entirely, and is still genuinely useful for the
  Windows user you're trying to support.
```
What do you think? Can we follow this strategy to significantly simply `install_tools.py`?

scripts/make_tasks.py
```diff
@@ -0,0 +1,107 @@
+"""Cross-platform implementations for the repository's make targets."""
```
XinweiHe: copied from claude code review

```
  scripts/make_tasks.py — you don't need it. It's a thin indirection layer that exists only because dev-reset and prod-reset were previously raw
  Makefile shell commands (which don't work on Windows). The fix is simpler: just add --reset and --prod-reset flags to launcher.py directly, which
  already owns all this logic anyway. Then the Makefile stays:

  dev:
      uv run python tmux_tools/launcher.py

  dev-reset:
      uv run python tmux_tools/launcher.py --reset
```

scripts/install_tools.py
sundaram2021: I think its not about simplification , Its about friction for the first time contributors/users trying to explore . 
Instead of assuming anything , why not we make our system more reliable enough so that it works everytime .

what do you think ?

tmux_tools/launcher.py
```diff
@@ -12,14 +12,17 @@
...
 
+from backend.db.clickhouse.migrate import run_goose
 from tmux_tools import schema
+from tmux_tools.process import run_command
 
 REST_PORT = 8000
 FRONTEND_PORT = 3000
 AGENT_PORT = 8100
-PROD_COMPOSE = "docker compose -f docker-compose.prod.yml"
+PROD_COMPOSE = ["docker", "compose", "-f", "docker-compose.prod.yml"]
```
XinweiHe: can we keep the original string? we can split it as needed?

tmux_tools/launcher.py
```diff
@@ -292,22 +286,22 @@ def prod_infra_services():
     return [
         schema.Service(
             title="PostgreSQL",
-            command=f"{PROD_COMPOSE} logs -f --tail=50 postgres",
+            command="docker compose -f docker-compose.prod.yml logs -f --tail=50 postgres",
```
XinweiHe: Use a constant instead of using raw string?

tmux_tools/launcher.py
```diff
@@ -292,22 +286,22 @@ def prod_infra_services():
...
             web_urls=[],
         ),
         schema.Service(
             title="ClickHouse",
-            command=f"{PROD_COMPOSE} logs -f --tail=50 clickhouse",
+            command="docker compose -f docker-compose.prod.yml logs -f --tail=50 clickhouse",
```
XinweiHe: ditto

tmux_tools/launcher.py
```diff
@@ -292,22 +286,22 @@ def prod_infra_services():
...
         schema.Service(
             title="Redis",
-            command=f"{PROD_COMPOSE} logs -f --tail=50 redis",
+            command="docker compose -f docker-compose.prod.yml logs -f --tail=50 redis",
             web_urls=[],
         ),
         schema.Service(
             title="MinIO",
-            command=f"{PROD_COMPOSE} logs -f --tail=50 minio",
+            command="docker compose -f docker-compose.prod.yml logs -f --tail=50 minio",
```
XinweiHe: ditto

tmux_tools/launcher.py
```diff
@@ -324,31 +318,31 @@ def make_prod_driver():
         services=[
             schema.Service(
                 title="Web",
-                command=f"{PROD_COMPOSE} logs -f --tail=50 web",
+                command="docker compose -f docker-compose.prod.yml logs -f --tail=50 web",
```
XinweiHe: ditto

tmux_tools/launcher.py
```diff
@@ -324,31 +318,31 @@ def make_prod_driver():
...
                 web_urls=[
                     ("Traceroot UI", f"http://localhost:{FRONTEND_PORT}"),
                 ],
             ),
             schema.Service(
                 title="REST API",
-                command=f"{PROD_COMPOSE} logs -f --tail=50 rest",
+                command="docker compose -f docker-compose.prod.yml logs -f --tail=50 rest",
```
XinweiHe: ditto

tmux_tools/launcher.py
```diff
@@ -324,31 +318,31 @@ def make_prod_driver():
...
             schema.Service(
                 title="Billing Worker",
-                command=f"{PROD_COMPOSE} logs -f --tail=50 billing",
+                command="docker compose -f docker-compose.prod.yml logs -f --tail=50 billing",
                 web_urls=[],
             ),
             schema.Service(
                 title="Agent",
-                command=f"{PROD_COMPOSE} logs -f --tail=50 agent",
+                command="docker compose -f docker-compose.prod.yml logs -f --tail=50 agent",
```
XinweiHe: ditto

tmux_tools/launcher.py
```diff
@@ -324,31 +318,31 @@ def make_prod_driver():
...
             schema.Service(
                 title="Celery Worker",
-                command=f"{PROD_COMPOSE} logs -f --tail=50 worker",
+                command="docker compose -f docker-compose.prod.yml logs -f --tail=50 worker",
                 web_urls=[],
             ),
             schema.Service(
                 title="Billing Worker",
-                command=f"{PROD_COMPOSE} logs -f --tail=50 billing",
+                command="docker compose -f docker-compose.prod.yml logs -f --tail=50 billing",
```
XinweiHe: ditto

tmux_tools/process.py
```diff
@@ -0,0 +1,72 @@
+"""Cross-platform subprocess helpers for developer tooling."""
```
XinweiHe: We do not have to support powershell?

scripts/install_tools.py
XinweiHe: Generally agree, however I do feel like we need to hit a balance between coverage and code maintainability. If we only support the cases as suggested as the above comment would that unblock your case? If so, I'm leaning towards support the bare minimum for now until someone using windows complains then we can decide whether we want to support the corner case.

The bottom line is I want to support people using the latest windows machine to play with this repo, but not trying to support all of the cases, which could become a burden to maintain by itself.

backend/db/clickhouse/migrate.py
```diff
@@ -0,0 +1,206 @@
...
+
+from __future__ import annotations
+
+import argparse
+import os
+import shutil
+import subprocess
+from pathlib import Path
+
+from tmux_tools.process import run_command
```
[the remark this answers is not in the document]
sundaram2021: Resolved. This is already fixed on the branch.

`backend/db/clickhouse/migrate.py` no longer imports `tmux_tools`; it uses a self-contained subprocess helper, so `worker.celery_app` remains importable in the worker image even though `docker/Dockerfile.worker` only copies `/backend` and sets `PYTHONPATH=/app/backend`.

tmux_tools/tmux.py
```diff
@@ -108,15 +147,14 @@ def set_exit_hook(self, cmd: str) -> None:
         self.exec(f'set-hook -g "session-closed" {quote(run_cmd)}')
```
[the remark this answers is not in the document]
sundaram2021: Resolved. This is already fixed on the branch.

The welcome/Instructions pane now uses `-c` instead of `-lc`, so it does not start a login shell and won’t source `~/.bash_profile`, `~/.profile`, pyenv/rbenv/conda init blocks, etc. before rendering the banner.

tmux_tools/launcher.py
```diff
@@ -39,59 +42,45 @@ def ensure_env_file():
...
 def ensure_migrations():
     """Run pending migrations for both Postgres and ClickHouse."""
     print("Running PostgreSQL migrations (Prisma)...")
-    subprocess.run(
+    run_command(
         ["pnpm", "db:migrate"],
         cwd="frontend/packages/core",
-        check=True,
     )
     print("Running ClickHouse migrations (goose)...")
```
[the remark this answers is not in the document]
sundaram2021: Resolved.

`ensure_infra()` no longer does two back-to-back `docker compose up -d` calls for the same long-running services. It now uses a single `docker compose up -d --wait postgres clickhouse minio redis`, then starts `minio-init` separately because that service is a one-shot setup container and is intentionally not part of the `--wait` group.

scripts/bootstrap_dev_tools.py
```diff
@@ -0,0 +1,971 @@
...
+    # Validate tool names
+    unknown = set(args.tools) - set(TOOL_KEYS)
+    if unknown:
+        p.error(f"Unknown tool(s): {', '.join(sorted(unknown))}. Valid: {', '.join(TOOL_KEYS)}")
+    return args
+
+
+def main() -> None:
+    args = parse_args()
+    selected = set(args.tools) if args.tools else set(TOOL_KEYS)
```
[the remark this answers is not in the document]
sundaram2021: `_msys2_candidates()` no longer uses `Path(os.environ.get("SCOOP", "")) or ...`. It now explicitly checks whether `SCOOP` is set and falls back to `Path.home() / "scoop"` only when it is unset, so the fallback branch is reachable and the generated MSYS2 candidate paths are correct.

scripts/install_tools.py
sundaram2021: Implemented.

I simplified the Windows tmux installer path in `scripts/install_tools.py` to the single supported flow we discussed: reuse an existing standard MSYS2 install when present, otherwise require `winget`, install MSYS2 with it, and then install tmux via `pacman`.

That removes the extra Scoop/Chocolatey/WSL fallback branches while still keeping the common existing-MSYS2 case working.

scripts/make_tasks.py
sundaram2021: Resolved. 

`scripts/make_tasks.py` has been removed, and the reset logic now lives directly in `tmux_tools/launcher.py` behind `--reset` and `--prod-reset`.

The Makefile now calls the launcher directly:
- `make dev` -> `uv run python tmux_tools/launcher.py`
- `make dev-autoreload` -> `uv run python tmux_tools/launcher.py --autoreload`
- `make dev-reset` -> `uv run python tmux_tools/launcher.py --reset`
- `make prod` -> `uv run python tmux_tools/launcher.py --prod`
- `make prod-reset` -> `uv run python tmux_tools/launcher.py --prod-reset`

tmux_tools/launcher.py
sundaram2021: resolved this one and all the similiar below ones

tmux_tools/process.py
sundaram2021: I’d keep it here.

`tmux_tools/process.py` is our generic Windows subprocess shim for developer tooling, and some Windows CLI entrypoints resolve to `.ps1` wrappers. Supporting `.ps1` here is intentional so those tools work correctly from Python, similar to the existing `.cmd`/`.bat` handling.

I removed PowerShell-specific handling from `backend/db/clickhouse/migrate.py` because it only runs `goose`/`docker`, so it was dead there. In `tmux_tools/process.py`, though, this is still useful compatibility rather than unnecessary complexity.

XinweiHe (approved):
LGTM!

Before merging this let's add some screenshot from running `make prod`, `make dev` in linux/mac and make sure there are no regression. And screenshot from running `make prod-lite` from a Windows system and make sure it really works.

sundaram2021:
make dev
image
image


please   run it on mac/linux  
these are from windows

cc: @XinweiHe

XinweiHe:
> please run it on mac/linux

I tried this earlier today already. Works.

sundaram2021:
> How does the `make dev` works with Windows in your case? Does your windows setup happens to work with tmux? I would expected `make prod-lite` to work with most of the windows setup though?

`make dev`  runs through tmux but `make prod-lite ` will work on all the windows without any tmux requirement

PR #941, opened 2026-05-24
Plan-tiered rate limiting

lqiu03:
## Summary

The REST API has no rate limiting today. A single workspace can potentially send unlimited trace ingestion or dashboard read requests, and because cloud is a shared, multi-tenant deployment, one workspace under heavy load can slow the API down for every other customer. This PR adds rate limiting to contain that.

## Type of Change
- [x] feat - A new feature

Limits are applied per workspace and scaled by billing plan. They only take effect on cloud — self-hosted deployments are unaffected (see Notes).

## What changed

Two kinds of traffic are limited, with separate budgets:

- Ingestion — `POST /public/traces`, where SDKs push trace data.
- Dashboard reads — the trace / session / user list endpoints the UI calls, which share one budget per workspace.

## Design notes:

- Limited per workspace, not per API key. Per-key limits would let a customer create more keys to get more quota; keying on the workspace prevents that.
- Fails open. Rate limiting must never be the reason a request fails. Counters live in Redis, shared across API replicas. If Redis is unreachable each replica counts in memory; if the limiter itself errors, the request is allowed through rather than returning a 500.
- Standard 429 contract. Over-limit requests return HTTP 429 with `Retry-After` and `X-RateLimit-Limit`, `X-RateLimit-Remaining`, and `X-RateLimit-Reset`, so SDKs and callers can back off.
- Internal service-to-service calls are exempt.

## Default limits (Unit: per minute)

| Plan | Ingestion | Dashboard reads |
| --- | ---: | ---: |
| Free | 1,000 | 60 |
| Starter | 5,000 | 300 |
| Pro | 20,000 | 1,000 |
| Enterprise | same as Pro | same as Pro |

Every value is overridable with `RATE_LIMIT_*` environment variables.

## Self-host vs cloud

The Helm chart defaults billing to off, so a plain `helm install` runs with no limiting. On cloud, Terraform sets billing on, which overrides the chart default. Cloud behavior does not change.

## Rolling it out

There's a shadow mode (`RATE_LIMIT_SHADOW=true`): the limiter checks every request and logs the ones it *would* block, without actually blocking them. The plan is to run shadow first on cloud, confirm the limits are reasonable against real traffic, adjust the numbers (env vars only, no code change), then switch to enforcing. If anything goes wrong, `RATE_LIMIT_ENABLED=false` disables it instantly with no deploy.

## Notes

Rate limiting only takes effect on Traceroot cloud, no limits on self-host but depends on user's infra. It is gated on `ENABLE_BILLING`: the Helm chart defaults this to `false`, so a plain `helm install` runs with limiting off and all tiers unlocked. Cloud sets `ENABLE_BILLING=true` via Terraform, which overrides the chart default. This matches Langfuse, which applies no limits when not running as cloud.

## Test plan

Automated — `tests/rest/test_rate_limit.py` (25 tests, part of the suite; 303 pass):

- [x] Per-tier enforcement: with a tier's limit overridden to a small number, the (N+1)th request returns 429 while a higher tier at the same request count does not — `test_resolve_limit_applies_per_plan_limits`.
- [x] 429 carries the JSON envelope + `Retry-After` / `X-RateLimit-*` — `test_enforcement_third_request_is_throttled_with_429_envelope`.
- [x] Ingestion and read buckets have independent counters — `test_ingest_and_read_buckets_do_not_share_a_counter`.
- [x] traces/sessions/users share one read budget; a missing workspace falls back to a shared bucket and logs a warning — `test_missing_workspace_falls_back_to_unknown_bucket_and_warns`.
- [x] Self-host disables the limiter when `ENABLE_BILLING=false` — `test_self_host_disables_limiter_*`.
- [x] Shadow mode counts over-limit requests without blocking — `test_shadow_mode_records_but_does_not_block`.

Need @XinweiHe a maintainer to run manually the following, on cloud staging before enforcing:

- [ ] Run with shadow mode on, watch logs for how often real traffic would be limited, tune the numbers.
- [ ] Set one tier's limit small, send requests past it, confirm a 429 with the headers below, then restore.
- [ ] On a self-hosted build, confirm requests are never throttled.

## Sample 429

Free tier overridden to 3/min, 4th read:

```http
HTTP/1.1 429 Too Many Requests
retry-after: 61
x-ratelimit-limit: 3
x-ratelimit-remaining: 0
x-ratelimit-reset: 1779579045

{"error":"too_many_requests","detail":"Rate limit exceeded. Slow down and retry after the cooldown.","retry_after":61,"limit":3,"remaining":0}
```

## Headroom

The dashboard auto-refreshes every ~5s and those reads share the per-workspace read budget. Free tier read = 60/min covers a single active tab with room to spare; this is the limit most worth validating in shadow mode before enforcing, and raising if real traffic warrants.

## Rollout

Enable shadow mode first (`RATE_LIMIT_SHADOW=true`) to measure against real traffic, tune limits via env vars (no code change), then turn shadow off to enforce. Instant rollback: `RATE_LIMIT_ENABLED=false` disables limiting with no deploy.

## Checklist

---
## Summary by cubic
Add tiered, per‑workspace rate limiting for trace ingestion and dashboard reads to protect cloud performance and enforce plan quotas. Over‑limit requests return 429 with a JSON error and a Retry‑After derived from the active window; self‑host remains unlimited.

- **New Features**
  - Per‑workspace limits keyed by workspace and plan; two buckets: ingestion (`POST /public/traces`) and a shared read budget for traces/sessions/users. Internal service calls are exempt.
  - Defaults per minute: Free 1k ingest / 60 reads; Starter 5k / 300; Pro 20k / 1k; Enterprise mirrors Pro.
  - 429 contract with `Retry-After` and `X-RateLimit-*` headers; exceeded requests are logged and counted via `opentelemetry-api`. Backed by `slowapi` (`>=0.1.9,<0.2`) with Redis storage and in‑memory fallback (fail‑open).
  - Cloud‑only gating via `ENABLE_BILLING` (trimmed, case‑insensitive). Ops knobs: `RATE_LIMIT_ENABLED` (kill‑switch), `RATE_LIMIT_STORAGE_URI` (defaults to `REDIS_URL`). Reads now require `workspaceId`; ingestion rejects empty `workspaceId`. Missing `billingPlan` falls back to `free`. Pro plan copy updated to “20k ingest + 1k dashboard requests/min”.

- **Migration**
  - Cloud: enable with `RATE_LIMIT_ENABLED=true`; rollback by setting `RATE_LIMIT_ENABLED=false`. Self‑host is off by default (`ENABLE_BILLING=false`). Helm defaults `enableBilling=false`, and prod compose now passes `ENABLE_BILLING` to the `rest` service (set to `false` to disable); a deploy test guards this.
  - Shadow mode removed; use the kill‑switch for rollout.

<sup>Written for commit 21e4eb097b7512ac854a6af858195a7a41bad618. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/941?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/rest/rate_limit.py
+++ b/backend/rest/rate_limit.py
@@ -159,8 +159,9 @@ def resolve_limit(key: str) -> str:
     WARNING: the parameter MUST be named ``key`` — slowapi only forwards the
     bucket key when the callable's signature contains a literal ``key`` param
     (otherwise it calls this with no args). Renaming it would silently disable
-    enforcement (errors are swallowed). ``test_resolve_limit_signature`` guards
-    against this; do not rename without updating that test.
+    enforcement (errors are swallowed). ``test_resolve_limit_signature_is_literal_key``
+    (signature) and ``test_resolve_limit_applies_per_plan_limits`` (behavior)
+    guard against this; do not rename without updating those tests.
     """
     parts = key.split(":", 3)
     if len(parts) == 4:
--- a/backend/rest/rate_limit.py
+++ b/backend/rest/rate_limit.py
@@ -10,6 +10,11 @@ Design
   a customer cannot multiply quota by minting more keys.
 * **Tiers**    per ``(bucket, billing-plan)`` limits resolved at request time
   from ``settings.rate_limit`` (see ``RateLimitSettings``).
+* **Self-host** deployments (``ENABLE_BILLING=false``) have no billing tiers, so
+  rate limiting is disabled entirely — the limiter is built with
+  ``enabled=False`` (see ``_build_limiter``). This matches Langfuse, which
+  applies no limits when not running as cloud; the operator's own infra is the
+  ceiling. Limits therefore apply on cloud (billing-enabled) deployments only.
 * **Buckets**  ``ingest`` (POST /public/traces) and ``read`` (dashboard GETs,
   which share one budget via ``shared_limit(scope="read")``).
 * **Response** HTTP 429 with a JSON envelope plus ``Retry-After`` and
@@ -40,6 +45,7 @@ from slowapi import Limiter
 from slowapi.errors import RateLimitExceeded
 from slowapi.util import get_remote_address
 
+from ee.license import is_billing_enabled
 from shared.config import normalize_plan, settings
 
 logger = logging.getLogger(__name__)
@@ -110,6 +116,9 @@ def set_rate_limit_identity(
     workspace), so it is a fail-safe (restrictive), not an expected state. We
     warn when a non-exempt request lands there so a broken validate contract is
     visible rather than silently collapsing tenants into one shared bucket.
+
+    Note: on self-host the limiter is disabled (see ``_build_limiter``), so this
+    runs but is never read; it only matters on cloud (billing-enabled).
     """
     if not workspace_id and not is_request_rate_limit_exempt():
         logger.warning(
@@ -161,23 +170,38 @@ def resolve_limit(key: str) -> str:
     return settings.rate_limit.limit_for(bucket, plan)
 
 
+def _storage_state(request: Request) -> str:
+    """ "ok" / "degraded": whether the limiter is on its in-memory fallback.
+
+    slowapi flips ``_storage_dead`` when Redis is unreachable and it falls back
+    to per-process in-memory counters — which multiplies the effective limit by
+    the replica count. Surfacing it on every throttle lets ops alert on
+    degraded-mode rate limiting (best-effort: only observable via throttle events).
+    """
+    app_state = getattr(getattr(request, "app", None), "state", None)
+    limiter_obj = getattr(app_state, "limiter", None)
+    return "degraded" if getattr(limiter_obj, "_storage_dead", False) else "ok"
+
+
 def _record_exceeded(request: Request, retry_after: int) -> None:
     bucket = getattr(request.state, "rl_bucket", "unknown")
     workspace_id = getattr(request.state, "rl_workspace_id", _UNKNOWN_WORKSPACE)
     plan = getattr(request.state, "rl_billing_plan", "unknown")
+    storage = _storage_state(request)
     logger.warning(
-        "rate limit exceeded: bucket=%s workspace=%s plan=%s retry_after=%ss",
+        "rate limit exceeded: bucket=%s workspace=%s plan=%s retry_after=%ss storage=%s",
         bucket,
         workspace_id,
         plan,
         retry_after,
+        storage,
     )
     if _exceeded_counter is not None:
         try:
             # Only bounded-cardinality attributes here. workspace_id is high
             # cardinality and would blow up a metrics backend — it stays in the
             # log line above for debugging, not on the metric.
-            _exceeded_counter.add(1, {"bucket": bucket, "plan": plan})
+            _exceeded_counter.add(1, {"bucket": bucket, "plan": plan, "storage": storage})
         except Exception:  # pragma: no cover - best-effort metric
             logger.debug("failed to record rate_limit.exceeded counter", exc_info=True)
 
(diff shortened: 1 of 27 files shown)


lqiu03:
Retry-After is now calculated from the limit's actual window instead of a hardcoded 60 seconds, we agreed to remove the shadow mode fallback (it relied on slowapi), and a request with no workspace now gets rejected — plus a follow-up outside of our discussion but is a deploy bug ([d297851](https://github.com/traceroot-ai/traceroot/pull/941/commits/d297851eed5a2fbd46cac5884b59b0d6c309364b))

backend/rest/routers/public/traces.py
```diff
@@ -39,6 +46,23 @@
 router = APIRouter(prefix="/public/traces", tags=["Traces (Public)"])
 
 
+async def authenticate_and_stamp_identity(request: Request, auth: Auth) -> AuthResult:
+    """Authenticate, then stamp the workspace/plan onto the request for limiting.
+
+    A thin wrapper over ``authenticate_api_key`` so the rate limiter can key the
```
XinweiHe, line 52: Add input variable type and return type.

backend/rest/rate_limit.py
```diff
@@ -0,0 +1,300 @@
...
+    """Return whether the current request is exempt (slowapi ``exempt_when``)."""
+    return _rate_limit_exempt.get()
+
+
+def set_rate_limit_identity(
+    request: Request, workspace_id: str | None, billing_plan: str | None
+) -> None:
+    """Stash the resolved workspace + plan on the request for the key/limit funcs.
+
+    Called from the dependency layer, which runs before slowapi evaluates the
```
XinweiHe, line 109: Add input variable type and return type.

backend/rest/rate_limit.py
```diff
@@ -0,0 +1,300 @@
...
+            # cardinality and would blow up a metrics backend — it stays in the
+            # log line above for debugging, not on the metric.
+            _exceeded_counter.add(1, {"bucket": bucket, "plan": plan, "storage": storage})
+        except Exception:  # pragma: no cover - best-effort metric
+            logger.debug("failed to record rate_limit.exceeded counter", exc_info=True)
+
+
+def rate_limit_exceeded_handler(request: Request, exc: RateLimitExceeded) -> JSONResponse:
+    """Return a JSON 429 with accurate ``Retry-After`` / ``X-RateLimit-*`` headers.
+
```
XinweiHe, line 218: Add input variable type and return type.

backend/shared/config.py
```diff
@@ -56,6 +56,68 @@ class RedisSettings(BaseSettings):
...
+_PLAN_LIMITS_READ: dict[str, str] = {
+    "free": "60/minute",
+    "starter": "300/minute",
+    "pro": "1000/minute",
+    "enterprise": "1000/minute",
+}
+
+
+def normalize_plan(plan: str | None) -> str:
+    """Normalize a billing-plan string to a known plan, defaulting to ``free``.
```
XinweiHe, line 80: ditto

backend/rest/routers/public/traces.py
lqiu03: `async def authenticate_and_stamp_identity(request: Request, auth: Auth) -> AuthResult:`
I think it's in line49 here~

backend/rest/rate_limit.py
lqiu03: `request: Request, workspace_id: str | None, billing_plan: str | None` on line 105~

backend/rest/rate_limit.py
lqiu03: `def ...(request: Request, exc: RateLimitExceeded) -> JSONResponse`: on line205~

backend/shared/config.py
lqiu03: `def normalize_plan(plan: str | None) -> str:` on line79~

XinweiHe (approved):

PR #1018, opened 2026-06-02
fix(detectors): support BYOK models for automatic RCA

### LINKED ISSUE #991: Automatic detector RCA does not use BYOK model selection
### Environment

Windows 11, Docker local dev, Chrome, Gemini BYOK configured, No Anthropic or OpenAI key in env

### Steps to Reproduce

1. Add a Gemini BYOK provider in workspace settings.
2. Set Project Settings → Detectors → Agent Model to `gemini-2.5-flash`.
3. Create a detector that always returns a finding.
4. Send a new trace.
5. Open the finding so automatic RCA starts.

### Expected Behavior

Automatic detector RCA should use the saved Gemini BYOK model.

### Actual Behavior

Automatic detector RCA falls back to the default system model.

Manual chat in the same RCA panel does use Gemini BYOK correctly.

```shell
[detector-rca] Unknown rca_model "gemini-2.5-flash", falling back to default
[Agent] POST message: session=..., model=undefined, provider=undefined, source=undefined
[Agent] Using model="claude-sonnet-4-5" source=system provider=—
[Agent] No API key for provider "anthropic" (no BYOK, no env var)

Manual chat in the same RCA panel:

[Agent] POST message: session=..., model=gemini-2.5-flash, provider=Google Gemini, source=byok
[Agent] Using model="gemini-2.5-flash" source=byok provider=Google Gemini
```

### Screenshots

_No response_

### TraceRoot Version

_No response_

### Additional Context

_No response_

rakshityadav1868:
Fixes #991

## Type of Change
- [x] fix - A bug fix

## Details
 Allowing automatic detector RCA sessions to resolve and use workspace BYOK/custom models instead of only models present in SYSTEM_MODELS.

Previously, when a project was configured with a BYOK model such as:

* deepseek/deepseek-chat-v3
* gemini-2.5-flash

the detector RCA worker failed to resolve the model, omitted model, providerName, and source from the agent request, and the agent service fell back to the default Claude model.

## Root Cause

resolveSystemModel() only checked the static SYSTEM_MODELS registry.

## Changes

* Replaced resolveSystemModel() with resolveProjectModel()
* Added support for resolving workspace BYOK models through modelProvider
* Preserved existing behavior for system models
* Added tests covering:
    * system model resolution
    * BYOK model resolution
    * unknown model fallback
    * null/undefined model handling

https://github.com/user-attachments/assets/0a54fbf3-6394-4c3c-a36a-f0eee960c5d2

## Checklist

---
## Summary by cubic
Fixes automatic RCA not using workspace BYOK models by resolving and sending the chosen model, provider, and source end-to-end. Also enforces free-plan RCA caps so blocked workspaces don’t run RCA.

- **Bug Fixes**
  - Replaced resolver with async resolveProjectModel(rcaModel, rcaProvider, rcaSource, workspaceId) using `@traceroot/core/model-resolver` (fetchProviderConfig + resolvePiModel); supports system/BYOK, returns ModelSource.SYSTEM or ModelSource.BYOK, and forwards the BYOK provider label to the agent.
  - Made the legacy BYOK fallback deterministic (provider lookup ordered by id) and added error handling; tests cover duplicate provider resolution and DB error path.
  - Removed prefix-based provider guessing and catalog inference; BYOK now requires a saved provider in the workspace or resolves via legacy fallback, otherwise returns null.
  - Persisted rcaProvider and rcaSource on Project; exposed via GET/PATCH APIs (including `/api/projects/[projectId]`) and wired into Detectors settings; worker reads these in processRcaJob and includes them in runRcaSession. Added tests for routes, runRcaSession, and processRcaJob covering system/BYOK resolution, missing provider, unknown/empty models, free-plan skip, and failure path.

- **Migration**
  - Adds rca_provider and rca_source columns; run DB migrations. No manual backfill needed due to the legacy BYOK fallback.

<sup>Written for commit 1b20e485e592fe6a6d2c9b8155c7250f097080d9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1018?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/packages/core/prisma/schema.prisma
+++ b/frontend/packages/core/prisma/schema.prisma
@@ -138,6 +138,8 @@ model Project {
   updateTime   DateTime  @default(now()) @map("update_time") @db.Timestamp(6)
   // Agent model for Step 2 deep analysis — project-scoped, shared across all detectors
   rcaModel     String?   @map("rca_model") @db.VarChar
+  rcaProvider  String?   @map("rca_provider") @db.VarChar
+  rcaSource    String?   @map("rca_source") @db.VarChar
   accessKeys   AccessKey[]
   aiSessions   AISession[]
   detectors    Detector[]
--- a/frontend/worker/src/processors/detector-rca-processor.ts
+++ b/frontend/worker/src/processors/detector-rca-processor.ts
@@ -31,7 +31,10 @@ export async function resolveProjectModel(
       return null;
     }
     const model = resolvePiModel(rcaModel, providerConfig);
-    return { model: model.id, providerName: model.provider, source: ModelSource.BYOK };
+    // Forward the saved BYOK provider LABEL (e.g. "myopenai"), not the pi-ai
+    // provider name. The agent resolves BYOK via fetchProviderConfig(workspaceId,
+    // providerName), which keys on ModelProvider.provider (the user label).
+    return { model: model.id, providerName: rcaProvider, source: ModelSource.BYOK };
   }
 
   // 2. System model: validate against catalog, then use shared resolver
@@ -69,7 +72,8 @@ async function resolveLegacyByok(
         const providerConfig = await fetchProviderConfig(workspaceId, p.provider);
         if (providerConfig) {
           const model = resolvePiModel(rcaModel, providerConfig);
-          return { model: model.id, providerName: model.provider, source: ModelSource.BYOK };
+          // Forward the BYOK provider LABEL the agent expects (see step 1 above).
+          return { model: model.id, providerName: p.provider, source: ModelSource.BYOK };
         }
       }
     }
--- a/frontend/worker/src/processors/detector-rca-processor.ts
+++ b/frontend/worker/src/processors/detector-rca-processor.ts
@@ -11,9 +11,20 @@ const AGENT_SERVICE_URL = process.env.AGENT_SERVICE_URL || "http://localhost:810
 // Returns null when the model is unset or unknown (caller should omit fields).
 export async function resolveProjectModel(
   rcaModel: string | null | undefined,
+  rcaProvider: string | null | undefined,
+  rcaSource: string | null | undefined,
   workspaceId: string,
 ): Promise<{ model: string; providerName: string; source: ModelSource } | null> {
   if (!rcaModel) return null;
+
+  if (rcaSource && rcaProvider) {
+    return {
+      model: rcaModel,
+      providerName: rcaProvider,
+      source: rcaSource === "byok" ? ModelSource.BYOK : ModelSource.SYSTEM,
+    };
+  }
+
   for (const group of SYSTEM_MODELS) {
     if (group.models.some((m) => m.id === rcaModel)) {
       return { model: rcaModel, providerName: group.piAIProvider, source: ModelSource.SYSTEM };
@@ -79,6 +90,8 @@ async function runRcaSession(params: {
   findings: DetectorRcaJob["findings"];
   hasGitHub: boolean;
   rcaModel?: string | null;
+  rcaProvider?: string | null;
+  rcaSource?: string | null;
 }): Promise<{ result: string; sessionId: string }> {
   const sessionRes = await fetch(
     `${AGENT_SERVICE_URL}/api/v1/projects/${params.projectId}/sessions`,
@@ -142,7 +155,12 @@ Output your findings in this format:
     "x-workspace-id": params.workspaceId,
   };
 
-  const resolved = await resolveProjectModel(params.rcaModel, params.workspaceId);
+  const resolved = await resolveProjectModel(
+    params.rcaModel,
+    params.rcaProvider,
+    params.rcaSource,
+    params.workspaceId,
+  );
   const msgBody: {
     message: string;
     traceId: string;
@@ -329,6 +347,8 @@ export function startDetectorRcaWorker(): Worker<DetectorRcaJob> {
           where: { id: projectId },
           select: {
             rcaModel: true,
+            rcaProvider: true,
+            rcaSource: true,
             alertConfig: {
               select: { emailAddresses: true, slackChannelId: true, slackChannelName: true },
             },
@@ -362,6 +382,8 @@ export function startDetectorRcaWorker(): Worker<DetectorRcaJob> {
           findings,
           hasGitHub,
           rcaModel: project?.rcaModel,
+          rcaProvider: project?.rcaProvider,
+          rcaSource: project?.rcaSource,
         });
 
         await prisma.detectorRca.update({
--- a/frontend/worker/src/processors/detector-rca-processor.ts
+++ b/frontend/worker/src/processors/detector-rca-processor.ts
@@ -42,7 +42,42 @@ export async function resolveProjectModel(
     }
   }
 
+  // 3. Legacy BYOK fallback for pre-existing configs
+  const legacy = await resolveLegacyByok(rcaModel, workspaceId);
+  if (legacy) return legacy;
+
   console.warn(`[detector-rca] Unknown rca_model "${rcaModel}", falling back to default`);
+  return null;
+}
+
+// 3. Legacy BYOK fallback: projects that saved a BYOK model before
+//    rcaSource and rcaProvider fields existed have both set to NULL.
+//    Try to resolve the model from the workspace's enabled providers.
+async function resolveLegacyByok(
+  rcaModel: string,
+  workspaceId: string,
+): Promise<{ model: string; providerName: string; source: ModelSource } | null> {
+  try {
+    const dbProviders = await prisma.modelProvider.findMany({
+      where: { workspaceId, enabled: true },
+      select: { provider: true, customModels: true },
+    });
+
+    for (const p of dbProviders) {
+      if (p.customModels.some((m) => m.trim() === rcaModel)) {
+        const providerConfig = await fetchProviderConfig(workspaceId, p.provider);
+        if (providerConfig) {
+          const model = resolvePiModel(rcaModel, providerConfig);
+          return { model: model.id, providerName: model.provider, source: ModelSource.BYOK };
+        }
+      }
+    }
+  } catch (err) {
+    console.error(
+      `[detector-rca] Failed to resolve legacy BYOK for workspace ${workspaceId}:`,
+      err,
+    );
+  }
   return null;
 }
 
--- a/frontend/worker/src/processors/detector-rca-processor.ts
+++ b/frontend/worker/src/processors/detector-rca-processor.ts
@@ -82,7 +82,7 @@ async function resolveLegacyByok(
   return null;
 }
 
-async function runRcaSession(params: {
+export async function runRcaSession(params: {
   findingId: string;
   projectId: string;
   workspaceId: string;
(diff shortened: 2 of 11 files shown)


frontend/worker/src/processors/detector-rca-processor.ts
```diff
@@ -9,15 +9,38 @@ const AGENT_SERVICE_URL = process.env.AGENT_SERVICE_URL || "http://localhost:810
...
   if (!rcaModel) return null;
   for (const group of SYSTEM_MODELS) {
     if (group.models.some((m) => m.id === rcaModel)) {
-      return { model: rcaModel, providerName: group.piAIProvider, source: "system" };
+      return { model: rcaModel, providerName: group.piAIProvider, source: ModelSource.SYSTEM };
     }
   }
+
+  try {
+    const dbProviders = await prisma.modelProvider.findMany({
```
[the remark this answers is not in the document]
rakshityadav1868: Thanks for the catch. Updated the provider resolution logic to disambiguate duplicate BYOK model IDs and added tests covering those cases.

rakshityadav1868:
Hey @XinweiHe could u check this pr !!

rakshityadav1868:
@cubic-dev-ai review

frontend/worker/src/processors/detector-rca-processor.ts
```diff
@@ -9,15 +9,63 @@ const AGENT_SERVICE_URL = process.env.AGENT_SERVICE_URL || "http://localhost:810
...
```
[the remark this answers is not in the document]
rakshityadav1868: Thanks for catching this.

The fallback path could become nondeterministic when multiple enabled providers shared the same customModel, because findMany() was executed without an explicit ordering and the first matching provider was selected. I’ve updated the query to use: orderBy: { id: "asc" } .which provides a stable deterministic ordering while preserving existing behavior.

frontend/worker/src/processors/detector-rca-processor.ts
```diff
@@ -9,15 +9,64 @@ const AGENT_SERVICE_URL = process.env.AGENT_SERVICE_URL || "http://localhost:810
...
+
+    // First pass: try to find a precise match using adapter prefix or curated catalogs
+    for (const p of dbProviders) {
+      if (p.customModels.some((m) => m.trim() === rcaModel)) {
+        // 1. Prefix match (e.g. "deepseek/deepseek-chat-v3" -> prefix is "deepseek")
+        const hasSlash = rcaModel.includes("/");
+        const prefix = hasSlash ? rcaModel.split("/")[0].toLowerCase() : null;
+        if (prefix && p.adapter.toLowerCase() === prefix) {
+          return { model: rcaModel, providerName: p.provider, source: ModelSource.BYOK };
+        }
```
lqiu03: **Finding 1:**
Here it's taking a guess at provider routing from the model's string prefix. 

I'd suggest taking a reference at:
`frontend/worker/src/detection/sandbox-eval.ts`:
- `sandbox-eval.ts:L122` reads `detector.detectionSource` (saved value)
- `sandbox-eval.ts:L127-130` reads `detector.detectionProvider` and passes it straight to `fetchProviderConfig`(workspaceId, detectionProvider)
- `sandbox-eval.ts:L142` resolves via the shared `resolvePiModel(...)`

You can easily read a provider that was saved when the user picked the model, then uses the shared resolver

frontend/worker/src/processors/detector-rca-processor.ts
```diff
@@ -9,15 +9,64 @@ const AGENT_SERVICE_URL = process.env.AGENT_SERVICE_URL || "http://localhost:810
...
+
+        // 2. Curated catalog match (e.g. "deepseek-chat" belongs to curated deepseek adapter catalog)
+        if (!hasSlash) {
+          const curatedAdapter = Object.keys(ADAPTER_MODELS).find((adapterKey) => {
+            const catalog = ADAPTER_MODELS[adapterKey as keyof typeof ADAPTER_MODELS];
+            return catalog?.some((m) => m.id === rcaModel);
+          });
+          if (curatedAdapter && p.adapter.toLowerCase() === curatedAdapter.toLowerCase()) {
+            return { model: rcaModel, providerName: p.provider, source: ModelSource.BYOK };
```
lqiu03: Finding 1 continued:
Same thing; it's guessing from a built-in catalog

frontend/worker/src/processors/detector-rca-processor.ts
```diff
@@ -9,15 +9,64 @@ const AGENT_SERVICE_URL = process.env.AGENT_SERVICE_URL || "http://localhost:810
...
+          }
+        }
+      }
+    }
+
+    // Second pass: fallback to the first provider containing the model
+    for (const p of dbProviders) {
+      if (p.customModels.some((m) => m.trim() === rcaModel)) {
+        return { model: rcaModel, providerName: p.provider, source: ModelSource.BYOK };
```
lqiu03: Finding 1 continued: Give up and grab the first provider that lists the model

frontend/worker/src/processors/detector-rca-processor.ts
```diff
@@ -9,15 +9,64 @@ const AGENT_SERVICE_URL = process.env.AGENT_SERVICE_URL || "http://localhost:810
...
+      },
+    });
+
+    // First pass: try to find a precise match using adapter prefix or curated catalogs
+    for (const p of dbProviders) {
+      if (p.customModels.some((m) => m.trim() === rcaModel)) {
+        // 1. Prefix match (e.g. "deepseek/deepseek-chat-v3" -> prefix is "deepseek")
+        const hasSlash = rcaModel.includes("/");
+        const prefix = hasSlash ? rcaModel.split("/")[0].toLowerCase() : null;
+        if (prefix && p.adapter.toLowerCase() === prefix) {
```
lqiu03: **Finding 2:**
This code snippet assumes the model name is the same as the model provider's. True for deepseek/deepseek-chat-v3.
 But:
  - claude → adapter is anthropic
  - gemini→ adapter is google
  - gpt→ adapter is openai

  None of those names contain their adapter, so would never fire, or fall back to "unknown model".

And for free-textadapters (azure, openrouter, amazon-bedrock) there's no catalog either, so they fall to "pick the first one"

frontend/worker/src/processors/detector-rca-processor.ts
```diff
@@ -9,15 +9,64 @@ const AGENT_SERVICE_URL = process.env.AGENT_SERVICE_URL || "http://localhost:810
...
     if (group.models.some((m) => m.id === rcaModel)) {
-      return { model: rcaModel, providerName: group.piAIProvider, source: "system" };
+      return { model: rcaModel, providerName: group.piAIProvider, source: ModelSource.SYSTEM };
     }
   }
+
+  try {
+    const dbProviders = await prisma.modelProvider.findMany({
+      where: { workspaceId, enabled: true },
+      orderBy: { id: "asc" },
```
lqiu03, line 67: Saving the provider (Finding 1) removes the need to disambiguate here - the code is doing orderBy: {id, "ASC"}

lqiu03 (commented):
I'm working with @XinweiHe on your PR here :) Please take a look at my comments and lmk if you have any thoughts

XinweiHe:
@rakshityadav1868 Also, it'd be great if we can take a look at why the detector image building is failing in here?

Screenshot 2026-06-08 at 5 57 05 PM

rakshityadav1868:
@lqiu03 cool!!

rakshityadav1868:
Hi @XinweiHe,

I checked the detector build logs. The failure happens during the docker/setup-buildx-action step while Buildx is being initialized:

Get "https://registry-1.docker.io/v2/": Client.Timeout exceeded while awaiting headers

The job fails before the detector image build starts, so it doesn't appear to be related to the detector code changes or Dockerfile itself. Could you please re-run the failed detector build? From the logs, this looks like a transient Docker Hub / GitHub Actions runner connectivity issue.

Thanks!

XinweiHe:
@rakshityadav1868 Ok, re-run and it passes. Can you take a look and resolve @lqiu03 's comment please.

rakshityadav1868:
Sure resolving it!! @XinweiHe

frontend/worker/src/processors/detector-rca-processor.ts
rakshityadav1868: I investigated this path. Detector execution does persist provider/source and uses them directly. RCA currently only persists rca_model, so the worker doesn’t have provider/source available. I’m updating the RCA flow to persist and use those values directly, following the same pattern as detector execution.

frontend/worker/src/processors/detector-rca-processor.ts
rakshityadav1868: Removed the ADAPTER_MODELS-based catalog inference from resolveProjectModel and updated the related tests. RCA no longer attempts to route providers using the built-in catalog.

frontend/worker/src/processors/detector-rca-processor.ts
rakshityadav1868: Good call. Switched to fetchProviderConfig plus resolvePiModel, same pattern as sandbox eval.ts. No more guessing from model string prefixes, the saved provider config drives everything now.

frontend/worker/src/processors/detector-rca-processor.ts
rakshityadav1868: Yep, removed the built in catalog guesswork entirely. The shared resolver handles this now, no more trying to match models against a local list of adapters.

frontend/worker/src/processors/detector-rca-processor.ts
rakshityadav1868: Gone. No more fallback to first provider. The saved config drives everything, and fetchProviderConfig either finds the exact provider or returns null.

frontend/worker/src/processors/detector-rca-processor.ts
rakshityadav1868: Fixed. resolvePiModel uses ADAPTER_TO_PI_AI under the hood so claude maps to anthropic, gemini to google, gpt to openai. Free text adapters like azure, openrouter, and bedrock also work correctly since the adapter is read straight from the saved provider row.

frontend/worker/src/processors/detector-rca-processor.ts
rakshityadav1868: Exactly right. Since we read the provider from saved config via fetchProviderConfig, there is nothing left to disambiguate. Removed the prisma findMany and the orderBy entirely, the whole block is gone.

rakshityadav1868:
hi @XinweiHe fixed all comment's suggestion by @lqiu03

frontend/packages/core/prisma/schema.prisma
```diff
@@ -138,6 +138,8 @@ model Project {
   updateTime   DateTime  @default(now()) @map("update_time") @db.Timestamp(6)
   // Agent model for Step 2 deep analysis — project-scoped, shared across all detectors
   rcaModel     String?   @map("rca_model") @db.VarChar
+  rcaProvider  String?   @map("rca_provider") @db.VarChar
+  rcaSource    String?   @map("rca_source") @db.VarChar
```
lqiu03, line 142: Finding 1:
`schema.prisma` is just a declaration of what Prisma thinks the database looks like. It does not change the database. The actual ALTER TABLE ... ADD COLUMN only happens when a migration file under `frontend/packages/core/prisma/migrations/` is applied. Traceroot uses check-in migration, and you can follow an existing format~

frontend/worker/src/processors/detector-rca-processor.ts
```diff
@@ -280,6 +311,8 @@ export function startDetectorRcaWorker(): Worker<DetectorRcaJob> {
           where: { id: projectId },
           select: {
             rcaModel: true,
+            rcaProvider: true,
+            rcaSource: true,
```
lqiu03: Finding 1 continued:
Both line 314 and 315 won't exist in the BD.
On a deployed database without the migration, Postgres rejects this query (column projects.rca_provider does not exist), so the Prisma call throws on every single RCA job — not just BYOK ones.

frontend/worker/src/processors/detector-rca-processor.ts
lqiu03: Finding 1 continued:
You see? So the regression will fall back to line362-366: 100% of automatic RCAs break, which is worse than the original bug.

frontend/worker/src/processors/detector-rca-processor.ts
```diff
@@ -1,23 +1,47 @@
...
   rcaModel: string | null | undefined,
-): { model: string; providerName: string; source: "system" } | null {
+  rcaProvider: string | null | undefined,
+  rcaSource: string | null | undefined,
+  workspaceId: string,
+): Promise<{ model: string; providerName: string; source: ModelSource } | null> {
   if (!rcaModel) return null;
+
+  // 1. BYOK: read provider from saved config (same pattern as sandbox-eval.ts L126-136)
+  if (rcaSource === "byok" && rcaProvider) {
```
lqiu03, line 25: **Finding 2**
```
25      if (rcaSource === "byok" && rcaProvider) {     // 
  gate requires BOTH fields
     ...
37      // 2. System model: validate against catalog ...
38      for (const group of SYSTEM_MODELS) { ... }
     ...  
45      console.warn(`[detector-rca] Unknown rca_model 
  "${rcaModel}", falling back to default`);
46      return null;                                   // 
  caller omits model → agent defaults to Claude
```
Take a look at Cubic's comment: - from what I understood, it's saying Line 25 gates BYOK on rcaSource === 'byok' && rcaProvider, but every project that saved aBYOK rca_model before this PR has both fields NULL — those rows skip the BYOK branch, fail the SYSTEM_MODELS check, and return null at line 46

This bug will affect only pre-existing BYOK configs

lqiu03 (commented):
@rakshityadav1868 , thanks for submitting over your newest commits. @Cubic-dev-ai and I have some comments on your implemented strategy - take a look and lmk if it makes sense. Also, worth resolving the CI failure as well. Thanks!

frontend/packages/core/prisma/schema.prisma
rakshityadav1868: Good catch, added the migration file at 20260613000000_add_rca_provider_source with the ALTER TABLE statements for both rca_provider and rca_source columns.

frontend/worker/src/processors/detector-rca-processor.ts
rakshityadav1868: Right, the migration is now in place so those columns will exist when deployed. Also added a legacy fallback so pre existing BYOK configs with NULL rcaSource and rcaProvider still resolve correctly.

frontend/worker/src/processors/detector-rca-processor.ts
rakshityadav1868: Fixed. The migration ensures the columns exist, and the legacy fallback resolves pre existing BYOK models that have NULL rcaSource and rcaProvider so they do not fall through to the catch block.

frontend/worker/src/processors/detector-rca-processor.ts
rakshityadav1868: Good point. Added a resolveLegacyByok fallback that looks up the model in the workspace enabled providers when rcaSource and rcaProvider are NULL, so pre existing BYOK configs do not silently fall back to the default model.

frontend/worker/src/processors/detector-rca-processor.ts
```diff
@@ -1,23 +1,53 @@
...
```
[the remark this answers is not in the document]
rakshityadav1868: Fixed. The resolveLegacyByok fallback handles pre existing configs where rcaSource and rcaProvider are NULL by looking up the model in the workspace enabled providers, so these projects do not silently fall back to the default model.

rakshityadav1868:
hey @lqiu03 fixed all the suggested fixes now if everything looks good may i commit the test coverage??

frontend/worker/src/processors/detector-rca-processor.ts
```diff
@@ -1,27 +1,86 @@
...
+  workspaceId: string,
+): Promise<{ model: string; providerName: string; source: ModelSource } | null> {
+  try {
+    const dbProviders = await prisma.modelProvider.findMany({
+      where: { workspaceId, enabled: true },
+      select: { provider: true, customModels: true },
+    });
+
+    for (const p of dbProviders) {
+      if (p.customModels.some((m) => m.trim() === rcaModel)) {
```
lqiu03, line 71: @Cubic-dev-ai suggested `resolveLegacyByok` should add `orderBy: { id: "asc" }` to the findMany for deterministic provider selection when a model appears under multiple enabled providers — same fix as d938fb66, which got dropped in the rewrite.

lqiu03 (commented):
Both findings resolved, and now a small additional change needed. Feel free to add more test scripts @rakshityadav1868

frontend/worker/src/processors/detector-rca-processor.ts
rakshityadav1868: Good catch, added orderBy id asc to the findMany in resolveLegacyByok so provider selection is deterministic when a model appears under multiple enabled providers.

lqiu03:
Hi! The CI test is failing on coverage — the frontend tests are at 76% but need to be at 80%. You'll see the big red X and error message in the checks section. Could you add tests to cover the 42 missing lines?

rakshityadav1868:
yes @lqiu03 i am fixing

rakshityadav1868:
@lucas added a few more tests to push the diff coverage past 80%. Covered the free plan RCA cap path, the legacy BYOK error handler, and the process job failure catch block.

lqiu03 (approved):
LGTM, cc @XinweiHe

frontend/packages/core/prisma/schema.prisma
```diff
@@ -138,6 +138,8 @@ model Project {
   updateTime   DateTime  @default(now()) @map("update_time") @db.Timestamp(6)
   // Agent model for Step 2 deep analysis — project-scoped, shared across all detectors
   rcaModel     String?   @map("rca_model") @db.VarChar
+  rcaProvider  String?   @map("rca_provider") @db.VarChar
```
XinweiHe, line 141: @dark-sorceror

dark-sorceror:
### E2E testing results
Naming of the model other than default name still results in failure of automatic RCA

```
agent-1    | [Agent] POST message: ... model=gpt-5.5, provider=openai, source=byok
agent-1    | Error: BYOK provider "openai" not found or disabled. Check workspace settings.
detector-1 | [RCA] Job rca-... failed: Failed to send RCA message: HTTP 500
```

### Root Cause
`resolveProjectModel` returns `providerName: model.provder` - the pi-ai provider name (i.e. openai), but the agent resolves BYOK via `fetchProviderConfig(workspaceId, providerName)` which keys on the  `ModelProvider` label (i.e. myopenai)

### Fix
Point to the saved provider label, not the pi-ai name.

frontend/packages/core/prisma/schema.prisma
dark-sorceror: `schema.prisma` reconciles cleanly when tested locally. Auto merged with all three fields: (`rcaProvider`, `rcaSource`, and `enableRca`)

dark-sorceror:
Detector RCA passes successfully using BYOK
- Specifically a custom named model provider (myopenai) is properly recognized

```
agent-1    | [Agent] POST message: session=cmqfxddjn00czpq3ibogo8pz9, model=gpt-5.5, provider=myopenai, source=byok
agent-1    | [Agent] Using model="gpt-5.5" source=byok provider=myopenai {"id":"gpt-5.5","name":"GPT-5.5","api":"openai-completions","provider":"openai","baseUrl":"https://api.openai.com/v1","reasoning":true,"thinkingLevelMap":{"off":null,"xhigh":"xhigh"},"input":["text","image"],"cost":{"input":5,"output":30,"cacheRead":0.5,"cacheWrite":0},"contextWindow":272000,"maxTokens":128000}
agent-1    | [Agent] Agent ready, running prompt: "A detector fired on this trace.
```
- `provider=myopenai` - pointing to the custom name
image

image

rakshityadav1868:
Thanks @dark-sorceror ! I missed this edge case around forwarding the saved BYOK provider label. Appreciate the fix.

XinweiHe (approved):
LGTM

PR #1020, opened 2026-06-03
feat(traces): persist the Live toggle per project

### LINKED ISSUE #1019: Persist the Live toggle on the traces list (per-project)
## Summary

The "Live" auto-refresh toggle on the traces list page resets to off on every reload, navigation, and re-login. Users who want a live view have to re-enable it each time. This persists the toggle's state per project so it's remembered across sessions.

## Problem

On `frontend/ui/src/app/projects/[projectId]/traces/page.tsx`, the toggle is backed by ephemeral component state:

```ts
const [autoRefresh, setAutoRefresh] = useState(false); // traces/page.tsx:57
```

It drives a 5s poll (`refetchInterval: autoRefresh ? 5000 : false`). Because the state lives only in React, the choice is lost on any reload or navigation. Date range and pagination are already persisted (to the URL); the live toggle is the odd one out.

## Videos

https://github.com/user-attachments/assets/8b00a162-e6b1-4d5b-b92d-318faa5cca80

## Proposed change

Persist the toggle to `localStorage`, scoped per project.

- **Storage key:** `traceroot:traces:live:v1:{projectId}` — per-project (so a busy project can be live while a quiet one isn't), and version-prefixed so the schema can evolve.
- **Default:** `false`. A project the user has never touched behaves exactly as today; only an explicit opt-in is remembered. This keeps the default polling cost unchanged — no project goes live unless a user asked for it.
- **Add a small reusable hook** in `frontend/ui/src/lib/hooks/` (e.g. `use-local-storage.ts`) — this is the codebase's first `localStorage` usage, so it's worth a clean, typed primitive alongside the existing `useListPageState` / `useUrlDateFilter` family rather than inlining.

### Implementation notes

- **SSR/hydration:** the traces page is a client component but is still server-rendered, where `localStorage` is undefined. The hook must return the default during hydration and only adopt the stored value after mount, so the first client render matches the server render (React 18's `useSyncExternalStore` is a clean way to do this). Consequence: on a hard load where the saved value is "on," there's a brief moment at default-off before it flips on — expected and acceptable.
- **Safety:** wrap every `getItem`/`setItem` in try/catch — these throw in private/incognito mode, on quota, or when storage is disabled. Fall back to the default on any error.
- **Cross-tab sync** (listen for the `storage` event) is a low-cost nice-to-have the hook can provide for free, but isn't required for this issue.

## Scope

**In scope**
- Persist the Live toggle per project, default off.
- New reusable `localStorage` hook.

**Out of scope (deferred)**
- A separate manual refresh button.
- Configurable / changed polling interval (stays at 5s).
- Auto-pausing live when a historical time range is selected.
- Switching the list from polling to a push/streaming transport.

## Acceptance criteria

- [ ] Enabling Live, then reloading the page, keeps Live enabled.
- [ ] The setting is independent per project.
- [ ] A project never visited (or never toggled) defaults to Live **off**.
- [ ] No hydration warnings in the console on load.
- [ ] Toggle still works with storage unavailable (private mode) — falls back to in-memory default, no crash.

ka1kqi:
## Summary

The traces list **Live** auto-refresh toggle resets to off on every reload, navigation, and re-login because it is backed by ephemeral component state. This persists the toggle per project in `localStorage`, so an explicit opt-in is remembered across sessions. Date range and pagination already persist (to the URL); this brings the Live toggle in line.

Closes #1019.

## What changed

- **New reusable, SSR-safe `useLocalStorage` hook** (`frontend/ui/src/lib/hooks/use-local-storage.ts`):
  - First render returns the default so it matches the server render — no hydration warning; the stored value is adopted in an effect after mount.
  - Every `localStorage` read/write is wrapped in try/catch, so it degrades safely in private mode or when storage is disabled — the toggle still works in-session from React state.
  - Cross-tab sync via the `storage` event.
  - Built on small pure helpers (`readStored` / `writeStored`) that hold all the guard logic and are unit-tested.
- **Wire the Live toggle to the hook**, keyed per project: `traceroot:traces:live:v1:{projectId}`, default `false`. A project that was never toggled behaves exactly as before — polling cost is unchanged unless a user opts in.

## Why this approach

Uses `useState` + an adopt-after-mount effect rather than `useSyncExternalStore`: it keeps the first client render identical to the server render (no hydration mismatch) and guarantees the toggle still functions in-session even when `localStorage` throws, which a pure external-store-over-`localStorage` cannot.

## Validation

- **8 unit tests** (Vitest, node env) for the storage helpers: SSR-absent window, missing key, valid parse, malformed JSON, and throwing `getItem` / `setItem`.
- **Manually verified in the running app** with the trace detail panel closed (to isolate the list behavior from the detail live-stream):
  - Live **ON** → a newly ingested trace appears in the list within ~5s, no reload.
  - Live **OFF** → the list stays frozen; the trace appears only after a manual reload.
  - Toggle state survives reload, is independent per project (distinct `localStorage` keys), defaults to off for never-toggled projects, produces no hydration warnings, and does not crash when storage is unavailable.

## Scope

Persists the toggle only. Out of scope (deferred per the issue): a separate manual refresh button, configurable poll interval, auto-pausing Live on historical ranges, and switching the list to a push/streaming transport.

## Demo

https://github.com/user-attachments/assets/80716bee-4ec1-4dfa-bdb2-641bac491e47

---
## Summary by cubic
Persist the Traces “Live” auto‑refresh toggle per project using `localStorage`, so the user’s choice survives reloads, navigation, and re-login. Defaults remain off; behavior is unchanged unless a user opts in.

- **New Features**
  - Added SSR-safe `useLocalStorage` hook with guarded read/write, adopt-after-mount to avoid hydration warnings, and cross‑tab sync; helpers are unit‑tested.
  - Wired the Traces Live toggle to the hook with key `traceroot:traces:live:v1:{projectId}`; per‑project persistence with default off.

<sup>Written for commit 8ba6c3c667571d8976498036bb4e5c12628f8bdd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1020?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/ui/src/app/projects/[projectId]/traces/page.tsx
+++ b/frontend/ui/src/app/projects/[projectId]/traces/page.tsx
@@ -20,6 +20,7 @@ import {
 import type { TraceListItem } from "@/types/api";
 import { useTraces } from "@/features/traces/hooks";
 import { useListPageState } from "@/lib/hooks/use-list-page-state";
+import { useLocalStorage } from "@/lib/hooks/use-local-storage";
 import { TraceViewerPanel, GettingStarted } from "@/features/traces/components";
 import { formatContentPreview } from "@/features/traces/utils";
 import { useSession as useAuthSession } from "@/lib/auth-client";
@@ -54,7 +55,12 @@ export default function TracesPage() {
   } = useListPageState();
 
   const [selectedTraceId, setSelectedTraceId] = useState<string | null>(traceIdFromUrl);
-  const [autoRefresh, setAutoRefresh] = useState(false);
+  // Persisted per-project so a live view survives reloads, navigation, and re-login.
+  // Default false: a project the user never toggled behaves exactly as before.
+  const [autoRefresh, setAutoRefresh] = useLocalStorage(
+    `traceroot:traces:live:v1:${projectId}`,
+    false,
+  );
 
   // Fetch traces with combined query options + user filter from URL
   const { data, isLoading, error } = useTraces(
--- a/frontend/ui/src/lib/hooks/use-local-storage.test.ts
+++ b/frontend/ui/src/lib/hooks/use-local-storage.test.ts
@@ -1,0 +1,89 @@
+import { afterEach, describe, expect, it } from "vitest";
+import { readStored, writeStored } from "./use-local-storage";
+
+// Minimal in-memory Storage stand-in (env is "node", so there is no real window).
+function makeStorage(initial: Record<string, string> = {}): Storage {
+  const data = new Map(Object.entries(initial));
+  return {
+    getItem: (k: string) => (data.has(k) ? (data.get(k) as string) : null),
+    setItem: (k: string, v: string) => {
+      data.set(k, v);
+    },
+    removeItem: (k: string) => {
+      data.delete(k);
+    },
+    clear: () => data.clear(),
+    key: (i: number) => Array.from(data.keys())[i] ?? null,
+    get length() {
+      return data.size;
+    },
+  } as Storage;
+}
+
+// The hook reads `window.localStorage`; in the node test env we install a fake
+// `window` on globalThis (a bare `window` reference resolves to globalThis.window).
+function setWindow(storage: Storage | undefined): void {
+  (globalThis as unknown as { window?: { localStorage: Storage } }).window = storage
+    ? { localStorage: storage }
+    : undefined;
+}
+
+afterEach(() => {
+  delete (globalThis as unknown as { window?: unknown }).window;
+});
+
+describe("readStored", () => {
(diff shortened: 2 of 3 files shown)


XinweiHe (approved):
LGTM

XinweiHe:
Let's sign the CLA first before the merge.

ka1kqi:
signed

PR #1022, opened 2026-06-03
fix(worker): price cache read/write tokens at ingest (#956)

### LINKED ISSUE #956: Fix cache token cost in ingest pipeline
### Problem

The Python ingest cost calculator (`backend/worker/tokens/pricing.py::calculate_cost`) only multiplies `input_tokens * input_rate + output_tokens * output_rate`. It ignores `cache_read` and `cache_write` token counts entirely, even though the rate JSON has correct `cacheRead` / `cacheWrite` fields for every Claude and OpenAI entry.

Result: `spans.cost` is silently undercounted for every trace that uses prompt caching. Most visible for Anthropic (where customers can have 90%+ of their prompt tokens come from cache) but also affects OpenAI prompt-cached spans.

The TS path at `frontend/packages/core/src/model-pricing/lookup.ts::calculateCost` already does this correctly — it's the Python ingest path that's broken.

### Scope

- Extend `backend/worker/tokens/pricing.py::calculate_cost` to accept `cache_read_tokens` and `cache_write_tokens` and include them in the sum.
- Update the caller in `backend/worker/otel_transform.py` (~lines 425-456) to extract cache token counts from the OTel attribute bag before computing cost. The values are already preserved as metadata JSON strings further down (lines 478-485) — pull them earlier.
- OTel attribute names to read (cover both provider conventions):
  - Anthropic: `gen_ai.usage.cache_read_input_tokens`, `gen_ai.usage.cache_creation_input_tokens`
  - OpenAI: `gen_ai.usage.input_cached_tokens`
  - Fallback: `gen_ai.usage.details.cache_read_tokens`, `gen_ai.usage.details.cache_write_tokens`

### Acceptance

- A Claude trace with non-zero cache_read tokens produces a `spans.cost` that includes the cache_read contribution.
- Existing unit tests in `backend/worker/tests/` updated; one new test case for a cache-heavy fixture.
- The two cost-calc readers (TS `lookup.ts` and Python `pricing.py`) produce the same cost for the same input. They are kept as separate readers by design — ingest is Python, internal agent path is TS — but both consume the same `standard-model-prices.json`.

ka1kqi:
Fixes #956

## Summary

At ingest, span cost only applied the `input` and `output` per-token rates —
cache read/write tokens were never priced at their own rates, so cache-heavy
traces were mis-costed. This normalizes each LLM span's gross (cache-inclusive)
token count into disjoint buckets — uncached input / output / cache read /
cache write — and prices each bucket exactly once at its own rate, so cost
reflects the cache discount instead of dropping it on the floor.

## Type of Change

- [x] fix - A bug fix

(Also includes a small refactor — consolidating cost onto one helper — and new
tests, but the user-facing change is the cost fix.)

## Details

- **`tokens/buckets.py` (new) — `normalize_token_usage`**: converts gross
  instrumentor counts into disjoint buckets, keyed on the OTel instrumentation
  *scope* (the convention is inclusive: cache is a breakdown of input, per the
  GenAI semconv, so cache is subtracted from input). Reconciliation invariant:
  `uncached + cache_read + cache_write == gross_input`.
- **`tokens/pricing.py` — `cost_from_buckets`**: single source of truth for
  cost; prices each disjoint bucket once (`input`/`output`/`cacheRead`/
  `cacheWrite`). `calculate_cost` now routes through it, so there's one cost
  formula across paths.
- **`otel_transform.py`**: at ingest, reads the cache read/write token
  attributes (both the OpenInference `llm.token_count.prompt_details.*` and
  GenAI `gen_ai.usage.cache_*` key families), normalizes, and prices
  cache-aware. Stored token columns stay gross for display continuity — only
  cost changes.

## How validated

- **186 unit tests pass** across `tests/worker/tokens/test_buckets.py`,
  `test_pricing.py`, and `tests/worker/test_otel_transform.py` — covering the
  reconciliation invariant, monotonicity (a cache-read call costs *less* than
  the same call uncached), no double-counting, and the multiple attribute-key
  families / provider conventions.
- pre-commit (ruff + format) passes.
- Manually verified the bucketed split and resulting cost against real cached
  OpenAI (`gpt-4o`) and Anthropic (`claude-sonnet-4-5`) calls; cache read and
  cache write are each priced once and reconcile to the gross input.

## Screenshots / Recordings
deepagents run
Screenshot 2026-06-03 at 4 42 47 PM

anthropic run
Screenshot 2026-06-03 at 4 40 44 PM

---
## Summary by cubic
Fixes LLM span cost by pricing cache read/write tokens at ingest using scope‑aware, disjoint buckets, with cache uncapped so net/exclusive emitters (e.g. claude‑agent‑sdk) are priced correctly. Stored token columns remain gross; only cost changes.

- **Bug Fixes**
  - Subtract cache read/write from input before pricing, with uncached = max(input − cache_read − cache_write, 0); keep cache buckets uncapped and non‑negative. Price each bucket once via `normalize_token_usage` + `cost_from_buckets`.
  - Read cache details from OpenInference `llm.token_count.prompt_details.*` and GenAI `gen_ai.usage.*` keys, including pydantic‑ai variants; treat `@arizeai/openinference*` scopes as cache‑inclusive.
  - Use `first_present_number` and `int_or_zero` to skip malformed numeric attrs so valid fallbacks are used; warn once per unknown scope, guard non‑string `scope.name` values, and bound the dedup set.

- **Refactors**
  - Added `tokens/buckets.py` for scope‑keyed normalization; `cost_from_buckets` is the single source of cost math. Routed `calculate_cost` through it (cache buckets zero on text estimation).

<sup>Written for commit d8a299e2a300fd0cab5462e39299d25a70c054e0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1022?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/worker/otel_transform.py
+++ b/backend/worker/otel_transform.py
@@ -493,8 +493,8 @@ def transform_otel_to_clickhouse(
                         span_record["output_tokens"] = output_tokens
                         span_record["total_tokens"] = total_tokens
 
-                        # Cost: normalize gross counts into disjoint buckets keyed on
-                        # the instrumentation scope, then price each bucket once.
+                        # Cost: normalize counts into disjoint buckets keyed on the
+                        # instrumentation scope, then price each bucket once.
                         from worker.tokens.buckets import normalize_token_usage
                         from worker.tokens.pricing import (
                             cost_from_buckets,
--- a/backend/worker/otel_transform.py
+++ b/backend/worker/otel_transform.py
@@ -84,6 +84,27 @@ def first_present(attrs: dict[str, Any], keys: list[str]) -> Any:
     return None
 
 
+def first_present_number(attrs: dict[str, Any], keys: list[str]) -> Any:
+    """Like ``first_present`` but for numeric token attributes: skip keys whose
+    value is missing or malformed (empty / non-numeric) so a malformed
+    high-priority attribute cannot suppress a valid lower-priority fallback.
+
+    Validity mirrors ``int_or_zero`` (the value must coerce via ``int``); a present
+    but unusable value is skipped rather than short-circuiting the candidate list.
+    Returns the first usable value, or None if none qualify.
+    """
+    for key in keys:
+        value = attrs.get(key)
+        if value is None or value == "":
+            continue
+        try:
+            int(value)
+        except (TypeError, ValueError):
+            continue
+        return value
+    return None
+
+
 def int_or_zero(value: Any) -> int:
     """Convert a present OTEL numeric attribute to int; missing/invalid -> 0.
 
@@ -409,7 +430,7 @@ def transform_otel_to_clickhouse(
 
                     # Try API-provided token counts first (from instrumentors).
                     # OpenInference: llm.token_count.*  ·  GenAI semconv: gen_ai.usage.*
-                    api_input_tokens = first_present(
+                    api_input_tokens = first_present_number(
                         span_attrs,
                         [
                             "llm.token_count.prompt",
@@ -417,7 +438,7 @@ def transform_otel_to_clickhouse(
                             "gen_ai.usage.prompt_tokens",
                         ],
                     )
-                    api_output_tokens = first_present(
+                    api_output_tokens = first_present_number(
                         span_attrs,
                         [
                             "llm.token_count.completion",
@@ -425,14 +446,14 @@ def transform_otel_to_clickhouse(
                             "gen_ai.usage.completion_tokens",
                         ],
                     )
-                    api_total_tokens = first_present(
+                    api_total_tokens = first_present_number(
                         span_attrs,
                         ["llm.token_count.total", "gen_ai.usage.total_tokens"],
                     )
                     # Cache buckets. The OpenInference keys (prompt_details.*) are the
                     # verified path for Anthropic/OpenAI and MUST be listed first —
                     # they are the same family as llm.token_count.prompt (read above).
-                    api_cache_read_tokens = first_present(
+                    api_cache_read_tokens = first_present_number(
                         span_attrs,
                         [
                             "llm.token_count.prompt_details.cache_read",
@@ -445,7 +466,7 @@ def transform_otel_to_clickhouse(
                             "gen_ai.usage.details.cache_read_input_tokens",
                         ],
                     )
-                    api_cache_write_tokens = first_present(
+                    api_cache_write_tokens = first_present_number(
                         span_attrs,
                         [
                             "llm.token_count.prompt_details.cache_write",
--- a/backend/worker/otel_transform.py
+++ b/backend/worker/otel_transform.py
@@ -558,7 +558,7 @@ def transform_otel_to_clickhouse(
                         trace_attrs[trace_id]["session_id"] or span_session_id
                     )
 
-                # Eager trace creation (Langfuse-style):
+                # Eager trace creation:
                 # Create a "shallow" trace record on the FIRST span we see for
                 # a trace_id, so it appears in the UI immediately. When the root
                 # span arrives later, upgrade to a "full" trace with rich metadata.
--- a/backend/worker/otel_transform.py
+++ b/backend/worker/otel_transform.py
@@ -85,8 +85,19 @@ def first_present(attrs: dict[str, Any], keys: list[str]) -> Any:
 
 
 def int_or_zero(value: Any) -> int:
-    """Convert a present OTEL numeric attribute to int; missing -> 0."""
-    return int(value) if value is not None else 0
+    """Convert a present OTEL numeric attribute to int; missing/invalid -> 0.
+
+    ``first_present`` returns falsy-but-present values (e.g. an empty string for
+    an attribute that exists with a non-numeric value), so guard the cast — a
+    single malformed attribute must not crash ingestion of the whole batch.
+    """
+    if value is None or value == "":
+        return 0
+    try:
+        return int(value)
+    except (TypeError, ValueError):
+        logger.warning("Non-numeric OTEL token attribute %r; treating as 0", value)
+        return 0
 
 
 def decode_otel_id(b64_value: str | None) -> str | None:
--- a/backend/worker/otel_transform.py
+++ b/backend/worker/otel_transform.py
@@ -440,6 +440,9 @@ def transform_otel_to_clickhouse(
                             "gen_ai.usage.cache_read_input_tokens",
                             "gen_ai.usage.input_cached_tokens",
                             "gen_ai.usage.details.cache_read_tokens",
+                            # pydantic-ai version variants (names differ by release):
+                            "gen_ai.usage.cache_read_tokens",
+                            "gen_ai.usage.details.cache_read_input_tokens",
                         ],
                     )
                     api_cache_write_tokens = first_present(
@@ -449,6 +452,8 @@ def transform_otel_to_clickhouse(
                             "gen_ai.usage.cache_creation.input_tokens",
                             "gen_ai.usage.cache_creation_input_tokens",
                             "gen_ai.usage.details.cache_write_tokens",
+                            # pydantic-ai version variant:
+                            "gen_ai.usage.details.cache_creation_input_tokens",
                         ],
                     )
 
--- a/backend/worker/otel_transform.py
+++ b/backend/worker/otel_transform.py
@@ -84,6 +84,11 @@ def first_present(attrs: dict[str, Any], keys: list[str]) -> Any:
     return None
 
 
+def int_or_zero(value: Any) -> int:
+    """Convert a present OTEL numeric attribute to int; missing -> 0."""
+    return int(value) if value is not None else 0
+
+
 def decode_otel_id(b64_value: str | None) -> str | None:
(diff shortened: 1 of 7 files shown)


ka1kqi:
@cubic-dev-ai review this PR — all three findings from the previous review are addressed in 1c54bab9:

- **P1** (buckets.py): cache buckets are now capped to the gross input, so cache counts exceeding input can no longer overbill or break the reconciliation invariant.
- **P2** (otel_transform.py): `int_or_zero` guards present-but-non-numeric attributes (e.g. empty string) instead of crashing ingestion.
- **P2** (buckets.py): the unknown-scope warning dedup set is now bounded.

Added tests for each; full worker suite (194) passes.

ka1kqi:
@cubic-dev-ai review this PR again, please.

XinweiHe:
A few points before merge:

1. **Mastra cache detection** — Mastra runs on the Vercel AI SDK, which emits spans under the `ai` scope and reports cache token information inside the `ai.response.providerMetadata` JSON blob rather than the flat `llm.token_count.prompt_details.*` / `gen_ai.usage.cache_*` attributes this PR currently reads. I also don't see an `ai`/`mastra` entry in `_SCOPE_CONVENTIONS`. Can you confirm what `@traceroot-ai/mastra`'s exporter actually emits? If AI SDK-native attributes are passed through unchanged, cached Mastra traces will not have cache usage detected and will be overcharged. That would leave the exact bug this PR is intended to fix unresolved for one of our own first-party integrations. Since Mastra is a documented integration, this should be addressed before merge.
This could be either resolved here or in an immediate follow-up PR is fine as well.

2. **pydantic-ai cache key coverage** — The cache-related attribute names vary across pydantic-ai versions. We should also recognize `gen_ai.usage.cache_read_tokens` (without `.details`) as well as `gen_ai.usage.details.cache_read_input_tokens` and `gen_ai.usage.details.cache_creation_input_tokens`. Otherwise, cache usage from some pydantic-ai versions will be silently missed.

3. **Simplify `buckets.py`** — Consider shortening the opening docstring and removing the `EXCLUSIVE` branch. The comment already notes that there is no known emitter for it, so it appears to be speculative generality. Dropping it would reduce both code and documentation complexity.

4. **Default values for `TokenBuckets`** — Consider giving all `TokenBuckets` fields default values of `0`. Since nothing is persisted, there are no compatibility concerns. Defaults would also make future additions (e.g. reasoning or audio token types) purely additive changes rather than requiring updates to every constructor call site.

ka1kqi:
Thanks @XinweiHe — concrete plan for each point:

**2. pydantic-ai cache key coverage** *(this PR)*
Append the missing variants to the `first_present` lists in `otel_transform.py` (verified OpenInference `prompt_details.*` keys stay first):
- `api_cache_read_tokens` += `gen_ai.usage.cache_read_tokens`, `gen_ai.usage.details.cache_read_input_tokens`
- `api_cache_write_tokens` += `gen_ai.usage.details.cache_creation_input_tokens`

Add a transform test asserting cache is picked up from each new variant (read + write).

**3. Simplify `buckets.py`** *(this PR)*
Remove the `InputConvention.EXCLUSIVE` member and the `else` branch in `normalize_token_usage` — it always subtracts cache from the gross input now. Collapse the opening docstring to ~3 lines (gross input → disjoint buckets, each priced once). Keep the warn-once on an unrecognized scope (reworded to "assuming cache-inclusive"), since it still surfaces a new emitter to verify. Drop the EXCLUSIVE case from `test_buckets`.

**4. `TokenBuckets` default values** *(this PR)*
Default all four fields to `0` (`input_uncached: int = 0`, `output: int = 0`, `cache_read: int = 0`, `cache_write: int = 0`) so future token types (reasoning/audio) are purely additive. No call-site changes needed.

**1. Mastra cache detection → follow-up PR.**
Confirmed: there's no `ai`/`mastra` entry in `_SCOPE_CONVENTIONS`, and the worker reads none of the Vercel AI SDK attributes (`ai.usage.*`) nor parses `ai.response.providerMetadata`. I'll confirm what `@traceroot-ai/mastra`'s exporter actually emits first; if it passes the AI-SDK-native attrs through unchanged, the follow-up will parse cache from `providerMetadata` (`cacheReadInputTokens` / `cacheCreationInputTokens`) and add an `ai`/`mastra` scope-convention entry so cached Mastra traces stop being overcharged. **Tracking this as a separate PR per your note** — keeping it out of this one to keep the diff focused.

ka1kqi:
@cubic-dev-ai review this PR again — pushed `d294d66c` addressing the pre-merge feedback:

- **(2)** `otel_transform.py`: read additional pydantic-ai cache key variants (`gen_ai.usage.cache_read_tokens`, `gen_ai.usage.details.cache_read_input_tokens`, `gen_ai.usage.details.cache_creation_input_tokens`) so cache isn't silently missed on some versions.
- **(3)** `buckets.py`: dropped the speculative `EXCLUSIVE` branch + collapsed the single-convention machinery; shortened the docstring.
- **(4)** `TokenBuckets` fields now default to `0`.
- Also added an `@arizeai/openinference` scope prefix so JS/TS OpenInference spans match explicitly as cache-inclusive.

208/208 worker tests pass. Mastra (1) remains a tracked follow-up PR.

ka1kqi:
@cubic-dev-ai re-review please — `06ce79b8` addresses your finding on `otel_transform.py:412`.

Root cause confirmed: `first_present` returns the first key that is `is not None`, so a present-but-malformed high-priority token attribute (empty string / non-numeric) short-circuited the candidate list and `int_or_zero` turned it into 0 — silently undercounting usage/cost even when a valid lower-priority fallback existed.

Fix: added `first_present_number` (validity mirrors `int_or_zero`: must coerce via `int`) and used it for the numeric token reads (input/output/total/cache_read/cache_write). String reads (span input/output) keep `first_present`. A valid `0` still wins; only missing/malformed values are skipped. Added unit + transform-level tests for the malformed-high-priority-+-valid-fallback case. 213/213 worker tests pass.

ka1kqi:
## Validated `018965d3` (clamp-free cache) against ground truth — 8 examples, Python + TypeScript

A ground-truth tracer captures the raw provider `usage` per call **and** the exact OTel attributes each instrumentor emits, then prices them through the worker's real `buckets.py`.

| Example | Lang | Ground truth | TraceRoot |
|---|---|---|---|
| anthropic (cached) | py | $0.030943 | $0.030943 ✅ |
| openai | py | $0.000002 | $0.000002 ✅ |
| pydantic-ai (cached) | py | $0.005034 | $0.005034 ✅ |
| langchain (cached) | py | $0.005034 | $0.005034 ✅ |
| **openai** | **ts** | $0.000002 | $0.000002 ✅ |
| **langchain (cached)** | **ts** | $0.005034 | $0.004986 ✅ (−1%) |
| claude-agent-sdk | py | (see note) | priced (was −61%) |
| anthropic | **ts** | $0.030943 | $0.000528 ⚠️ |

**Results**
- **Matches ground truth to the cent** on all well-instrumented emitters, cached and uncached.
- **The clamp-free rule handles every net/exclusive emitter without a scope list** — confirmed on claude-agent-sdk (py) **and** the JS langchain instrumentor (ts), both of which emit a *net* prompt + additive cache (e.g. `prompt=8`, `cache_read=7510`). The floor-and-uncapped-cache rule — `input = max(prompt − cache_read − cache_write, 0)` with cache kept uncapped — prices these correctly; a hardcoded exclusive-scope list would have had to enumerate each one.

**Two non-blocking notes (not normalization issues)**
- **claude-agent-sdk**: GT can't be precisely anchored (the SDK aggregates usage across its internal agent loop differently than the single emitted span), but TraceRoot now *prices* the cache (was clamped to ~2 tokens, −61%).
- **TS anthropic** (`@arizeai/openinference-instrumentation-anthropic@0.1.7`, the version bundled by the TS SDK): this JS instrumentor **drops the cache attributes entirely** (emits only a net `prompt`), so the cache cost is undercounted because the data isn't in the span. Upstream instrumentor gap (JS is behind Python); follow-up is to bump that instrumentor, not change normalization.

@cubic-dev-ai please re-review. `018965d3` replaces the `min(cache, input)` cap with the clamp-free rule. Re: the earlier P1 ("cache exceeds input"): that cap was mis-clamping *legitimate* net-input emitters (claude-agent-sdk, JS langchain) down to ~0 — the floor + uncapped-cache rule is the correct generalization, verified against ground truth above. 213 worker tests pass; lint clean.

ka1kqi:
@cubic-dev-ai fixed in `d8a299e2` — valid catch. `scope_name` is typed `str | None` but comes from untrusted OTLP (`scope.name`), so a malformed non-string would crash `scope_name.lower()` and take down the batch. Now guarded with `isinstance(scope_name, str)` (a non-string scope is treated as unknown), consistent with the `int_or_zero` 'never crash ingestion on malformed data' hardening. Added a regression test (`test_non_string_scope_does_not_crash`); 214 worker tests pass.

PR #1120, opened 2026-06-09
fix(worker): surface git metadata during live traces

### LINKED ISSUE #1099: Trace-level git_repo/git_ref only populate from the root span (appear late in live streaming)
## Summary

Trace-level `git_repo` / `git_ref` only populate **after the root span arrives** (i.e. at the very end of a live run), while every other span in the trace already carries the correct values from the start. During live trace streaming this makes the GitHub repo/ref appear to "fill in at the last moment," unlike the trace **name**, which shows up live.

This is **not** an SDK regression — `traceroot-ts` has stamped `traceroot.git.repo` / `traceroot.git.ref` on **every span** at `onStart` since v0.1.0, unchanged through 0.1.6 (0.1.6 only changed git-context *resolution sources*, not when attributes are attached). The asymmetry is purely in the backend transform.

## Root cause

`backend/worker/otel_transform.py` (~line 649) extracts trace-level git fields **only from the root span**:

```python
if not parent_span_id:          # only the root span
    git_ref  = span_attrs.get("traceroot.git.ref")
    git_repo = span_attrs.get("traceroot.git.repo")
    ...
    traces[trace_id]["git_ref"]  = git_ref
    traces[trace_id]["git_repo"] = git_repo
```

The root span wraps the whole run, so it **ends last** and — with batch export — **arrives last** in the live stream. Therefore trace-level repo/ref only populate on completion.

This gate has existed since `6012a860` (2026-02-28), so it is long-standing, not new.

## Why the trace name doesn't have this problem

The identical "root arrives last" issue was already solved for the trace **name** in `3bd7e598` (2026-04-20) via a shallowest-`traceroot.span.path[0]` candidate system (`otel_transform.py` ~lines 633-647), which recovers the name from child spans before the root lands. Git repo/ref never received the same treatment — even though every child span already carries them.

## Proposed fix

Backend-only (the data is already present on every span): lift `git_ref` / `git_repo` extraction out of the `if not parent_span_id` block and populate from the first span that carries them (idempotent — values are identical across all spans in a trace), mirroring the existing name-candidate pattern.

## Acceptance criteria

- [ ] In a batch where child spans arrive before the root, `git_repo` / `git_ref` populate from a child span.
- [ ] Root span value still wins / remains consistent (no overwrite churn).
- [ ] Regression test: child-before-root batch yields populated repo/ref without the root present.

## Notes

- SDK side needs no change.
- Repro: run any agent example that wraps work in a root `observe(...)` span (e.g. `examples/typescript/claude-agent-sdk`) and watch the trace live — name appears immediately, repo/ref appear only when the run finishes.

RitwijParmar:
Fixes #1099.

## What changed
- Promote `traceroot.git.repo` and `traceroot.git.ref` from the first span that carries them, so live trace records have GitHub context before the root span finishes.
- Keep root span values authoritative when they arrive, while preserving child values if an older or partial root span does not include git context.
- Add regression coverage for child-before-root live batches, root override, and missing-root-git cases.

## Why
The SDK already stamps git repo/ref on every span. The backend only copied those fields from the root span, which usually arrives last in live streaming. That made repo/ref appear only at the end of a run even though the data was already present in child spans.

## Validation
- `/Users/ritwij/.cache/codex-runtimes/codex-primary-runtime/dependencies/python/bin/python3 -m pytest tests/worker/test_otel_transform.py tests/worker/test_otel_transform_metadata.py`
- `/Users/ritwij/.cache/codex-runtimes/codex-primary-runtime/dependencies/python/bin/python3 -m ruff check backend/worker/otel_transform.py tests/worker/test_otel_transform.py`
- `git diff --check`

### DIFF
--- a/backend/worker/otel_transform.py
+++ b/backend/worker/otel_transform.py
@@ -306,6 +306,7 @@ def transform_otel_to_clickhouse(
     # Shortest ids_path = closest to root. Used to correct eager trace names
     # when the first span in a batch isn't the closest-to-root span for that trace.
     _trace_name_candidates: dict[str, tuple[int, str]] = {}
+    trace_git_attrs: dict[str, dict[str, str | None]] = {}
 
     # camelCase: resourceSpans
     resource_spans = otel_data.get("resourceSpans", [])
@@ -587,9 +588,13 @@ def transform_otel_to_clickhouse(
                 # Priority: root span values overwrite, child span values only set if empty
                 span_user_id = _extract_user_id(span_attrs)
                 span_session_id = _extract_session_id(span_attrs)
+                span_git_ref = span_attrs.get("traceroot.git.ref")
+                span_git_repo = span_attrs.get("traceroot.git.repo")
 
                 if trace_id not in trace_attrs:
                     trace_attrs[trace_id] = {"user_id": None, "session_id": None}
+                if trace_id not in trace_git_attrs:
+                    trace_git_attrs[trace_id] = {"git_ref": None, "git_repo": None}
 
                 if not parent_span_id:
                     # Root span: always use its values if present (overwrites child values)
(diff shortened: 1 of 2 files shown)


dark-sorceror:
Closed since #1134 contains merged fix

Issue #1006, opened 2026-06-01
Live trace view mis-signals completion when a root span ends before its descendants


XinweiHe:
## Summary
In the live trace view, traces whose **root span finishes before its descendants** can appear stuck on **"live"** with a duration that grows against `now()`, then snap to the correct value on refresh. Happens with streaming HTTP handlers that return their response while background work continues, and with long-lived agent sessions.

**Severity: minor.** The data is always correct in ClickHouse; a page refresh recovers the right view. Same root-ends-last assumption as the timeline-duration bug (#1003 / #1004) — this is its live-path sibling.

## Root cause
Completion is inferred from **"the root span has an `end_time`"**:
- `backend/worker/ingest_tasks.py` `_publish_live_spans` emits `trace_complete` as soon as an ingested batch contains a root span (no parent) with an `end_time`.
- `backend/rest/routers/live.py` `_is_trace_complete_in_clickhouse` declares completion on the same condition.

When the root ends *first*, it lands in the **first** ingest batch (~5-7s in), so `trace_complete` fires while dozens of descendants are still arriving. The frontend (`use-trace-stream.ts`) tears down the SSE stream on that signal and does a single refetch onto a *partial* trace; its orphaned descendants render as null-end pending placeholders whose duration ticks against `now()`, and the late spans never load until a manual refresh.

## Evidence
Repro trace (59 spans): real extent **10.7s**, ClickHouse fully consistent within ~0s of completion (not a data lag). Redis live channel during one run:
```
spans          <- first batch (contains root)
trace_complete <- fires here, ~7s in, because the batch had the root span
spans          <- 44 more spans still arriving afterward
```

## Recommended direction (platform-side)
A trace is never definitively "complete" in OpenTelemetry — it is monotonically growing and eventually consistent. Embrace that instead of pinning completion to the root span:
1. Stop treating "root span has `end_time`" as completion.
2. Keep `trace.end_time = max(span.end_time)`, updated as spans arrive (the trace-list query already does this).
3. Drive the "live" visual state from pending-span resolution (we already emit `traceroot.span.ids_path`); bound pending placeholders to their children's real times so the duration does not run against `now()`.
4. Use a quiet-window/TTL only for SSE *lifecycle* (when to stop streaming), never as ground truth for "complete."

[...]

PR #1134, opened 2026-06-10
fix(live): keep trace git metadata and late spans visible

RitwijParmar:
References #1099 and #1006.

This fixes two live-trace edge cases that come from treating the root span as the whole trace lifecycle.

What changed

- Promote `traceroot.git.repo` and `traceroot.git.ref` from child spans into the trace record as soon as they arrive, instead of waiting for the root span. The root span still wins if it carries a value later.
- Keep the live SSE stream open for a short quiet window after a root-complete signal. This lets late descendant span batches arrive before the frontend receives `trace_complete` and closes the stream.
- Preserve the subscribe-before-ClickHouse-check invariant so concurrent Redis span events are still not missed.
- Add regression coverage for child-before-root git metadata, root-over-child precedence, root-without-git fallback, already-complete quiet completion, and late spans arriving after `trace_complete`.

Why this matters

For live agent runs, the trace can be useful before the root span lands or before all descendants have arrived. This makes repo/ref visible earlier for debugging and prevents the UI from closing the stream while the trace is still growing.

Validation

- `uv run ruff check backend/rest/routers/live.py backend/worker/otel_transform.py tests/rest/test_live_router.py tests/worker/test_otel_transform.py`
- `uv run python -m py_compile backend/rest/routers/live.py backend/worker/otel_transform.py tests/rest/test_live_router.py tests/worker/test_otel_transform.py`
- `uv run pytest tests/rest/test_live_router.py tests/worker/test_otel_transform.py -q` passed, 98 tests
- `git diff --check`

---
## Summary by cubic
Improve live trace reliability by surfacing git repo/ref early and holding the SSE stream open for a short quiet window after completion. The window is anchored to the latest of the root’s end time and the last ingest activity so late spans are delivered while old completed traces close immediately.

- **Bug Fixes**
  - Promote `traceroot.git.repo` and `traceroot.git.ref` from child spans into the trace as they arrive; root values overwrite, and child values persist if the root has none.
  - Emit `trace_complete` after a quiet window triggered by ClickHouse root completion or a Redis `trace_complete`; reset on late `spans` or redundant `trace_complete`; anchor the deadline to max(root end, last ingest) and close immediately if it’s already past; subscribe to Redis before the ClickHouse check and forward spans received during the check/window before closing; at the hard ceiling emit `trace_complete` if the root completed, otherwise `stream_timeout`; the window length is `settings.trace_complete_quiet_seconds` (default 10s, above the SDK flush interval).

<sup>Written for commit ce0261fa20dfe4ef0acede9a407ea6419f846a43. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1134?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/rest/routers/live.py
```diff
@@ -72,50 +78,57 @@ async def event_generator():
...
-                        break
+                    if event_type == "spans" and completion_deadline is not None:
+                        completion_deadline = time.monotonic() + TRACE_COMPLETE_QUIET_SECONDS
                 else:
+                    if completion_deadline is not None and time.monotonic() >= completion_deadline:
+                        yield "event: trace_complete\ndata: {}\n\n"
+                        return
                     yield ": heartbeat\n\n"
             else:
                 yield "event: stream_timeout\ndata: {}\n\n"
```
dark-sorceror: Emit `trace_complete` here instead of `stream_timeout`

backend/rest/routers/live.py
```diff
@@ -72,50 +78,57 @@ async def event_generator():
...
-                        data = json.loads(msg["data"])
-                        if data.get("type") == "spans":
-                            yield f"event: spans\ndata: {msg['data']}\n\n"
-
-                yield "event: trace_complete\ndata: {}\n\n"
-                return
-
-            # Live trace: stream normally until trace_complete or timeout.
+            completion_deadline = (
+                time.monotonic() + TRACE_COMPLETE_QUIET_SECONDS if already_complete else None
```
dark-sorceror: Let's anchor to root span's end time, so opening a old trace doesn't hold the SSE connection + Redis subscription. Something like:
```python
completion_deadline = monotonic() + max(0, QUIET - age)
```

backend/rest/routers/live.py
```diff
@@ -20,10 +20,16 @@
 
 HEARTBEAT_INTERVAL = 15  # seconds
 MAX_STREAM_SECONDS = 600  # 10 minutes — hard ceiling for idle connections
+TRACE_COMPLETE_QUIET_SECONDS = 5
```
dark-sorceror: Make this settings/env-configurable and default to a slightly higher value like 10s since the SDK's default flush interval is 5 seconds

dark-sorceror (changes requested):

dark-sorceror:
I also noticed you had a similar PR open for git half (https://github.com/traceroot-ai/traceroot/pull/1120). Looks identical, so let's keep this one and close the other when this is merged.

backend/rest/routers/live.py
dark-sorceror: Resolved in 7f2eb57

backend/rest/routers/live.py
dark-sorceror: Resolved in 7f2eb57

backend/rest/routers/live.py
dark-sorceror: Resolved in 7f2eb57

dark-sorceror:
@cubic-dev-ai review

dark-sorceror:
LGTM

XinweiHe (approved):

PR #1146, opened 2026-06-10
fix(ui): align row ordering between token and cost breakdowns

### LINKED ISSUE #1141: fix(ui): align row ordering between token and cost breakdowns
## Problem

The token usage breakdown and the cost breakdown render their input sub-rows in different orders:

- `frontend/ui/src/features/traces/components/TokenUsageBreakdown.tsx` renders **cache read → cache write → uncached**
- `frontend/ui/src/features/traces/components/CostBreakdown.tsx` renders **uncached input → cache read → cache write**

Since the two panels show the same conceptual split (uncached input vs. cache read vs. cache write, once as tokens and once as dollars), the mismatched ordering makes them harder to compare side by side.

## Expected

Both breakdowns should enforce the same strict row ordering for the input split, e.g. uncached → cache read → cache write (or whichever canonical order we pick), applied consistently in both components.

## Affected code

- `TokenUsageBreakdown.tsx` — input section rows
- `CostBreakdown.tsx` — input cost rows

ka1kqi:
## Summary
- Reorder the token usage breakdown's input rows to uncached → cache read → cache write, matching the cost breakdown
- Normalize the row label too: the cost panel called the row "uncached input" while the token panel says "uncached"; both now use "uncached" (the section header already says Input)
- Add a server-render vitest test pinning the shared row order and label in both panels so they can't drift apart again

## Test plan
- `pnpm exec vitest run` in `frontend/ui` (new ordering test + existing suites; unrelated pre-existing flakes aside, all green)
- `pnpm lint` and `pnpm exec tsc --noEmit` clean
- Verified visually against a seeded trace on the local stack with distinct uncached / cache read / cache write values on both panels

## Video

https://github.com/user-attachments/assets/f2ea7388-1a5c-4dbd-a23d-11d21e54fdef

closes #1141

### DIFF
--- a/frontend/ui/src/features/traces/components/CostBreakdown.tsx
+++ b/frontend/ui/src/features/traces/components/CostBreakdown.tsx
@@ -33,7 +33,7 @@ export function CostBreakdown({ details }: CostBreakdownProps) {
         <span className="tabular-nums">{formatCost(c.inputCost)}</span>
       </div>
       <div className="mt-1 space-y-0.5">
-        <Row label="uncached input" value={c.inputUncachedCost} />
+        <Row label="uncached" value={c.inputUncachedCost} />
         <Row label="cache read" value={c.cacheReadCost} />
         <Row label="cache write" value={c.cacheWriteCost} />
       </div>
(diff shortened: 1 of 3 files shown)


XinweiHe:
Need a screenshot.

XinweiHe (approved):

PR #1183, opened 2026-06-13
fix(ui): keyboard accessibility — Escape closes trace panel, autoFocus on dialogs

### LINKED ISSUE #1179: feat(ui): keyboard accessibility — close the trace detail panel with Escape, confirm dialogs handle Escape/Enter
## Problem

Keyboard-driven navigation has gaps — the main one being the trace detail panel, which can only be dismissed with the mouse.

### 1. Trace detail panel can't be closed with Escape

`TraceViewerPanel` (`frontend/ui/src/features/traces/components/TraceViewerPanel.tsx`), the slide-in panel opened from the trace list on `/projects/{projectId}/traces`, is a custom fixed-position overlay — not a dialog primitive — so it gets no keyboard handling for free. The only close affordance is the X button.

**Expected:** pressing Escape closes the panel and returns to the trace list, matching the X button behavior.

Implementation notes:
- Add a `keydown` listener (registered while the panel is open) that triggers the same close path as the X button.
- If a nested popover/dropdown/dialog is open inside the panel, Escape should close that innermost layer first, not the whole panel.
- Don't hijack Escape while focus is in an element that uses it (e.g. an open select).

### 2. Project / workspace creation dialogs: Escape to dismiss, Enter to submit

`CreateProjectDialog` (`frontend/ui/src/features/projects/components/CreateProjectDialog.tsx`) and `CreateWorkspaceDialog` (`frontend/ui/src/features/workspaces/components/CreateWorkspaceDialog.tsx`) should both:
- close on Escape, and
- submit on Enter while the name input is focused.

Per the code this should already work — both use the Radix-based `Dialog` (Escape closes natively; nothing overrides `onEscapeKeyDown`) and wrap their content in `<form onSubmit>` with a `type="submit"` button, so Enter submits. As part of this issue, verify both behaviors in the running app and fix any that don't hold in practice (e.g. focus not landing in the input on open).

## Acceptance criteria

- [ ] Escape closes the trace detail panel; trace list state (filters, scroll) is preserved.
- [ ] Escape inside the panel closes an open nested layer (popover/select/dialog) before closing the panel itself.
- [ ] Project creation dialog: Escape dismisses, Enter in the name field creates the project (verified in-app).
- [ ] Workspace creation dialog: Escape dismisses, Enter in the name field creates the workspace (verified in-app).
- [ ] Tests cover the new Escape handling on the trace detail panel.

anishkumar0504:
Fixes #1179

Changes:
- `TraceViewerPanel`: adds `keydown` listener that calls `onClose` on Escape, yields to nested dialogs/popovers/selects first
- `CreateProjectDialog` + `CreateWorkspaceDialog`: adds `autoFocus` to name input so Enter-to-submit works on open
- `TraceViewerPanel.test.tsx`: two new tests covering Escape closes panel and Escape inside dialog does not close panel

Tested manually. CI will verify tests.

---
## Summary by cubic
Improves keyboard accessibility: Escape now closes the trace and session detail panels, and create dialogs auto-focus the name field so Enter submits (addresses #1179).

- **Bug Fixes**
  - Escape closes TraceViewerPanel and SessionDetailPanel; ignored when `defaultPrevented` (e.g., inside dialogs/popovers/menus).
  - Create Project/Workspace dialogs: `autoFocus` the name input for immediate typing and Enter-to-submit.
  - Tests cover Escape-to-close with `defaultPrevented` guard and autofocus on create dialogs.

<sup>Written for commit 33a125f89023fb903a43ecd2b6425f1b607332fb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1183?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/ui/src/features/projects/components/CreateProjectDialog.tsx
+++ b/frontend/ui/src/features/projects/components/CreateProjectDialog.tsx
@@ -76,6 +76,7 @@ export function CreateProjectDialog({
           </DialogHeader>
           <div className="py-4">
             <Input
+              autoFocus
               placeholder="Project name"
               value={name}
               onChange={(e) => setName(e.target.value)}
--- a/frontend/ui/src/features/traces/components/TraceViewerPanel.tsx
+++ b/frontend/ui/src/features/traces/components/TraceViewerPanel.tsx
@@ -161,6 +161,25 @@ export function TraceViewerPanel({
     });
   }, [viewMode]);
 
+
+// #1179 — close panel with Escape key
+useEffect(() => {
+  const handleKeyDown = (e: KeyboardEvent) => {
+    if (e.key !== "Escape") return;
+    // Let nested popovers/selects/dialogs handle Escape first
+    const activeEl = document.activeElement;
+    if (
+      activeEl instanceof HTMLElement &&
+      activeEl.closest('[role="dialog"], [role="listbox"], [data-radix-popper-content-wrapper]')
+    ) {
+      return;
+    }
+    onClose();
+  };
+  document.addEventListener("keydown", handleKeyDown);
+  return () => document.removeEventListener("keydown", handleKeyDown);
+}, [onClose]);
+
   const handleToggleCollapse = useCallback((id: string) => {
     setCollapsedIds((prev) => {
       const next = new Set(prev);
--- a/frontend/ui/src/features/traces/components/TraceViewerPanel.tsx
+++ b/frontend/ui/src/features/traces/components/TraceViewerPanel.tsx
@@ -160,26 +160,25 @@ export function TraceViewerPanel({
       timelineScrollRef.current.scrollTop = treeScrollRef.current.scrollTop;
     });
   }, [viewMode]);
-
-
-// #1179 — close panel with Escape key
-useEffect(() => {
-  const handleKeyDown = (e: KeyboardEvent) => {
-    if (e.key !== "Escape") return;
-    // Let nested popovers/selects/dialogs handle Escape first
-    const activeEl = document.activeElement;
-    if (
-      activeEl instanceof HTMLElement &&
-      activeEl.closest('[role="dialog"], [role="listbox"], [data-radix-popper-content-wrapper]')
-    ) {
-      return;
-    }
-    onClose();
-  };
-  document.addEventListener("keydown", handleKeyDown);
-  return () => document.removeEventListener("keydown", handleKeyDown);
-}, [onClose]);
-
+  // Close the panel on Escape, unless a nested overlay (dialog/select/popover) owns focus and should consume it first.
+  useEffect(() => {
+    const handleKeyDown = (e: KeyboardEvent) => {
+      if (e.key !== "Escape") return;
+      // Let nested popovers/selects/dialogs handle Escape first
+      const activeEl = document.activeElement;
+      if (
+        activeEl instanceof HTMLElement &&
+        activeEl.closest(
+          '[role="dialog"], [role="listbox"], [role="menu"], [data-radix-popper-content-wrapper]',
+        )
+      ) {
+        return;
+      }
+      onClose();
+    };
+    document.addEventListener("keydown", handleKeyDown);
+    return () => document.removeEventListener("keydown", handleKeyDown);
+  }, [onClose]);
   const handleToggleCollapse = useCallback((id: string) => {
     setCollapsedIds((prev) => {
       const next = new Set(prev);
--- a/frontend/ui/src/features/traces/components/TraceViewerPanel.tsx
+++ b/frontend/ui/src/features/traces/components/TraceViewerPanel.tsx
@@ -160,20 +160,13 @@ export function TraceViewerPanel({
       timelineScrollRef.current.scrollTop = treeScrollRef.current.scrollTop;
     });
   }, [viewMode]);
-  // Close the panel on Escape, unless a nested overlay (dialog/select/popover) owns focus and should consume it first.
+  // Close the panel on Escape. A nested Radix overlay (dialog/select/popover/menu) that
+  // consumes the Escape calls preventDefault() in the capture phase, before this
+  // bubble-phase listener runs — so defaultPrevented means "already handled, leave the panel open".
   useEffect(() => {
     const handleKeyDown = (e: KeyboardEvent) => {
       if (e.key !== "Escape") return;
-      // Let nested popovers/selects/dialogs handle Escape first
-      const activeEl = document.activeElement;
-      if (
-        activeEl instanceof HTMLElement &&
-        activeEl.closest(
-          '[role="dialog"], [role="listbox"], [role="menu"], [data-radix-popper-content-wrapper]',
-        )
-      ) {
-        return;
-      }
+      if (e.defaultPrevented) return;
       onClose();
     };
     document.addEventListener("keydown", handleKeyDown);
--- a/frontend/ui/src/features/workspaces/components/CreateWorkspaceDialog.tsx
+++ b/frontend/ui/src/features/workspaces/components/CreateWorkspaceDialog.tsx
@@ -72,6 +72,7 @@ export function CreateWorkspaceDialog({
           </DialogHeader>
           <div className="py-4">
             <Input
+              autoFocus
               placeholder="Workspace name"
               value={name}
               onChange={(e) => setName(e.target.value)}
--- a/frontend/ui/src/features/projects/components/CreateProjectDialog.test.tsx
+++ b/frontend/ui/src/features/projects/components/CreateProjectDialog.test.tsx
@@ -1,0 +1,28 @@
+// @vitest-environment jsdom
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { render, cleanup } from "@testing-library/react";
+
+vi.mock("next/navigation", () => ({
+  useRouter: () => ({ push: vi.fn() }),
+}));
+
+vi.mock("@tanstack/react-query", () => ({
+  useMutation: () => ({ mutate: vi.fn(), isPending: false, isError: false }),
+  useQueryClient: () => ({ invalidateQueries: vi.fn() }),
+}));
+
+vi.mock("@/lib/api", () => ({ createProject: vi.fn() }));
+
+import { CreateProjectDialog } from "./CreateProjectDialog";
+
+afterEach(() => {
+  cleanup();
+});
+
+describe("CreateProjectDialog focus", () => {
+  it("autofocuses the name input when opened", () => {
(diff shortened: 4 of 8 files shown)


anishkumar0504:
All checks are passing and cubic-dev-ai's automated review found no issues. Ready for a maintainer review whenever you get a chance thanks!

frontend/ui/src/features/traces/components/TraceViewerPanel.tsx
```diff
@@ -161,6 +161,24 @@ export function TraceViewerPanel({
     });
   }, [viewMode]);
 
+  // #1179 — close panel with Escape key
```
ka1kqi: Let's drop the issue number from this comment — our convention is no issue/PR numbers in code comments (they're opaque to future readers and rot over time). Rewrite it as rationale only, e.g.:

```ts
// Close the panel on Escape, unless a nested overlay (dialog/select/popover) owns focus and should consume it first.
```

The `closes #1179` link can live in the PR description instead.

frontend/ui/src/features/traces/components/TraceViewerPanel.tsx
```diff
@@ -161,6 +161,24 @@ export function TraceViewerPanel({
...
+  useEffect(() => {
+    const handleKeyDown = (e: KeyboardEvent) => {
+      if (e.key !== "Escape") return;
+      // Let nested popovers/selects/dialogs handle Escape first
+      const activeEl = document.activeElement;
+      if (
+        activeEl instanceof HTMLElement &&
+        activeEl.closest('[role="dialog"], [role="listbox"], [data-radix-popper-content-wrapper]')
```
ka1kqi: Heads up on a Radix footgun worth a quick manual check (non-blocking): this guard inspects `document.activeElement` at keydown time, but for Radix `Select`/`DropdownMenu`/`Popover`, Escape can close the overlay and return focus to its trigger (which lives *inside* the panel) *before* this document-level listener runs — so the same Escape could also close the panel (double-close).

Could you manually verify that pressing Escape inside an open Select/dropdown within the trace panel closes only that overlay, not the whole panel? Also note the selector matches `role="listbox"` (Select) but not `role="menu"` (DropdownMenu) — dropdowns rely solely on the `data-radix-popper-content-wrapper` catch-all here.

frontend/ui/src/features/projects/components/CreateProjectDialog.tsx
```diff
@@ -76,6 +76,7 @@ export function CreateProjectDialog({
           </DialogHeader>
           <div className="py-4">
             <Input
+              autoFocus
```
ka1kqi, line 79: This `autoFocus` addition isn't covered by any test — our CI diff-coverage gate (≥80% on changed lines) will see it at 0%. A small jsdom test that renders the dialog open and asserts the name `Input` is `document.activeElement` would cover it and lock in the behavior.

frontend/ui/src/features/workspaces/components/CreateWorkspaceDialog.tsx
```diff
@@ -72,6 +72,7 @@ export function CreateWorkspaceDialog({
           </DialogHeader>
           <div className="py-4">
             <Input
+              autoFocus
```
ka1kqi, line 75: Same as the projects dialog — please add a small jsdom test asserting the name `Input` receives initial focus, so this `autoFocus` line is covered by the diff-coverage gate.

ka1kqi (commented):
Thanks for this @anishkumar0504 — clean, well-scoped accessibility fix and nice that you added Escape-handler tests. The implementation logic is sound (the panel is conditionally mounted, so the document listener isn't live when no trace is open). A few things to address before merge — I've left inline notes:

**Please fix (blocking):**
- The `// #1179` code comment references an issue number. Our convention is to keep issue/PR numbers out of code comments (and commit messages) — they're opaque to anyone reading the source later and rot over time. Describe the rationale directly instead. Issue links belong in the PR description (`closes #1179` there is great for auto-close).
- Same applies to the commit message: the `(#1179)` suffix on the commit title should be dropped. Mind amending it to just `fix(ui): keyboard accessibility — Escape closes trace panel, autoFocus on dialogs`?

**Worth addressing:**
- The two `autoFocus` additions have no test coverage — our CI runs a diff-coverage gate (≥80% on changed lines), and these land at 0%. A small jsdom test per dialog asserting the name input is `document.activeElement` would cover them and lock in the a11y behavior. Worth running `pnpm test:coverage` + `diff-cover` locally to confirm the gate passes before pushing.

**Non-blocking, but please sanity-check manually:**
- Possible Escape double-close with Radix overlays — see the inline note on the guard.

No need to rename the branch — renaming it on GitHub would close this PR. Just something to keep in mind for next time (branch titles should describe the change rather than embed the issue number).

ka1kqi:
Also, please include a screenshot or video in future prs

anishkumar0504:
Thanks for the detailed review! Addressed everything:

* Removed issue number from the code comment, rewrote as rationale
* Amended commit message to drop (#1179)
* Added jsdom tests for both autoFocus additions (CreateProjectDialog, CreateWorkspaceDialog) — both passing locally
* Added role="menu" to the Escape guard selector for the Radix double-close issue
Re: the manual Radix double-close check -> I wasn't able to spin up the full dev stack locally (Docker/ClickHouse/Postgres setup on Windows), so I traced through the logic instead: since the guard now checks `[role="menu"]` in addition to `[role="dialog"]` and `[role="listbox"]`, and Radix's own Escape handling stops propagation on its internal content layer before our document listener's check runs on the new focused element (the trigger), the close should be scoped correctly. Happy to verify live if you can point me to a quick way to run just the frontend without the full stack.
Ready for re-review on the rest.

frontend/ui/src/features/traces/components/TraceViewerPanel.tsx
```diff
@@ -160,7 +160,25 @@ export function TraceViewerPanel({
...
+      const activeEl = document.activeElement;
+      if (
+        activeEl instanceof HTMLElement &&
+        activeEl.closest(
+          '[role="dialog"], [role="listbox"], [role="menu"], [data-radix-popper-content-wrapper]',
+        )
+      ) {
+        return;
+      }
+      onClose();
```
ka1kqi: This `activeElement` check is the source of the double-close described above — focus has already returned to the (out-of-dialog) trigger by the time this bubble-phase listener runs. Keying off `defaultPrevented` is timing- and focus-independent:

```suggestion
  // Close the panel on Escape. A nested Radix overlay (dialog/select/popover/menu) that
  // consumes the Escape calls preventDefault() in the capture phase, before this
  // bubble-phase listener runs — so defaultPrevented means "already handled, leave the panel open".
  useEffect(() => {
    const handleKeyDown = (e: KeyboardEvent) => {
      if (e.key !== "Escape") return;
      if (e.defaultPrevented) return;
      onClose();
```

ka1kqi (commented):
Thanks for the quick turnaround @anishkumar0504 — the comment is reworded, the `role="menu"` selector is added, and the two dialog focus tests are great. I ran them locally and all pass.

Two things before merge: a real keyboard bug in the Escape guard, and a missing case (the session detail panel).

## 1. Escape double-close bug (blocking)

Reproduce manually: open a trace's detail panel, then open the **Create Project** (or Create Workspace) dialog on top of it, and press Escape once — **both** the dialog and the trace panel close, when only the dialog should.

**Why:** the guard decides based on where focus is (`document.activeElement`) at keydown time. But Radix listens for Escape on `document` in the **capture phase** and, when its top overlay consumes the key, it closes that overlay and `preventDefault()`s — all before this bubble-phase listener runs. Closing the dialog returns focus to its trigger button, which lives *outside* the dialog, so by the time this listener reads `document.activeElement` the guard no longer matches and the panel closes too. jsdom defers the unmount/focus-return until after the event dispatch, so focus still looks like it's inside the dialog there — which is why the unit tests stay green while the real browser breaks.

**Fix:** key off `e.defaultPrevented` instead of focus (Radix sets it in the capture phase whenever a nested overlay handles the Escape). Robust regardless of where focus lands, and it lets us drop the role/popper selector list entirely. Inline suggestion below.

## 2. Apply the same handler to the session detail panel

`SessionDetailPanel` (`frontend/ui/src/features/traces/components/SessionDetailPanel.tsx`) is the sibling slide-in on the sessions page — same shape as the trace panel (conditionally mounted, takes `onClose`, is/will be expandable) — but it has no Escape handler at all. Please add the same effect there so Escape-to-close is consistent across both detail panels, using the `defaultPrevented` form. Something like (drop it in alongside the existing effects in the component body):

```ts
// Close the panel on Escape, unless a nested Radix overlay already consumed the key.
useEffect(() => {
  const handleKeyDown = (e: KeyboardEvent) => {
    if (e.key !== "Escape") return;
    if (e.defaultPrevented) return;
    onClose();
  };
  document.addEventListener("keydown", handleKeyDown);
  return () => document.removeEventListener("keydown", handleKeyDown);
}, [onClose]);
```

## Verification

After the changes, manually verify on **both** panels (trace detail and session detail, expanded and inline): (a) Escape with an overlay open closes only the overlay, and (b) Escape with no overlay open still closes the panel. A small jsdom regression test asserting the bubble-phase Escape carries `defaultPrevented === true` while a Radix dialog is open would lock in the intent. If it's easier to factor the handler into a tiny shared `useEscapeToClose(onClose)` hook so both panels share one tested implementation, that's welcome but not required.

ka1kqi:
@anishkumar0504 please make fixes to have your pr merged

anishkumar0504:
Thanks for the thorough catch on the double-close bug - makes total sense given Radix's capture-phase timing. Updated:
- Switched the Escape guard to check e.defaultPrevented instead of document.activeElement, dropped the role/popper selector list entirely

- Added the same Escape-to-close handler to SessionDetailPanel for consistency

- Updated/added jsdom tests on both panels — the regression test dispatches a KeyboardEvent with preventDefault() already called to assert onClose is not triggered when a nested overlay has consumed the Escape
Ready for re-review.

XinweiHe:
@anishkumar0504 If any previous comments are resolved. Feel free to click "resolve" s.t. future code review becomes easier.

anishkumar0504:
Got it, thanks for the tip! Going through and resolving the threads that were addressed in the latest commit now. Let me know if anything else needs another look.

anishkumar0504:
Resolved all previously addressed threads. PR should be good to go  let me know if anything else needs a look before merge!

XinweiHe (approved):

PR #1016, opened 2026-06-02
fix: store filter state in sessionStorage

Adarsh-Raj-Jaiswal:
Fixes #951 

## Summary
Storing the URL with filter state in sessionStorage and setting it again on navigation.
## Type of Change

- [ ] feat - A new feature
- [X] fix - A bug fix
- [ ] docs - Documentation only changes
- [ ] style - Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc)
- [ ] refactor - A code change that neither fixes a bug nor adds a feature
- [ ] perf - A code change that improves performance
- [ ] test - Adding missing tests or correcting existing tests
- [ ] build - Changes that affect the build system or external dependencies (example scopes: gulp, broccoli, npm)
- [ ] ci - Changes to our Cl configuration files and scripts (example scopes: Travis, Circle, BrowserStack, SauceLabs)
- [ ] chore - Other changes that don't modify sc or test files
- [ ] revert - Reverts a previous commit
- [ ] security - A security fix or improvement
- [ ] github - Changes to our GitHub configuration files and scripts
- [ ] other (please describe):

## Screenshots / Recordings (if applicable)

https://github.com/user-attachments/assets/661fc85d-4758-4c31-bbd8-59d810a97767

## Checklist

- [ ] I have read the [CONTRIBUTING.md](../CONTRIBUTING.md)
- [ ] I have added/updated tests where applicable
- [X] I have added screenshots or recordings for UI changes (if applicable)
- [ ] I have updated documentation where necessary

---
## Summary by cubic
Persist filter and search state on the Detectors list by saving the current query string in session storage and restoring it when returning from a Detector detail. Users land back on the list with the same filters applied.

- **Bug Fixes**
  - Store the current list URL (with query params) in `sessionStorage` as `detectorsReturnUrl` on row click, then navigate to the detail page.
  - On the detail page, the “Detectors” button restores and clears `detectorsReturnUrl`; falls back to `/projects/:projectId/detectors` if missing.
  - Replace direct `router.push` calls with `handleRowClick` and `handleDetectorsClick` to manage this flow.

<sup>Written for commit 420963f3ad3900e533896127a83ddf00852fda19. Summary will update on new commits.</sup>

[...]

PR #1184, opened 2026-06-13
detectors: persist the selected time range across navigation 

### LINKED ISSUE #951: Detector page time picker does not persist selection
### Problem

The time range picker on the Detectors page does not cache/persist the user's selection. When navigating away and returning to the page, the time range resets to its default value.

### Proposed Solution

The Detectors page time picker should behave consistently — persisting the user's selected time range across navigation.

### Additional Context

https://github.com/user-attachments/assets/2f5ef5c3-aef0-4633-a147-02b7ed1ef1e3

- [ ] I would like to work/help with this

lqiu03:
Supersedes #1016. Fixes #951.

**Summary**
The Detectors list time-range picker did not persist: leaving the page and returning reset it to the default. This now remembers the selected range so it survives every way of returning to the list.

## Type of Change

- [x] fix - A bug fix

**Context**:
#1016 stored the list URL in sessionStorage and restored it from a single in-page "Detectors" button. That left the two most common ways back uncovered — the left-sidebar "Detectors" link and the browser back button never ran that code, so the range was still forgotten — and the shared storage slot could route across projects. #951 asks for persistence across all navigation, which a single-button or URL-only mechanism cannot provide, since the sidebar link carries no query params.

**What changed**
New date-filter-persistence.ts adds a small localStorage helper that remembers a list's selected range, reusing the shared readStored / writeStored guards. It rejects unknown or removed preset ids so a stale entry falls back to the page default instead of a silently wrong window.

The helper writes the selection on change and restores it once per key on mount, only when the URL has no explicit date_filter (so shared and bookmarked links still win). It restores into state rather than the URL (no URL pinning), re-runs when the key changes so switching projects does not keep the previous project's range, and resets pagination so a narrower window does not strand an out-of-range page. useListPageState now threads the key through to useUrlDateFilter.

The Detectors list opts in with a per-project key, so returning via the sidebar link, browser back, after creating or viewing a detector, or after a full reload all restore the last selected range.

End-to-end tests in local docker build:
https://github.com/user-attachments/assets/0cee581a-dbf9-4ccb-9bf4-85c056ca5e87
Demo 1
https://github.com/user-attachments/assets/60936d31-9da7-423b-9a1c-d04f59549601
Demo 2

**Scope & safety**
This is opt-in: Traces, Sessions, Users, and the detector detail page pass no persistKey and are unchanged. The per-project key ensures one project's range never leaks into another. An explicit ?date_filter= link still wins over the stored value.

**Test plan**
Unit tests in date-filter-persistence.test.ts (10 cases) cover preset and custom-range round-trips, per-key isolation, and rejection of malformed JSON, missing ids, unknown ids, and incomplete custom ranges.

Manual: on a project's Detectors list, set "Last 1 hour", then confirm it persists via (a) the sidebar "Detectors" link, (b) opening a detector and returning, (c) creating a detector and returning, and (d) a full reload. Then switch to a different project's Detectors and confirm it shows that project's last range or the 14‑day default, not the first project's.

**Notes**
The detector detail page keeps its own 24‑hour default; #951 concerns returning to the list. This picks up the original branch, which had not seen further updates.

## Checklist

---
## Summary by cubic
Keep the selected time range consistent between the Detectors list and detail by passing it in the URL. Arriving from the sidebar resets to the default, matching Traces/Users/Sessions behavior.

- **Bug Fixes**
  - Replaced localStorage persistence with URL propagation using `buildUrlWithFilters` in `DetectorsPage` and `DetectorDetailPage` so row clicks and the "Detectors" link include `date_filter` (and custom ranges).
  - Reverted prior `useUrlDateFilter`/`useListPageState` persistence changes; added page tests to assert the propagated URLs in both directions.

- **Refactors**
  - Centralized the detectors default time range as `DETECTORS_DEFAULT_DATE_FILTER_ID` (14d) in `date-filter.ts` and used it in both pages; added `date-filter.test.ts` to pin the constant to a valid option.

<sup>Written for commit f5a133774f6a3ca7e43cb32fb29afc78a28e6402. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1184?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lqiu03:
@cubic-dev-ai review again

lqiu03:
Use tracing-style navigation: drop the #1184 localStorage and pivoting to keeping detector time range only via URL propagation between list and detail. Keeping "Tracing" and "Detector" consistent across the board. Commits and tests incoming soon

lqiu03:
**E2e test via docker build:** 
https://github.com/user-attachments/assets/c88aacfe-562c-4026-8ed1-aed2605548b9
1. Project → Detectors → set range to Last 7 days.
2. Click a detector row → detail should open on 7 days (URL…/detectors/<id>?date_filter=7d).
3. Change it on the detail to 1 hour → click Detectors back →list should return on 1 hour (?date_filter=1h).
4. Use the sidebar "Detectors" (or leave the section and come back via sidebar) → reset to 14‑day default

Overall, it seems complicated but in short, this PR aligns with "Tracing" page.

XinweiHe:
claude code review

```
  "14d" as a bare string duplicated across two files is fragile: change the
  detector default later and you must edit both, and they will drift (they
  already did — that's how detail ended up at 1d while list was at 14d). Compare
  to how the codebase already centralizes this: DEFAULT_DATE_FILTER lives once
  in lib/date-filter.ts and Traces/Users/Sessions all inherit it by calling
  useListPageState() with no arg.

  My recommendation — add one constant next to the existing default and import
  it in both pages:

  // lib/date-filter.ts
  export const DETECTORS_DEFAULT_DATE_FILTER_ID = "14d";

  // both detector pages
  useListPageState({ defaultDateFilterId: DETECTORS_DEFAULT_DATE_FILTER_ID });

  That makes the dependency you're asking for explicit (single source of truth),
  prevents the exact drift that already happened, and matches the existing
  DEFAULT_DATE_FILTER pattern. "Making them dependent" via shared selection at
  runtime isn't needed — when you navigate list→detail the range already rides
  through the URL (?date_filter=), so the default is only the fallback for
  direct/bookmarked/refreshed detail loads. They just need the same fallback
  value, which one constant gives you.
```

after addressing this, I feel like this is mostly ready to merge.

XinweiHe (approved):

PR #1187, opened 2026-06-14
fix(worker): recognize Vercel AI SDK token attributes so cache breakdown persists

### LINKED ISSUE #1181: Ingest transform doesn't recognize Vercel AI SDK token attributes — cache breakdown lost
## Summary

Spans emitted by the Vercel AI SDK (`ai/6.x` — `ai.generateText`, `ai.generateText.doGenerate`) consistently land in ClickHouse with `usage_details` containing **zeros** for `cache_read_tokens` and `cache_write_tokens`, even when the span's metadata clearly shows cache being used heavily. As a result, the trace viewer's "Usage breakdown" panel shows the entire input as `uncached`, hiding real cache hits and misrepresenting cost behavior.

## Symptom

A span with metadata like:
```
ai.usage.cachedInputTokens: 22041
ai.usage.inputTokenDetails.cacheReadTokens: 22041
ai.usage.inputTokenDetails.cacheWriteTokens: 6422
ai.usage.inputTokenDetails.noCacheTokens: 3
```

writes to ClickHouse as:
```
usage_details = {'cache_read_tokens': 0, 'cache_write_tokens': 0, 'reasoning_tokens': 0}
```

Trace viewer then shows:
```
Input usage:  N
  uncached:   N      ← wrong; ~78% of this was cache_read
  cache read: 0
  cache write: 0
```

## Root cause

`backend/worker/otel_transform.py` lines ~440–498. The attribute-key lookup lists for cache tokens cover:
- **OpenInference**: `llm.token_count.prompt_details.cache_read` / `cache_write`
- **OTel GenAI semconv**: `gen_ai.usage.cache_read.input_tokens`, `gen_ai.usage.cache_creation.input_tokens`, several `gen_ai.usage.details.*` variants used by pydantic-ai

They do **not** cover the **Vercel AI SDK** namespace, which uses its own keys: `ai.usage.cachedInputTokens`, `ai.usage.inputTokenDetails.cacheReadTokens`, `ai.usage.inputTokenDetails.cacheWriteTokens`, `ai.usage.inputTokens`, `ai.usage.outputTokens`.

When none of the lookup keys match, `api_cache_read_tokens` resolves to `None`, gets coerced to `0`, and zero is persisted to the `usage_details` Map alongside (otherwise correct) other token fields.

## Scope

Roughly half of recent model spans in production have empty/zero `usage_details` — a meaningful chunk of that population is Vercel-AI-SDK traffic. Anything from the Vercel AI SDK is affected regardless of the underlying LLM provider (Anthropic, OpenAI, etc.) — the SDK abstracts those behind its own attribute namespace.

## Fix

Add the Vercel AI SDK attribute names to each lookup list in `otel_transform.py`:

```python
# api_input_tokens — add:
\"ai.usage.inputTokens\",

# api_output_tokens — add:
\"ai.usage.outputTokens\",

# api_cache_read_tokens — add (after the OpenInference / GenAI semconv keys):
\"ai.usage.cachedInputTokens\",
\"ai.usage.inputTokenDetails.cacheReadTokens\",

# api_cache_write_tokens — add:
\"ai.usage.inputTokenDetails.cacheWriteTokens\",
```

Add a corresponding test fixture in `tests/worker/test_otel_transform.py` that exercises the Vercel AI SDK attribute shape.

## Caveats

- Backfill is not possible — existing rows already have zeros written. Only newly-ingested spans will be correct after the fix. Acceptable, since cost columns aren't affected (the cost path uses the same buckets but cost was being computed correctly via text-estimation fallback for these spans).
- We should probably add a similar check for *other* common SDKs that use their own attribute namespaces (langchain, llamaindex, etc.) — track separately if not already covered.

ka1kqi:
## Summary

Vercel AI SDK spans (`ai.generateText`, `ai.generateText.doGenerate`, `ai.generateObject`) were landing in ClickHouse with `usage_details` cache/reasoning tokens zeroed out, so the trace viewer's usage breakdown showed the entire input as uncached and cost was overcharged. Reported against Vercel AI SDK **v6.x** (`ai/6.x`).

**Root cause:** the SDK exposes cache and reasoning token detail *only* under its raw `ai.usage.*` attribute namespace. The OpenInference vercel span processor normalizes input/output **totals** to `llm.*`/`gen_ai.*` but never the cache or reasoning split (verified through the latest upstream release), and the ingest transform's lookup lists didn't cover the raw keys — so they resolved to `None` → `0`.

## Changes

- **`backend/worker/otel_transform.py`** — append `ai.usage.*` keys as lowest-priority fallbacks to all five token lookup lists (input, output, cache-read, cache-write, reasoning). Normalized `llm.*`/`gen_ai.*` keys keep priority, so existing SDK paths are unaffected. Includes the legacy `ai.usage.promptTokens`/`completionTokens` spellings that current `ai.generateObject` spans still emit.
- **`backend/worker/tokens/buckets.py`** — register the SDK's exact `"ai"` tracer scope as a known gross (cache-inclusive) emitter via a new exact-match set. A bare `"ai"` prefix was avoided deliberately — it would swallow unrelated `ai*`-named scopes (e.g. `aiohttp.client`).
- Tests in `tests/worker/test_otel_transform.py` and `tests/worker/tokens/test_buckets.py`.

All attribute shapes were verified against the published `ai@6.0.204` source: tracer scope is exactly `"ai"`, `inputTokens` totals are gross (noCache + cacheRead + cacheWrite), and `generateObject` emits only the legacy total spellings.

## Caveat — Vercel AI SDK version coverage

The cache-token telemetry attributes are **version-dependent**, which is why the issue specifies v6.x:

| SDK major | Emits | Cache breakdown captured by this fix |
|-----------|-------|--------------------------------------|
| **v4** | `ai.usage.promptTokens` / `completionTokens` only | Totals only — the SDK emits **no** cache attributes, so there is no cache data to capture |
| **v5** | adds `ai.usage.cachedInputTokens` (cache **read**) | Cache **read** + totals; v5 emits no cache-**write** attribute |
| **v6+** | adds `ai.usage.inputTokenDetails.{cacheReadTokens,cacheWriteTokens,noCacheTokens}` | Full breakdown — read **and** write; this is the version the issue reports |

The fix carries keys for all of these spellings, so it degrades gracefully across versions: v6+ gets the complete read/write split (the reported case), v5 gets cache-read, and v4 has no cache data to lose. Verified end-to-end with a live v6 + Anthropic prompt-caching run against the local stack: a real `ai.generateText.doGenerate` span landed in ClickHouse with non-zero `cache_read_tokens`/`cache_write_tokens` in `usage_details`.

## Cost impact

For `doGenerate` spans the prior behavior was not cost-neutral: they had API totals, so cost *was* computed but priced the entire input at the uncached rate — roughly a 2.7× overcharge on a cache-heavy span. Newly-ingested spans are now priced correctly. No backfill: existing rows already persisted zeros.

## Testing

- Full backend suite: 355 passed; worker suite 236 passed.
- Diff-coverage vs `main`: 100% on changed lines.
- New tests reproduce the exact production span shape (3 uncached + 22041 cache-read + 6422 cache-write = 28466 gross) and assert the cache-discounted cost by hand; red phase confirmed by reverting only the source change.

## Screenshots
Screenshot 2026-06-15 at 9 30 08 AM
Screenshot 2026-06-15 at 9 30 17 AM

closes #1181

---
## Summary by cubic
Fixes incorrect pricing for Vercel AI SDK spans: we now read `ai.usage.*` token attributes so cache read/write and reasoning tokens persist to `usage_details`, and we only use raw `ai.usage` gross totals on LLM spans to prevent wrapper double-counting. Also registers the SDK’s `ai` scope as a known gross emitter. Closes #1181.

- **Bug Fixes**
  - Added `ai.usage.*` fallbacks for input/output totals on LLM spans only (includes legacy `promptTokens`/`completionTokens`); kept `llm.*`/`gen_ai.*` as higher priority.
  - Added `ai.usage.*` fallbacks for cache-read, cache-write, and reasoning tokens so the full breakdown persists.
  - Registered exact `ai` tracer scope as a known gross emitter; avoids prefix matching.
  - Tests cover `ai.generateText`, `ai.generateText.doGenerate`, and `ai.generateObject`, including no double-count on AGENT wrappers and legacy spellings.

<sup>Written for commit b73917c072dd740993ba27a4f4daf76bcf65ca05. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1187?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/worker/otel_transform.py
+++ b/backend/worker/otel_transform.py
@@ -448,6 +448,12 @@ def transform_otel_to_clickhouse(
                             "llm.token_count.prompt",
                             "gen_ai.usage.input_tokens",
                             "gen_ai.usage.prompt_tokens",
+                            # Vercel AI SDK raw attrs (its OpenInference span
+                            # processor normalizes totals only on LLM-kind
+                            # spans, so outer agent spans keep the raw keys):
+                            "ai.usage.inputTokens",
+                            # generateObject still emits only this spelling:
+                            "ai.usage.promptTokens",
                         ],
                     )
                     api_output_tokens = first_present_number(
@@ -456,6 +462,10 @@ def transform_otel_to_clickhouse(
                             "llm.token_count.completion",
                             "gen_ai.usage.output_tokens",
                             "gen_ai.usage.completion_tokens",
+                            # Vercel AI SDK raw attrs:
+                            "ai.usage.outputTokens",
+                            # generateObject still emits only this spelling:
+                            "ai.usage.completionTokens",
                         ],
                     )
                     # Cache buckets. The OpenInference keys (prompt_details.*) are the
@@ -472,6 +482,10 @@ def transform_otel_to_clickhouse(
                             # pydantic-ai version variants (names differ by release):
                             "gen_ai.usage.cache_read_tokens",
                             "gen_ai.usage.details.cache_read_input_tokens",
+                            # Vercel AI SDK: cache detail is NEVER normalized to
+                            # llm.*/gen_ai.* — it exists only under ai.usage.*:
+                            "ai.usage.cachedInputTokens",
+                            "ai.usage.inputTokenDetails.cacheReadTokens",
                         ],
                     )
                     api_cache_write_tokens = first_present_number(
@@ -483,6 +497,8 @@ def transform_otel_to_clickhouse(
                             "gen_ai.usage.details.cache_write_tokens",
                             # pydantic-ai version variant:
                             "gen_ai.usage.details.cache_creation_input_tokens",
+                            # Vercel AI SDK raw attr (never normalized upstream):
+                            "ai.usage.inputTokenDetails.cacheWriteTokens",
                         ],
                     )
                     # Reasoning tokens (o-series / GPT-5): a SUBSET of output
@@ -495,6 +511,9 @@ def transform_otel_to_clickhouse(
                             "gen_ai.usage.reasoning_tokens",
                             "gen_ai.usage.output_details.reasoning_tokens",
                             "gen_ai.usage.details.reasoning_tokens",
+                            # Vercel AI SDK raw attrs (aliases of the same value):
+                            "ai.usage.outputTokenDetails.reasoningTokens",
+                            "ai.usage.reasoningTokens",
                         ],
                     )
 
--- a/backend/worker/otel_transform.py
+++ b/backend/worker/otel_transform.py
@@ -442,32 +442,30 @@ def transform_otel_to_clickhouse(
 
                     # Try API-provided token counts first (from instrumentors).
                     # OpenInference: llm.token_count.*  ·  GenAI semconv: gen_ai.usage.*
-                    api_input_tokens = first_present_number(
-                        span_attrs,
-                        [
-                            "llm.token_count.prompt",
-                            "gen_ai.usage.input_tokens",
-                            "gen_ai.usage.prompt_tokens",
-                            # Vercel AI SDK raw attrs (its OpenInference span
-                            # processor normalizes totals only on LLM-kind
-                            # spans, so outer agent spans keep the raw keys):
-                            "ai.usage.inputTokens",
-                            # generateObject still emits only this spelling:
-                            "ai.usage.promptTokens",
-                        ],
-                    )
-                    api_output_tokens = first_present_number(
-                        span_attrs,
-                        [
-                            "llm.token_count.completion",
-                            "gen_ai.usage.output_tokens",
-                            "gen_ai.usage.completion_tokens",
-                            # Vercel AI SDK raw attrs:
-                            "ai.usage.outputTokens",
-                            # generateObject still emits only this spelling:
-                            "ai.usage.completionTokens",
-                        ],
-                    )
+                    input_token_keys = [
+                        "llm.token_count.prompt",
+                        "gen_ai.usage.input_tokens",
+                        "gen_ai.usage.prompt_tokens",
+                    ]
+                    output_token_keys = [
+                        "llm.token_count.completion",
+                        "gen_ai.usage.output_tokens",
+                        "gen_ai.usage.completion_tokens",
+                    ]
+                    if span_kind == SpanKind.LLM:
+                        # Vercel AI SDK raw GROSS totals are the only token source it
+                        # normalizes to neither llm.* nor gen_ai.*, so we read them as
+                        # a fallback. But they sit on BOTH the LLM doGenerate span AND
+                        # its AGENT/CHAIN wrapper (ai.generateText/ai.generateObject),
+                        # where they restate the SUM of the wrapper's LLM children.
+                        # Trust them only on the LLM span itself — otherwise the
+                        # wrapper is priced on top of its children (double count).
+                        # generateObject emits only the legacy *Tokens spelling; its
+                        # real usage still lands on an LLM .doGenerate child.
+                        input_token_keys += ["ai.usage.inputTokens", "ai.usage.promptTokens"]
+                        output_token_keys += ["ai.usage.outputTokens", "ai.usage.completionTokens"]
+                    api_input_tokens = first_present_number(span_attrs, input_token_keys)
+                    api_output_tokens = first_present_number(span_attrs, output_token_keys)
                     # Cache buckets. The OpenInference keys (prompt_details.*) are the
                     # verified path for Anthropic/OpenAI and MUST be listed first —
                     # they are the same family as llm.token_count.prompt (read above).
--- a/backend/worker/tokens/buckets.py
+++ b/backend/worker/tokens/buckets.py
@@ -49,6 +49,14 @@ _KNOWN_SCOPE_PREFIXES: tuple[str, ...] = (
     "traceroot",
 )
 
+# Scopes recognized by EXACT name — too short to prefix-match safely (a bare
+# "ai" prefix would also swallow unrelated ai*-named scopes).
+_KNOWN_SCOPE_EXACT: frozenset[str] = frozenset(
+    {
+        "ai",  # Vercel AI SDK tracer; GROSS input with cache detail under ai.usage.*
+    }
+)
+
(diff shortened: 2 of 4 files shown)


XinweiHe:
Reviews from Claude. Maybe we could adjust this PR a bit after https://github.com/traceroot-ai/traceroot/pull/1174 lands and do e2e test one more time to verify.

```
  The catch: #1174's gate only guards the estimation (elif) branch. #1187's ai.usage.* keys feed the API-count (if api_input_tokens is not None)
  branch — which is not span-kind-gated (#1174 deliberately trusts API counts on any kind). So the production ai.generateText AGENT wrapper, which
  carries ai.usage.inputTokens totals, gets priced again alongside its doGenerate children.

  I trial-merged both onto main (clean — no git conflict, all 119 merged tests pass) and ran a production-shaped trace (AGENT wrapper with ai.usage.*
  + LLM doGenerate child):

                         both #1174+#1187          #1174 alone
  ai.generateText  (AGENT)  in=28466 out=120   →   in=None  out=None   ← correctly suppressed
  ai.generateText.doGenerate (LLM) in=28466    →   in=28466 out=120
  TRACE total               57172  (2×)        →   28586  (correct)
  
    Before landing both worker PRs, reconcile them — they shouldn't merge as-is regardless of order:
  - Restrict the gross-total ai.usage.* keys (inputTokens/outputTokens/promptTokens/completionTokens) to LLM-kind spans, mirroring #1174's gate, so
  the AGENT wrapper isn't re-priced. The cache/reasoning detail keys (cacheReadTokens, etc.) are safe to leave ungated — they only refine a total
  that already exists on the child.
  - Resolve the contradiction: #1187's test_vercel_outer_span_raw_totals_used_as_api_counts and #1174's "wrapper contributes nothing" cannot both be
  correct. One must change. (Watch generateObject, which #1187 says emits only the legacy spellings — confirm its real usage lands on an LLM child
  before gating, or that case loses its counts.)
  - Add the missing regression fixture: AGENT wrapper with ai.usage.* totals + an LLM doGenerate child, asserting the trace total equals the child
  only. That's the gap that hid this.

  Simplest path: merge #1174 first, then rebase #1187 on top and adapt its outer-span handling/test against the live gate.
```

ka1kqi:
@cubic-dev-ai please re-review — pushed a follow-up that addresses the double-count finding.

**What changed since the last review** (branch rebased onto `main`, which now includes the LLM-only text-estimation gate):

- New commit gates the Vercel raw **gross-total** keys (`ai.usage.inputTokens` / `promptTokens` / `outputTokens` / `completionTokens`) to **LLM-kind spans only**. The SDK stamps these aggregates on both the `ai.generateText.doGenerate` LLM span *and* its AGENT wrapper (`ai.generateText` / `ai.generateObject`), where the wrapper restates the sum of its children — so reading them on the wrapper double-counted the trace.
- The wrapper carries `openinference.span.kind=AGENT` (set by the OpenInference Vercel processor) and no normalized `llm.*`/`gen_ai.*` keys, so gating drops it entirely while every LLM child keeps its counts.
- The **cache/reasoning detail keys are unchanged and still ungated** — they're only consumed once an input/output total is present (which the wrapper no longer has), so the cache-token fix is fully preserved.

**Verification** (real Vercel AI SDK v6 capture, `ai@6.0.205` + `@arizeai/openinference-vercel@2.7.8`):
- Multi-step `generateText`: wrapper `ai.usage.inputTokens` = 203 = 67 + 136 (sum of its two LLM children). Before the gate the trace summed to 406; after, it's 203 (children only) with the wrapper contributing 0.
- Regression tests added: AGENT wrapper + LLM children (asserts trace total = children, not 2×), `generateObject` (AGENT suppressed, LLM child priced via legacy spellings), an LLM-only span where the raw totals are the sole source (still adopted), and a non-LLM (AGENT/CHAIN) drop test. Full worker suite green (262 passed).

ka1kqi:
@cubic-dev-ai re-review please — update with cross-instrumentor regression verification.

Branch is current at `b73917c0` (rebased on `main`, includes the LLM-only estimation gate + this fix). No code changes since the last review; this is added verification context.

**Cross-instrumentor regression sweep.** To make sure gating the Vercel `ai.usage.*` gross totals didn't leave (or mask) the same wrapper-double-count for other instrumentors, I ran live `gpt-4o-mini` agents across five framework/language combinations and checked whether any **non-LLM/wrapper span carries token-usage totals**:

| Instrumentor | Lang | Wrapper carries usage totals? | Wrapper has model name? | Double-counts via transform? |
|---|---|---|---|---|
| Vercel AI SDK | TS | yes (`ai.usage.*`) | yes (`llm.model_name`) | was yes → **fixed here** |
| OpenAI Agents SDK | Py | no | — | no |
| OpenAI Agents SDK | TS | no | — | no |
| CrewAI | Py | no | — | no |
| LangGraph / LangChain | Py | no | — | no |
| pydantic-ai | Py | yes (`gen_ai.usage.*`) | no | no (model-name guard skips it) |

**Result: no double-count regression in any tested instrumentor.** Vercel is the unique case — its OpenInference processor stamps *both* a model name and aggregate `ai.usage.*` on the AGENT wrapper — which is exactly what this PR gates. pydantic-ai rolls usage onto its `invoke_agent` wrapper too, but is safe because that span carries no model name, so the transform's `if model_name:` guard never prices it (confirmed by replaying the real captured spans through the transform).

**Follow-up filed:** #1206 — generalize the gate to *all* API token-count adoption (not just `ai.usage.*`) so the wrapper-double-count class is closed permanently. That's intentionally **out of scope here** (it touches a deliberately-ungated path and needs its own cross-instrumentor verification before implementation); this PR stays the focused Vercel fix.

XinweiHe (approved):
LGTM

Issue #1070, opened 2026-06-04
Distinguish Anthropic 5m vs 1h prompt-cache tokens for accurate pricing


ka1kqi:
## Problem

We currently collapse all Anthropic prompt-cache **write** tokens into a single `cache_write` / `cacheWrite` bucket priced at one rate. Anthropic charges different rates depending on the cache TTL:

| Token type | Multiplier vs base input |
|---|---|
| 5-minute cache write | **1.25×** |
| 1-hour cache write | **2.0×** |
| cache read (hit, either TTL) | 0.1× |

**Confirmed: our single rate is the 5-minute rate.** Every Anthropic entry in `frontend/packages/core/src/standard-model-prices.json` carries `cacheWrite = 1.25× input` (e.g. opus 4.x: $6.25/MTok vs $5 input; sonnet 4.x: $3.75 vs $3; haiku 4.5: $1.25 vs $1), and that JSON is synced verbatim into the `standard_model_prices` table that both the Python worker and the TS services price from. So cost is systematically **under-counted whenever 1h caching is used**: a 1h write is billed by Anthropic at 2.0× but recorded by us at 1.25× — a 37.5% undercount on that line item. Concrete example: a session writing 200k tokens to a 1h cache on an opus 4.x model costs $2.00 but we record $1.25. Reads are unaffected (same 0.1× regardless of TTL).

## The data exists upstream — but is dropped at the instrumentation layer

The provider `usage` payload reports the breakdown:

- `cache_creation_input_tokens` (total writes)
- `cache_creation.ephemeral_5m_input_tokens`
- `cache_creation.ephemeral_1h_input_tokens`
- `cache_read_input_tokens`

However — and this answers the open question in the original write-up — **none of the instrumentation sources we ingest from propagate the split into span attributes.** Verified against upstream source:

- **OpenInference** (Python + JS): the semantic conventions define only the collapsed `llm.token_count.prompt_details.cache_write` (`spec/semantic_conventions.md`; emitted via `openinference-instrumentation/_attributes.py` and JS `attributeHelpers.ts`). The `cache_creation.ephemeral_*` fields appear in their test cassettes (raw recorded API responses) but are never mapped to an attribute.
[...]

Issue #1069, opened 2026-06-04
Surface per-category cost breakdown in the UI (cost popup)


ka1kqi:
### Why

Now that cached tokens are tracked as first-class buckets (#956, #958) — uncached input, cache read, cache write, reasoning — we can show not just *how many* tokens fell in each category but *how much each cost*. Cache read is ~0.1× input and cache write ~1.25×, so the cost composition is non-obvious and worth surfacing.

### What

A **cost breakdown hover popup**, same style/placement as the token-count "Usage breakdown" popup from #958 but for **cost instead of tokens**. Hovering the cost chip (the `$…` on a span and the trace-level rollup) shows the per-category dollar breakdown:

- **Input cost** — uncached input · cache read · cache write (each = category tokens × that category's rate)
- **Output cost** — output (reasoning is part of output, already at the output rate)
- **Total** — reconciles to the span/trace cost

### Notes

- Reuses the stored `cache_read_tokens` / `cache_write_tokens` columns and the per-category rates already in the pricing table (`cost_from_buckets`).
- Display-only — no change to how cost is computed or stored.

Part of #959.

Issue #963, opened 2026-05-28
"Error loading findings" on detector findings page


shirleyai-co:
### Environment

staging.traceroot.ai

### Steps to Reproduce

1. Go to a project's Detectors page
2. Click on a detector (e.g. "Task Completion Detector")
3. The Findings tab loads with "Error loading findings" error

### Expected Behavior

The findings tab should either:
  - Display the list of findings produced by the detector
  - Show an empty state if no findings exist yet

### Actual Behavior

<img width="3112" height="1866" alt="Image" src="https://github.com/user-attachments/assets/67c038ee-174e-4813-b796-69ac518a34b4" />

### Relevant Logs/Console Output

```shell

```

nehaprasad-dev:
Hello @XinweiHe please check out the PR!

Issue #958, opened 2026-05-27
Surface cache & reasoning token breakdown in the UI


XinweiHe:
### Context

Customer traces already carry richer token telemetry than we surface:

- **Anthropic prompt caching** produces `cache_read_input_tokens` and `cache_creation_input_tokens` per span. Cache reads can be the majority of prompt tokens for agent workloads.
- **OpenAI prompt caching** produces `input_cached_tokens` per span.
- **OpenAI o-series / GPT-5 reasoning** produces `reasoning_tokens` as a subset of completion tokens.

Today none of these are visible in our UI. Users can't tell whether a high-cost span was driven by uncached input, cache writes, or reasoning. This makes our cost numbers feel opaque.

### Goal

Make the token breakdown visible somewhere on the span / trace detail view. Exact UX is open — this issue is the design + first implementation pass.

### Scoping questions to answer in design

1. **Display surface:** inline chip next to the `x → y (z)` total? Hover tooltip on the cost number? A dedicated section in the span detail panel? Some combination?
2. **Which fields to expose first:** cache_read alone? cache_read + cache_creation? Add reasoning_tokens too?
3. **Trace-level rollup:** do trace-list/trace-header rollups need to surface cached token sums, or only per-span?

### Likely prerequisites

- ClickHouse `spans` schema does not have first-class columns for `cache_read_tokens`, `cache_creation_tokens`, or `reasoning_tokens` today (the values exist only as JSON metadata strings in the `metadata` column — `backend/worker/otel_transform.py` lines 478-485). Adding columns + populating them at ingest is likely a prerequisite.
- API response types in `frontend/ui/src/types/api.ts` and the corresponding FastAPI response models will need new optional fields.

### Acceptance

- A design proposal (comment or short doc) for the display surface, approved before implementation.
- Implementation makes at least cache_read visible on a span detail view for both Anthropic and OpenAI traces.

rakshityadav1868:
## Summary
Expose cache and reasoning token information in the span detail view through a dedicated Token Breakdown section rather than adding additional chips to the existing header.

The current header already contains span kind, latency, status, model, token count, and cost. Adding multiple token categories inline would quickly become cluttered and difficult to scale as more providers expose additional token types.

### Display Surface

Proposed:
Keep the existing token summary unchanged, Add a new Token Breakdown section in the span detail panel and Only render rows when the corresponding value exists.

### Fields To Expose

[...]

PR #1189, opened 2026-06-14
feat(pricing): price Anthropic 1-hour cache-write tokens

claygeo:
## what

prices anthropic 1-hour cache writes. a 1-hour write is 2.0x base input; the default 5-minute write is 1.25x, which is already today's single `cacheWrite` rate. reads stay 0.1x and are unchanged. today every write is priced at the 5m rate, so a 1h write is under-counted by 37.5% on that line item.

part of #1070.

## why this is the downstream half

no instrumentation source we ingest from propagates the 5m/1h distinction today. openinference (py + js), pydantic-ai/logfire and openllmetry all collapse cache writes into one cache_creation total, and the agent path (pi-ai) does too. the traceroot sdks that could emit it are separate repos. so this PR builds the receiving end: purely additive, byte-identical to today's behavior until an emitter reports a 1-hour portion, then prices it correctly.

an earlier revision also tracked a separate 5m bucket. per review it was dropped, since the 5m rate is identical to the base `cacheWrite` rate so a separate bucket never changed any cost. 5m and untagged tokens fold into the remainder, priced at `cacheWrite`.

## how

the 1-hour portion is modeled as a sub-partition of the existing cache_write total, not a new disjoint bucket:

```
cache_write = cache_write_1h + remainder
```

keeping cache_write as the single total means the gross-input reconstruction (uncached + read + write) is untouched, and the cost formula stays one canonical cache_write_cost term so the total can't double-count:

```
cache_write_cost = remainder*cacheWrite + cache_write_1h*(cacheWrite1h || cacheWrite)
```

- `backend/worker/otel_transform.py` extracts cache_write_1h from new lowest-priority attribute fallbacks (e.g. `gen_ai.usage.cache_creation.ephemeral_1h_input_tokens`); stores it in usage_details only when present, so spans with no 1h portion keep an identical map
- `backend/worker/tokens/buckets.py`: TokenBuckets gains cache_write_1h (default 0); `reconcile_cache_write_1h` clamps non-negative and caps so 1h <= cache_write (shared by the ingest and read paths)
- `backend/worker/tokens/pricing.py`: the single cost formula prices the 1h portion, reconciling internally so it's correct for any TokenBuckets; cacheWrite1h falls back to cacheWrite
- `backend/rest/services/trace_reader.py`: read-path rebuild of the 1h portion from usage_details
- `frontend/packages/core/src/model-pricing/lookup.ts` mirrors the worker: pure `calculateCostFromPricing` helper (extracted so it's unit-testable without prisma) plus the 1h portion
- `standard-model-prices.json` adds cacheWrite1h = 2x input to the anthropic entries. cacheWrite already is the 5m rate so it's the only new key. usage_type is generic, no schema/migration change

## cost impact

zero change for existing data and current traffic. with no 1-hour portion present (all traffic today) every cost is identical to before, proven by regression tests on the ingest and read paths. when an emitter does report a 1-hour portion, that portion is priced at 2x input instead of 1.25x.

## testing

- backend: full suite passes. new tests for the reconcile/cap, 1h pricing (remainder + rate fallback + a defensive cap on unreconciled buckets), the byte-identical-when-absent invariant, ingest extraction + conditional storage, the read-path rebuild, and a real-json check that every anthropic entry's cacheWrite1h == 2x input
- frontend: core vitest passes; new tests for the 1h math, the cap, rate fallback and negative clamping. tsc clean
- ruff clean

## not in this PR (on purpose)

- sdk emission of the split: separate repos (python `traceroot`, npm `@traceroot-ai/traceroot`)
- the per-ttl breakout in the cost popup: the combined cache-write row keeps working unchanged; tracking with #1069

### DIFF
--- a/backend/worker/tokens/buckets.py
+++ b/backend/worker/tokens/buckets.py
@@ -13,6 +13,20 @@ the cache is a subset that subtracts cleanly; for net emitters the cache simply
 exceeds the input, so the uncached bucket floors to zero while the additive cache
 is still priced in full. Each physical token is therefore priced exactly once
 without a per-emitter convention branch.
+
+Cache WRITES can additionally carry a TTL split: Anthropic prices a 1-hour write at
+2.0x base input, versus 1.25x for the default 5-minute write (reads are 0.1x
+regardless of TTL). Since the existing ``cacheWrite`` rate already IS the 5-minute
+rate, only the 1-hour portion needs its own rate; it is modeled as a SUB-PARTITION
+of the ``cache_write`` total, never as a new disjoint bucket::
+
+    cache_write  =  cache_write_1h  +  remainder
+                    \\_ priced at _/    \\__ priced at the combined cacheWrite rate
+                       cacheWrite1h          (the 5-minute / default write rate)
+
+Keeping ``cache_write`` as the single total means the gross-input reconstruction
+(uncached + cache_read + cache_write) is unchanged, and any emitter that does not
+report the 1-hour portion (every emitter today) is priced exactly as before.
 """
 
 from __future__ import annotations
@@ -29,12 +43,18 @@ class TokenBuckets:
 
     Fields default to ``0`` so new token categories (e.g. reasoning, audio) can be
     added as purely additive changes without touching existing call sites.
+
+    ``cache_write_1h`` is NOT an additional disjoint bucket: it is a sub-partition of
+    ``cache_write`` (``cache_write_1h <= cache_write``), used only to price the 1-hour
+    write rate. It defaults to ``0``, so an emitter that does not report the 1-hour
+    portion prices identically to before.
     """
 
     input_uncached: int = 0
     output: int = 0
     cache_read: int = 0
     cache_write: int = 0
+    cache_write_1h: int = 0
 
 
 # Instrumentation scopes traceroot recognizes today. Used only to surface an
@@ -84,6 +104,18 @@ def _warn_once_if_unknown_scope(scope_name: str | None) -> None:
         )
 
 
+def reconcile_cache_write_1h(cache_write: int, cache_write_1h: int) -> int:
+    """Clamp the 1-hour cache-write portion to a valid sub-partition of the write total.
+
+    Returns ``cache_write_1h`` clamped non-negative and capped at ``cache_write`` (also
+    floored at 0), so the portion can never over-count the write total (the leftover
+    ``cache_write - cache_write_1h`` is the remainder priced at the combined cache-write
+    rate). Deterministic for any input. Shared by the ingest path (normalize_token_usage)
+    and the read path (span_cost_details).
+    """
+    return min(max(cache_write_1h, 0), max(cache_write, 0))
+
+
 def normalize_token_usage(
     scope_name: str | None,
     *,
@@ -91,6 +123,7 @@ def normalize_token_usage(
     output_tokens: int,
     cache_read_tokens: int,
     cache_write_tokens: int,
+    cache_write_1h_tokens: int = 0,
 ) -> TokenBuckets:
     """Convert an instrumentor's token counts into disjoint priced buckets.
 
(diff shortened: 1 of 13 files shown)


ka1kqi:
Thanks for this — the sub-partition modeling is careful and the test coverage is genuinely thorough.

We'd like to keep issue and PR numbers out of the source code.

Could you strip the `#1070` / `#1069` references from the code and test comments to start? The surrounding prose already carries the rationale in each spot, so they can simply be dropped — no rewording needed.

Places they appear:
- `backend/worker/otel_transform.py` — `See issue #1070.`
- `backend/worker/tokens/buckets.py` — docstring `Cache-write TTL split (#1070)`
- `tests/rest/test_trace_reader.py` — `every span before #1070`
- `tests/worker/test_otel_transform.py` — `See #1070.` / `identical to before #1070.`
- `tests/worker/tokens/test_buckets.py` — `Cache-write TTL split (#1070):`
- `tests/worker/tokens/test_pricing.py` — `Cache-write TTL split pricing (#1070).`
- `frontend/packages/core/src/__tests__/model-pricing.test.ts` — `describe("… (#1070)")`

(The `#958` reference in `otel_transform.py` is pre-existing on `main`, not introduced here — leave that one alone.)

backend/worker/tokens/buckets.py
```diff
@@ -74,13 +94,32 @@ def _warn_once_if_unknown_scope(scope_name: str | None) -> None:
...
+) -> tuple[int, int]:
+    """Clamp a cache-write TTL split to a valid sub-partition of the write total.
+
+    Returns ``(cache_write_5m, cache_write_1h)`` clamped non-negative and
+    sequentially capped so ``5m + 1h <= cache_write``. Deterministic for any input,
+    so the split can never over-count the write total (the leftover ``cache_write -
+    5m - 1h`` is the remainder priced at the combined cache-write rate). Shared by
+    the ingest path (normalize_token_usage) and the read path (span_cost_details).
+    """
+    cw = max(cache_write, 0)
```
ka1kqi: the variable names here are a little undescriptive, can you shadow the parameter names? e.g.  
```
cache_write_1h = min(max(cache_write_1h, 0), cache_write - cache_write_5m
```

ka1kqi (changes requested):

ka1kqi:
you mentioned in your comment on the issue that the rate for 5-minute cache tokens is the same as normal. maybe we should just eliminate that bucket entirely?

open for discussion about this

claygeo:
done, all of it:

- **eliminated the 5-minute bucket.** you're right that it never moved a number, since `cacheWrite` already is the 5-minute rate (1.25x input), so it was just dead weight. the model is now `cache_write = cache_write_1h + remainder`, with the remainder (untagged plus any 5m tokens) priced at `cacheWrite`. every span with no 1-hour portion (all traffic today) is byte-identical to before, both the cost and the `usage_details` map (the 1h key is omitted, not stored as 0). there's a regression test pinning that.

- **reconcile now shadows its params** like you asked. it collapsed to a single clamp: `reconcile_cache_write_1h(cache_write, cache_write_1h)` returning `min(max(cache_write_1h, 0), max(cache_write, 0))`. no more `cw/cw5/cw1`.

- **stripped the `#1070` refs** from the code and test comments. left `#958` and `#963` alone (both pre-existing on main).

- **rebased onto main** and resolved the `index.ts` conflict (kept the `.ts` import extensions from the Turbopack change, re-added the `calculateCostFromPricing` export). main also added `claude-opus-4-8` since i branched, so it gets `cacheWrite1h = 2x input` too, keeping the price-table invariant (a test asserts every anthropic entry with a cacheWrite carries the 2x 1h rate).

one extra hardening pass while i was in there: `_bucket_cost_terms` now clamps every count non-negative (not just `cache_write`), so it matches the TS `calculateCostFromPricing` mirror exactly and can never price negative on a malformed/negative span. added a parity test for the all-negative case.

backend + frontend suites green, ruff/tsc/eslint clean.

ka1kqi (approved):

XinweiHe (approved):

PR #1190, opened 2026-06-15
fix(detector): enqueue detection once per trace and evaluate after the trace settles

### LINKED ISSUE #1177: Logic Error detector returns a false-positive finding on a short/early-finishing trace (Vercel AI SDK example)
## Summary
The **Logic Error** detector produced a finding that is **not a real logic error** when run against a trace generated by the **Vercel AI SDK** example. The detector's verdict appears to be wrong (false positive).

> 📎 _screenshot: the finding + its summary_
> 📎 _screenshot: the trace it fired on_

Image

## Context: this example finishes/exports early
The Vercel AI SDK example is special: its **root span ends (and exports) early** relative to the rest of the trace's spans. That means at the moment detection runs, ClickHouse may hold only a **partial** view of the trace. A detector that judges an incomplete span set can easily reach a wrong conclusion. (This early-finish behavior is also a contributing condition in the sibling issue about the Failure detector firing multiple times.)

## How to replicate
1. Run the streaming example: `examples/typescript/vercel-ai/agent-streaming.ts`. Its trace's root span finishes/exports **early**, so detection runs against a partial trace (this is what makes the example a reliable repro).
2. In the **onboarding flow**, select the **Failure** and **Logic Error** detectors for the project.
3. Let the example trace ingest, then open the Logic Error detector and inspect the finding — it fires a logic-error finding the trace doesn't actually support (false positive).

## Observed
- Trace: `f90d84d3858197f8ef0077272cfa6e8c` (~59 spans)
- Detector: **Logic Error** (`cmqam0gi80005...`), template "Logic Error", model Default (haiku)
- Finding summary (truncated in logs): _"The agent gathers research mate..."_
- Human judgment: the trace does not actually contain a logic error → **false positive**

## Leading hypotheses (not yet confirmed)
1. **Incomplete-trace evaluation.** Because the root span finishes early, detection runs before all spans have landed; the judge sees a partial trace and infers a problem that the full trace would not support. (Most likely given the example's early-finish behavior.)
2. **Input truncation distorts the judge.** The detector judge serializes spans to JSONL and applies a naive char-budget truncation before sending to the model. On larger/unordered traces this can drop the spans that would have exonerated the agent, biasing toward false positives. See `frontend/worker/src/detection/sandbox-eval.ts` (truncation) and the spans-jsonl source `backend/rest/routers/internal.py` (`/traces/{trace_id}/spans-jsonl`).
3. **Prompt sensitivity.** The "Logic Error" template prompt may over-trigger on benign multi-step agent reasoning. Worth a prompt-quality pass with a few labeled examples.

## Investigation steps
- [ ] Capture the exact spans-jsonl payload sent to the judge for this trace and confirm whether it was complete at eval time (compare span count at eval vs final).
- [ ] Re-run the detector against the **fully settled** trace and check if the verdict flips. If it flips → hypothesis (1) confirmed, fix is the completeness/settle gate (see sibling issue).
- [ ] Check whether truncation dropped spans for this trace (log pre/post-truncation char + span counts).
- [ ] Evaluate the Logic Error template prompt against a small labeled set for over-triggering.

## Possible fixes
- Gate detection on **trace completeness** (don't evaluate until the trace has settled) — this is the shared remedy with the sibling Failure-detector issue and likely resolves (1).
- Replace head-truncation with a smarter span-selection / skeleton-first strategy so the judge never loses decisive spans.
- Tighten the Logic Error template prompt.

## Acceptance criteria
- The Vercel AI SDK example trace no longer yields a false-positive Logic Error finding.
- We have a documented root cause (completeness vs truncation vs prompt) backed by a re-run.

### LINKED ISSUE #1178: Failure detector runs 3× on a single trace: duplicate evals, duplicate billing, inflated sample rate
## Summary
A single trace causes a detector to be **evaluated multiple times** (observed: **3 runs** of the Failure detector for one trace). This is a confirmed bug with a full root-cause chain. It wastes LLM spend, writes duplicate billing rows, emits BullMQ lock-mismatch errors, and **inflates the effective sample rate** (a "25%" detector can fire on ~58–76% of traces). The UI "Runs" counter appears to "heal" back to 1 later — that is **ClickHouse hiding the evidence, not a fix**.

> 📎 _screenshot: Failure detector showing Runs = 3_
> 📎 _screenshot: same view later showing Runs = 1 (after background merge)_

Image

> **Note:** this is the **same trace** (`f90d84d3858197f8ef0077272cfa6e8c`) as #1177 — that issue is the false-positive Logic Error finding; this issue is the duplicate-run/billing problem. They share the early-finishing Vercel AI SDK example as the trigger condition.

## How to replicate
1. Run the streaming example: `examples/typescript/vercel-ai/agent-streaming.ts`. Its spans export in **multiple batches** via the OTel BatchSpanProcessor, which is what drives the per-batch re-enqueue.
2. In the **onboarding flow**, select the **Failure** and **Logic Error** detectors for the project.
3. Let the example trace ingest, then open the **Failure** detector — its "Runs" count exceeds 1 for that single trace (e.g. **3**), and findings can appear duplicated pre-merge.
4. Refresh a minute later: the "Runs" count drifts back to **1** as ClickHouse's background merge collapses the duplicate rows — confirming the runs happened but were hidden, not prevented.

## Root cause (confirmed)
A single trace is ingested across **multiple S3/Celery batches**. Detection is enqueued on **every batch**, and the enqueue path has **no deduplication**, so the same trace is evaluated once per batch.

Step by step:
1. **Per-batch enqueue.** `backend/worker/ingest_tasks.py:145-150` calls `enqueue_detector_runs` for the union of traces ∪ spans in *every* batch, so each batch of a multi-batch trace triggers detection again.
2. **Gate is presence-based, not this-batch-based.** `backend/worker/detector_tasks.py:208-210` gates on `root_span_finished`, which means "is the root span present in ClickHouse **right now**" — not "did **this batch** deliver the root." Once the root lands, *every* later batch re-passes the gate.
3. **Per-batch sampling roll.** `detector_tasks.py:216` rolls `random.random()` per batch, so a detector gets N independent sampling chances → effective rate `1-(1-p)^N` instead of `p`.
4. **Raw enqueue bypasses BullMQ dedup.** `detector_tasks.py:142-174` (`_enqueue_to_bullmq`) writes Redis directly: `hset` the job hash (overwriting) + `rpush` the **deterministic** job id `{project}--{trace}` (`detector_tasks.py:223`). A real `Queue.add({jobId})` would ignore the duplicate; the raw push does not — so the same job id is pushed to the wait list multiple times.
5. **Concurrent execution → lock mismatch → retry.** The same job id active twice produces `Lock mismatch ... Cmd retryJob from active`, and BullMQ's `attempts: 3` turns the collision into a third execution. Net: `processTrace` runs 3×.

## Evidence
- Worker log for trace `f90d84d3...`:
  ```
  [Detector] Processing trace f90d84d3... with 2 detector(s)
  [Detector] Processing trace f90d84d3... with 1 detector(s)
  Error: Lock mismatch for job 7dbc33a3...--f90d84d3...  Cmd moveToFinished from active
  Error: Lock mismatch for job 7dbc33a3...--f90d84d3...  Cmd retryJob from active
  [Detector] Processing trace f90d84d3... with 1 detector(s)
  ```
  The differing payloads ("2 detector(s)" vs "1 detector(s)") prove **separate enqueues with independent sampling rolls**, not one mishandled job.
- ClickHouse `detector_runs` for the Failure detector: 3 rows at `07:34:43`, `07:34:50`, `07:35:47`, all sharing the same deterministic `run_id` (`00686261...`).

## Why the counter "heals" to 1 (and why that's dangerous)
`detector_runs` / `detector_findings` are `ReplacingMergeTree` keyed on deterministic ids (`backend/db/clickhouse/migrations/003_create_detector_tables.sql`). A background merge later collapses the duplicate rows to 1 — so the **Runs counter drifts back to 1** even though 3 real evaluations (and 3 LLM charges) already happened. The read queries make this worse: they apply **no `FINAL` / dedup** and the findings list `INNER JOIN`s two un-deduplicated tables, so before a merge the UI shows duplicates (and can fan them out):
- `backend/rest/routers/internal.py:475-482` (findings list — no FINAL, INNER JOIN)
- `backend/rest/routers/internal.py:515-520` (trace findings — no dedup)
- `backend/rest/routers/internal.py` runs-count query (no dedup)

## Impact
- **Cost:** N× haiku detection spend per trace (permanent; merges don't refund it).
- **Billing:** N× `AIMessage` rows written (`frontend/worker/src/processors/detector-run-processor.ts`), over-counting detector usage.
- **Correctness:** sample rate silently inflated to `1-(1-p)^N`.
- **Noise:** `Lock mismatch` / retry error spam.
- **UX:** transient duplicate findings + inflated Runs count until a merge runs.

## Proposed solution
Two independent properties to enforce: (a) evaluate a trace **at most once** (dedup), and (b) evaluate it **only when complete** (completeness). The fix combines both:

**Primary — trigger once, on the root-bearing batch:**
- Change the trigger so detection enqueues only for traces whose **root span is in the current batch**, not "root present in ClickHouse." We already compute this at `ingest_tasks.py:117` (`traces_with_root`). The root span lives in exactly one batch, so this naturally fires once and — given the root span ends last — on the complete trace.

**Backstop — idempotency guard at enqueue (defense in depth):**
- Before enqueuing, acquire a Redis `SET key NX EX <ttl>` lock keyed `(project, trace)` (TTL ~1h). First batch wins; later batches no-op. Release the key if the enqueue itself fails so a dropped detection can be retried. This absorbs duplicate root delivery, concurrent batch processing, and export retries even if the primary assumption is violated.
- Place the guard **above** the sampling roll so sampling is decided exactly once per trace (fixes the rate-inflation defect for free).

**Completeness (for early-finishing/streaming traces):**
- Optionally add a short **settle delay** on the single enqueued job (e.g., 30s). Because `processTrace` re-reads spans-jsonl from ClickHouse at execution time (`detector-run-processor.ts:220`), a delayed run captures straggler spans → the one eval sees the complete trace. (Shared remedy with the Logic-Error false-positive issue.)

**Follow-up — read-time dedup (cosmetic hardening):**
- Apply `FINAL` (or `ORDER BY ... LIMIT 1 BY` / `argMax`) on the findings + runs read queries and dedup each table before the `INNER JOIN`, so the UI is correct even in the pre-merge window. Lower priority once duplicate inserts are eliminated at the source.

## Acceptance criteria
- A single multi-batch trace yields exactly **1 run per detector** and **1 finding per trace**.
- No `Lock mismatch` / spurious retry logs for detector-run jobs.
- A 25% detector fires on ~25% of traces (one sampling decision per trace).
- Exactly one `AIMessage` billing row per (detector, trace).
- (Follow-up) Findings/runs read queries return deduplicated rows even before a CH merge.

ka1kqi:
## Problem

A trace ingested across multiple S3/Celery batches re-enqueued detector evaluation on **every** batch with no deduplication, and evaluated before the trace was complete. For a single multi-batch trace this caused:

- **N evaluations** of one trace → N× LLM spend and N× billing rows (`ai_messages`)
- **Inflated sample rate** — each batch rolled sampling independently, so a 25% detector fired at `1-(1-p)^N`
- **False findings on incomplete traces** — for early-finishing/streaming traces (the root span exports before its children) evaluation ran on a partial trace

## What this does

Separates the two questions the old path conflated — *when to enqueue* (exactly once) and *when to evaluate* (only once the trace has settled):

- **Enqueue exactly once.** Only the batch carrying the trace's root span claims it via a Redis `SET NX` lock and enqueues a single delayed BullMQ job. The trigger conditions + a hash-deterministic sampling decision (`hash(trace, detector) < p`, idempotent across batches/retries) run under the lock; later batches are a no-op.
- **Evaluate after quiescence.** The worker evaluates a trace once **no span has arrived for `EVALUATOR_DELAY` (60s)**. When the job fires it reads `last_arrival_age`; if `>= EVALUATOR_DELAY` it evaluates, otherwise it reschedules the job to exactly `last_span + EVALUATOR_DELAY` and re-checks (a quiescence debounce).

## How it works

- **Enqueue** (`backend/worker/detector_tasks.py`): root-bearing batch takes the NX lock on `(project, trace)`, evaluates trigger conditions + deterministic sampling, and enqueues one delayed job (`delay = EVALUATOR_DELAY`) with a deterministic `jobId` via the official `bullmq` client. Token-checked Lua release on failure so a failed attempt never deletes a successor's state.
- **Worker** (`frontend/worker/src/processors/detector-run-processor.ts`): `handleDetectorRunJob` fetches `last_arrival_age` and either evaluates (`>= EVALUATOR_DELAY`) or `moveToDelayed(now + (EVALUATOR_DELAY - last_arrival_age))` + `DelayedError`. The TS worker re-reads the full span set at execution time.
- **settle-status** (`backend/rest/routers/internal.py`): returns only `last_arrival_age_seconds` = `now64() − max(ch_create_time)`, computed inside ClickHouse to avoid cross-service clock skew.
- **Reads** dedupe pre-merge `ReplacingMergeTree` rows with `FINAL`.

## Guarantees

- **Exactly once per trace** — NX lock (only the root batch enqueues, single atomic winner) + deterministic `jobId` (BullMQ dedup) + deterministic `runId`/`findingId` (a retry overwrites the same rows, never duplicates).
- **No early evaluation** — the worker's only path to evaluation is gated on `last_arrival_age >= EVALUATOR_DELAY`; nothing evaluates before 60s of quiet.

## Constant

`EVALUATOR_DELAY = 60_000` (ms) — canonical in `frontend/worker/src/queues/detector-run-queue.ts`, mirrored in `backend/worker/detector_tasks.py`. It is both the initial enqueue delay and the quiet-window threshold.

## Tradeoffs

- **No structural completeness check** — relies purely on 60s of quiet; a dropped mid-tree span just means a best-effort evaluation after the window.
- **Post-eval stragglers are not re-evaluated** — a batch arriving >60s after evaluation is ignored; the quiet window makes this rare.
- **Unbounded quiet wait** — a trace that receives spans forever never evaluates (no absolute cap).

## Video

https://github.com/user-attachments/assets/6ea244c8-68eb-4fd8-9eed-3bf2b2bae22e

https://github.com/user-attachments/assets/ee87b35c-1cf1-420c-bdf8-bba3be5086e3

## Validation

Backend + worker test suites updated; the enqueue-once and quiescence invariants are covered by unit tests (`test_non_root_batch_enqueues_nothing`, the worker's evaluate-vs-redelay test). Exercised end-to-end on a live stack across **11 traces (Python + TypeScript, streaming + non-streaming, ~7 frameworks)**:

- every trace evaluated **63–67s after its last span** (= 60s quiescence + eval) — **0 early evaluations**;
- **exactly one run per (trace, detector) and one finding per trace**;
- the early-root **streaming** case (root ingests ~9s before its last span) correctly waited 60s after the **last** span, not the early root.

Closes #1178
Closes #1177


ka1kqi:
Addressed the RCA suppression finding in `b32b591b`.

The re-eval path no longer suppresses RCA unconditionally — it now suppresses only when a `DetectorRca` already exists for the finding (`prisma.detectorRca.findUnique` guard); a re-eval that triggers for the first time (clean initial eval, then late spans reveal a problem) falls through and creates the RCA. Added a regression test for that path (`detector-settle-gate.test.ts`), and the `findUnique` short-circuits on `isReeval` so the normal first-eval path does no extra query.

@cubic-dev-ai please re-review.

ka1kqi:
@cubic-dev-ai please re-review — the design changed since your last pass.

**What's different:** the multi-condition settle gate (bounce/backoff/dangling-parent/cap/no-progress) and the single re-eval + retraction are **removed** in favor of a simple quiescence debounce: evaluate a trace once no span has arrived for `EVALUATOR_DELAY` (60s). The Redis `SET NX` enqueue lock and deterministic sampling are kept. The PR description has been rewritten to describe only the current implementation.

**This commit (2fe73e76):** fixes the failing Frontend Tests — the new worker test's mocks returned `undefined` where the code chains `.catch`; they now resolve. Full worker suite passes locally (102/102).

Validated end-to-end on a live stack: every trace evaluated ~60s after its last span (0 early evals), exactly one run per (trace, detector) and one finding per trace, including the early-root streaming case.

ka1kqi:
@cubic-dev-ai addressed your finding in `61aeefc6`.

On the `DelayedError` → `failed` handler comment: strictly it's a false positive — BullMQ special-cases `DelayedError` (alongside `WaitingChildrenError`/`RateLimitError`) in the worker's failure path and returns early without `moveToFailed` or emitting `failed`, so routine re-delays don't actually reach the handler. That said, the guard is a cheap, explicit invariant and future-proofs against version edge cases, so I added it:

```ts
worker.on("failed", (job, err) => {
  if (err instanceof DelayedError) return; // re-delay signal, not a failure
  console.error(`[Detector] Job ${job?.id} failed:`, err.message);
});
```

Please re-review.

XinweiHe:
```
  1. Stale comments describing the removed re-eval path

  The re-eval mechanism is gone, but a couple of comments still describe it:

  - backend/worker/ingest_tasks.py — the new comment says "Later batches for an already-known trace only feed the bounded re-eval check (at most one
  re-eval when the span count grew after evaluation)." That's no longer true — non-root batches are dropped outright (root_traces = [t for t in trace_ids
  if t in traces_with_root]). This also contradicts the comment in detector_tasks.py that correctly says "nothing reads it back now that re-eval is
  gone."
  - tests/worker/test_detector_tasks.py:1606 has a leftover # ── Re-eval path: late batches ── section header with no tests under it.

  Could you reconcile the comment with the actual root-only / quiescence model and drop the dead header?

  2. BullMQ job has attempts: 3 but no backoff

  In _add_bullmq_job, a settle-status fetch failure (or eval failure) throws a real error and consumes an attempt. With no backoff configured, the 3
  attempts retry immediately — so a brief backend/ClickHouse blip can burn all of them in milliseconds and the trace is silently never evaluated. Suggest
  adding an exponential backoff, e.g.:

  "backoff": {"type": "exponential", "delay": 5000},

  (and bumping attempts if 3 feels low for a transient-failure window). The expected-opts assertion in tests/worker/test_detector_tasks.py will need the
  same update.
  ```

**Comment by ka1kqi** [comment]
@cubic-dev-ai please re-review.

Addressed @XinweiHe's review in `7712485a`:

1. **Reconciled the stale re-eval comments.** The non-root-batch comment in `ingest_tasks.py` now matches the actual model (non-root batches are ignored; the worker waits out the quiescence window before evaluating). Dropped the empty `# Re-eval path: late batches` test section header, and fixed a now-stale `(BullMQ attempts: 3)` aside in the worker processor comment.

2. **Added retry backoff to the BullMQ detection job.** The worker throws on a transient settle-status fetch failure and relies on these retries, but with no `backoff` configured the attempts fired immediately — a brief backend/ClickHouse blip could burn all of them in milliseconds and the trace would silently never evaluate. Added `backoff: {type: exponential, delay: 5000}` and raised `attempts` 3→5 (retries at ~5/10/20/40s, ~75s window). Test assertion updated to match.

Worker test suite passes locally.

File: backend/worker/otel_transform.py [hunk 1]
```diff
@@ -142,13 +142,21 @@ def decode_otel_id(b64_value: str | None) -> str | None:
...
-        Hex string representation, or None if input is None/empty
+        Hex string representation, or None if input is None/empty or the ID is
+        all zero bytes
     """
     if not b64_value:
         return None
     try:
         decoded = base64.b64decode(b64_value)
-        return decoded.hex()
+        hex_id = decoded.hex()
```
XinweiHe, line 152: for my understanding what is the point of this change?

backend/worker/detector_tasks.py
```diff
@@ -139,45 +217,73 @@ def _get_active_detectors(project_id: str) -> list[dict]:
...
-def _enqueue_to_bullmq(redis_client, queue_name: str, job_id: str, data: dict) -> None:
-    """
-    Enqueue a job to a BullMQ queue via Redis.
+def _claim_and_enqueue(
+    redis_client,
+    project_id: str,
+    trace_id: str,
+    detectors: list[dict],
+    summary: dict,
+) -> None:
```
XinweiHe, line 286: add doc string for input and return type?

backend/rest/routers/internal.py
```diff
@@ -510,11 +547,14 @@ async def list_detector_findings(
     dependencies=[Depends(verify_internal_secret)],
 )
 async def get_trace_findings(trace_id: str, project_id: str):
-    """List all detector findings for a specific trace."""
+    """List all detector findings for a specific trace.
```
XinweiHe: docstrings

backend/worker/detector_tasks.py
```diff
@@ -25,6 +51,60 @@ def _get_redis():
     return redis.from_url(celery_app.conf.broker_url)
 
 
+def _lock_key(project_id: str, trace_id: str) -> str:
```
XinweiHe, line 54: docstrings

backend/worker/detector_tasks.py
```diff
@@ -25,6 +51,60 @@ def _get_redis():
...
+    return f"detector-enq:{project_id}:{trace_id}"
+
+
+def _release_lock_if_value(redis_client, key: str, expected: str) -> None:
```
XinweiHe, line 70: docstrings

backend/worker/detector_tasks.py
```diff
@@ -25,6 +51,60 @@ def _get_redis():
...
+    redis_client.eval(_RELEASE_IF_VALUE_LUA, 1, key, expected)
+
+
+def _sample_passes(trace_id: str, detector_id: str, sample_rate: float) -> bool:
```
XinweiHe: docstrings

XinweiHe: also guard the sample rate

backend/worker/detector_tasks.py
```diff
@@ -25,6 +51,60 @@ def _get_redis():
...
+    """Deterministic per-(trace, detector) sampling decision.
+
+    Hash-based rather than random.random() so the decision is idempotent
+    across batches and Celery retries — a replay can never re-roll the dice.
+    """
+    digest = hashlib.sha256(f"{trace_id}:{detector_id}".encode()).digest()
+    return int.from_bytes(digest[:8], "big") / 2**64 < sample_rate / 100.0
+
+
+def _add_bullmq_job(job_id: str, data: dict) -> None:
```
XinweiHe, line 121: docstrings

backend/worker/ingest_tasks.py
```diff
@@ -105,6 +105,8 @@ def process_s3_traces(self, s3_key: str, project_id: str) -> dict:
         traces, spans = transform_otel_to_clickhouse(otel_data, project_id)
         logger.info(f"Transformed {len(traces)} traces and {len(spans)} spans from {s3_key}")
 
+        traces_with_root = {s["trace_id"] for s in spans if s.get("parent_span_id") is None}
+
```
XinweiHe, line 109: Just for my understand `traces_with_root` is defined in here but `traces_without_root` is defined later in L120. Is this expected?

backend/worker/otel_transform.py
ka1kqi: some emitters may send parent_id as all zero bytes instead of a null value, meaning it can look like a valid parent span_id. this line is just a guard for those edge cases, since all the downstream root detection relies on trace id being None/NULL, not zeros. can be removed if you think it is confusing since it was working fine before

backend/worker/ingest_tasks.py
ka1kqi: traces_with_root needs to be in scope for both the trace/insert dedup at L120/131, and the detector enqueue at L153, while traces_without_root  is only needed within the if traces block at L114.

ka1kqi:
@cubic-dev-ai, rereview please, added docstrings

backend/worker/detector_tasks.py
```diff
@@ -3,18 +3,44 @@
...
+import uuid
 
 logger = logging.getLogger(__name__)
 
 # BullMQ queue name — must match TypeScript DETECTOR_RUN_QUEUE constant
 DETECTOR_RUN_QUEUE = "detector-run"
 
+# Lock TTL for the per-trace enqueue claim. Detection only ever fires from the
+# root-bearing batch; the NX lock makes that enqueue exactly-once.
+_LOCK_TTL_SECONDS = 3600
```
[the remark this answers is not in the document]
ka1kqi: Intentional tradeoff — not changing it. The 1h TTL is a dedup *cache*, not a permanent ledger; durable exactly-once comes from elsewhere:

- `runId` (`deterministicRunId(project, trace, detector)`) and `findingId` (`traceFindingId(project, trace)`) are deterministic hashes, and `detector_runs` / `detector_findings` are ReplacingMergeTree keyed on them — a re-run **overwrites the same rows** rather than producing duplicate runs or findings.
- RCA is an upsert on `findingId` plus a `rca-${findingId}` jobId, so it doesn't duplicate either.

So a late root replay can't produce duplicate detector *data*. The only residual on lock expiry is the LLM cost of one re-evaluation, and only if a root span is genuinely re-delivered **>1h later** (past the S3/Celery retry horizon) *and* the original BullMQ job has already been evicted (`removeOnComplete: 100`) — a narrow edge.

Removing or lengthening the TTL doesn't fix that case; it just shifts the window and leaks a Redis key per trace forever. The correct guard, if we ever decide the rare wasted compute is worth closing, is short-circuiting evaluation when a finding already exists for (project, trace) — a separate change from this PR.

backend/rest/routers/internal.py
```diff
@@ -416,6 +416,38 @@ def _default(obj):
     return PlainTextResponse("\n".join(lines))
 
 
+class TraceSettleStatusResponse(BaseModel):
```
XinweiHe: delete it.

backend/rest/routers/internal.py
XinweiHe: make sure to add docstring

backend/worker/detector_tasks.py
```diff
@@ -25,6 +51,85 @@ def _get_redis():
...
+    """
+    redis_client.eval(_RELEASE_IF_VALUE_LUA, 1, key, expected)
+
+
+def _sample_passes(trace_id: str, detector_id: str, sample_rate: float | None) -> bool:
+    """Deterministic per-(trace, detector) sampling decision.
+
+    Hash-based rather than random.random() so the decision is idempotent
+    across batches and Celery retries — a replay can never re-roll the dice.
+
```
XinweiHe, line 95: add variable types. see change below.

backend/worker/detector_tasks.py
```diff
@@ -25,6 +51,85 @@ def _get_redis():
...
+    Args:
+        trace_id: Trace being considered.
+        detector_id: Detector whose sampling is being rolled.
```
XinweiHe:
```suggestion
        detector_id (str): Detector whose sampling is being rolled.
```

backend/worker/ingest_tasks.py
```diff
@@ -138,16 +139,18 @@ def process_s3_traces(self, s3_key: str, project_id: str) -> dict:
...
+        # duplicate root delivery. Batches without the root span are ignored
+        # here; the worker waits out the quiescence window before evaluating,
+        # so late spans need no enqueue.
         if traces or spans:
             try:
                 from worker.detector_tasks import enqueue_detector_runs
 
                 trace_ids = list({t["trace_id"] for t in traces} | {s["trace_id"] for s in spans})
-                enqueue_detector_runs(project_id, trace_ids)
+                enqueue_detector_runs(project_id, trace_ids, traces_with_root)
```
XinweiHe: if we just ignore, maybe we should remove `traces_with_root` from entering the function?

backend/worker/otel_transform.py
XinweiHe: okay.

frontend/worker/src/detection/sandbox-eval.ts
```diff
@@ -159,6 +159,20 @@ RULES:
 - summary must be one sentence. If identified=true, describe what you found. If false, state why it is clean.
 - data fields are only required when identified=true.`;
 
+  // Disclose incomplete input to the judge so it doesn't infer problems from
```
XinweiHe: remove this mechanism

frontend/worker/src/processors/detector-run-processor.ts
```diff
@@ -18,11 +19,23 @@ const BACKEND_URL = process.env.BACKEND_INTERNAL_URL || "http://localhost:8000";
...
- * Deterministic run id keyed on (projectId, traceId, detectorId).
- * On a BullMQ retry, the same triple lands on the same runId — so re-writes
- * collapse with detector_findings.findingId rather than producing duplicate
- * run rows for the same (detector, trace).
+ * Milliseconds since the most recent span of a trace was ingested. The worker
+ * evaluates once this reaches EVALUATOR_DELAY — i.e. the trace has been quiet
+ * that long (a simple quiescence debounce). The age is computed inside
+ * ClickHouse (now64() vs max(ch_create_time)) to avoid cross-service clock skew.
  */
+async function fetchLastArrivalAgeMs(projectId: string, traceId: string): Promise<number> {
```
XinweiHe:
```suggestion
async function fetchLastArrivalTimestampMs(projectId: string, traceId: string): Promise<number> {
```

frontend/worker/src/processors/detector-run-processor.ts
```diff
@@ -405,23 +441,53 @@ async function processTrace(
...
+
+  // Not quiet yet — sleep until exactly EVALUATOR_DELAY after the most recent
+  // span, then re-check. moveToDelayed + DelayedError is BullMQ's supported
+  // re-delay pattern; unlike a thrown failure it does not consume an attempt.
+  const delayMs = EVALUATOR_DELAY - lastArrivalAgeMs;
+  console.log(
+    `[Detector] settle-wait trace=${traceId} quiet=${Math.round(lastArrivalAgeMs / 1000)}s ` +
+      `re-check in ${Math.round(delayMs / 1000)}s`,
+  );
+  await job.moveToDelayed(Date.now() + delayMs, token);
```
XinweiHe, line 477: Check whether there are timezone issues with this. Do we have unit tests covering this?

XinweiHe:
add a video for trace listing page + detector page.

ka1kqi:
@cubic-dev-ai Intentional — the `incompleteNote` was removed at a reviewer's request. The judge prompt already instructs the model to only report problems visible in the spans shown, and truncation at the 150k-char cap is rare, so the note added prompt noise without enough upside. Keeping it removed.

frontend/worker/src/processors/detector-run-processor.ts
ka1kqi: no issues, timezone is standardized to utc, timezone only affects date formatting. Date.now() returns UTC epoch milliseconds

ka1kqi:
@cubic-dev-ai please re-review.

Since the last review (`736a0524`):
- `f88bfa99` — simplified `enqueue_detector_runs` to take only `traces_with_root` (dropped the redundant `trace_ids` arg + no-op intersection; behaviour-preserving, tests updated).
- `e5df8ae5` — removed the single-field `TraceSettleStatusResponse` model; the settle-status endpoint now returns a plain dict (wire format unchanged).

Both address review feedback; backend + worker suites pass.

ka1kqi:
@cubic-dev-ai

backend/worker/ingest_tasks.py
```diff
@@ -114,7 +116,6 @@ def process_s3_traces(self, s3_key: str, project_id: str) -> dict:
                 # OR the trace is genuinely new (no existing ClickHouse record).
                 # Intermediate batches without the root span must not overwrite a
                 # correctly-named trace record with a wrong name.
-                traces_with_root = {s["trace_id"] for s in spans if s.get("parent_span_id") is None}
                 traces_without_root = [t for t in traces if t["trace_id"] not in traces_with_root]
```
XinweiHe: variable naming here is a bit harder to follow. `traces_without_root` is extracted from `traces_with_root`. With `traces_with_root`, I think developer would get the impression that these traces are with root?

backend/rest/routers/internal.py
```diff
@@ -416,6 +416,42 @@ def _default(obj):
     return PlainTextResponse("\n".join(lines))
 
 
+@router.get(
+    "/traces/{trace_id}/settle-status",
+    dependencies=[Depends(verify_internal_secret)],
+)
+async def get_trace_settle_status(trace_id: str, project_id: str):
```
XinweiHe: is this api not used in anywhere?

XinweiHe:
For comments that are outdated or resolved. Feel free to click the "resolve" button in github.

backend/rest/routers/internal.py
ka1kqi: its used in fetchLastArrivalTimestampMs within detector-run-processor.ts at L30

backend/worker/ingest_tasks.py
ka1kqi: renamed

backend/rest/routers/internal.py
ka1kqi: renamed the function and endpoint to be more descriptive, old name is an artifact of old debouncing logic

ka1kqi:
@cubic-dev-ai review

XinweiHe (approved):
LGTM

PR #1191, opened 2026-06-15
fix: restrict auth redirect targets

weijun-xia:
## Summary
- Add a shared redirect sanitizer for post-auth return paths.
- Apply it to sign-in callback URLs and GitHub login/install callback flows.
- Prevent external, protocol-relative, and malformed redirect targets from escaping the app origin.

## Test plan
- IDE diagnostics: no errors
- Not run: dependencies are not installed locally; frozen pnpm install is blocked by lockfile/overrides mismatch.

---
## Summary by cubic
Adds a shared redirect sanitizer across sign-in and GitHub auth to block open redirects and keep users on our origin.

- **Bug Fixes**
  - Add `sanitizeRedirectPath` (accepts only path-absolute "/..."; else falls back to "/"); apply to sign-in `callbackUrl` and GitHub `returnTo` in `api/github/{login,install,install-callback,callback}`.
  - Add unit tests for the helper and the install-callback route; make the install-callback test self-contained to avoid `vi.mock` hoisting issues.

<sup>Written for commit 3352c3784a53698ef05a94942e42dbbf0c0f7c2c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1191?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/ui/src/app/api/github/callback/route.ts
+++ b/frontend/ui/src/app/api/github/callback/route.ts
@@ -2,6 +2,7 @@ import { NextRequest, NextResponse } from "next/server";
 import { env } from "@/env";
 import { prisma } from "@traceroot/core";
 import { requireAuth, requireWorkspaceMembership } from "@/lib/auth-helpers";
+import { sanitizeRedirectPath } from "@/lib/safe-redirect";
 import {
   GITHUB_AUTH_STATE_COOKIE,
   GITHUB_INSTALL_STATE_COOKIE,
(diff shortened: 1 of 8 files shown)


XinweiHe:
Would be cool if we can add a bit more unit testing.

weijun-xia:
@XinweiHe added a bit more unit testing in 657141b4:

- `safe-redirect.test.ts` — covers `sanitizeRedirectPath`: same-origin path-absolute URLs (incl. query/hash) pass through; external, protocol-relative (`//`), backslash (`/\`), control-char (CRLF), leading-space, and empty/null/undefined inputs all fall back to `/`.
- `install-callback/route.test.ts` — verifies the install-callback handler redirects to a same-origin `returnTo` and falls back to `/` for an external one, following the existing `callback/route.test.ts` mock pattern.

Happy to extend the same route-level coverage to the login/install handlers if useful.

PR #1203, opened 2026-06-15
fix(frontend): allow clearing pagination page input

### LINKED ISSUE #1198: Page-number input in pagination can't be cleared (always keeps one digit)
## Summary
Customers reported a small UI bug in the pagination control: the **Page** number input field cannot be fully cleared. When trying to delete the digits to type a new page number, the last digit can't be removed — there's always at least one digit left in the box.

## Steps to reproduce
1. Go to any paginated list view (e.g. a list showing "Page 8 of 249").
2. Click into the **Page** number input field.
3. Try to select/backspace to clear the field completely.

## Expected
The field can be cleared to empty so a new page number can be typed (e.g. clear `8` and type `150`). Value is validated/clamped on blur or Enter.

## Actual
The field always retains at least one digit; it cannot be emptied, making it awkward to jump to a different page.

## Likely cause
The page input appears to be a controlled number input that coerces an empty string back to a valid number on every keystroke (e.g. `Number(value) || currentPage`), so the input is never allowed to be empty while editing.

## Suggested fix
Allow the input to hold a transient empty/string value during editing, and only clamp/validate (min `1`, max `totalPages`) on blur or Enter.

## Severity
Low — minor UX annoyance, no data impact.

TanishValesha:
# fix(frontend): allow clearing the pagination page input

Fixes #1198

## Summary

This PR fixes the pagination page-number input so it can be fully cleared while editing. The input no longer forces a digit to remain in the field on every keystroke. Instead, it allows transient empty/string states during typing and only validates/clamps the value on blur or Enter.

## Type of Change

- [x] bug fix

## Approach

The input was fully controlled by `page`, with validation running on every keystroke - any empty/invalid value was dropped, snapping the field back. Fixed by adding local string state for the input's displayed value (decoupled from `page`), so it can be freely cleared/edited. Validation and clamping `(1–totalPages)` now happen only on blur or Enter, with the field resyncing to `page` after commit or on external page changes (nav buttons).

## Details

The change is made in the shared pagination component, `frontend/ui/src/components/list-pagination.tsx`, which is the universal pagination control used wherever list pagination is rendered in the UI.

This component is used by:

- `frontend/ui/src/app/projects/[projectId]/traces/page.tsx`
- `frontend/ui/src/app/projects/[projectId]/users/page.tsx`
- `frontend/ui/src/app/projects/[projectId]/sessions/page.tsx`
- `frontend/ui/src/app/projects/[projectId]/detectors/page.tsx`
- `frontend/ui/src/app/projects/[projectId]/detectors/[detectorId]/page.tsx`
- `frontend/ui/src/app/dev/pagination/page.tsx` for local UI verification

## Changes

- The Page input can now be fully cleared.
- Users can type a new page number after clearing the field.
- Invalid or out-of-range values are handled on blur or when pressing Enter.
- The component remains the shared pagination implementation for all applicable paginated list views.

## Verification and Screenshot

- Added a development-only mock pagination page for reproducing the issue without backend data.

**Empty Page State**
Screenshot from 2026-06-16 01-31-46

**New Entered Page State**
Screenshot from 2026-06-16 01-32-11

## Checklist

- [x] I have added screenshots or recordings for UI changes, if applicable

---
## Summary by cubic
Fix the pagination page input so it can be fully cleared while typing. The field accepts empty text and only commits a clamped page on blur or Enter, staying in sync with nav buttons and parent-driven changes.

- **Bug Fixes**
  - Keep page text in local state and sync from external `page`.
  - On blur/Enter, clamp to 1–totalPages; if empty/invalid, restore the current page. The input reflects the clamped value. Added tests for clearing, empty-blur restore, and out-of-range clamping.

<sup>Written for commit 13b2ccf9b8854d95cc0878b4ccc09f0f901fe236. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1203?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/ui/src/components/list-pagination.test.tsx
+++ b/frontend/ui/src/components/list-pagination.test.tsx
@@ -56,4 +56,24 @@ describe("ListPagination", () => {
     expect(onPageChange).not.toHaveBeenCalled();
     expect(input.value).toBe("8");
   });
+
+  it("clamps an out-of-range page number on blur", () => {
+    const onPageChange = vi.fn();
+    render(
+      <ListPagination
+        page={0}
+        limit={50}
+        total={500}
+        onPageChange={onPageChange}
+        onLimitChange={vi.fn()}
+      />,
+    ); // 500 / 50 = 10 total pages
+
+    const input = screen.getByRole("spinbutton") as HTMLInputElement;
+    fireEvent.change(input, { target: { value: "9999" } });
+    fireEvent.blur(input);
+
+    expect(onPageChange).toHaveBeenCalledWith(9); // clamped to last page (0-indexed: page 10 → index 9)
+    expect(input.value).toBe("10"); // input reflects the clamped value, not 9999
+  });
 });
--- a/frontend/ui/src/components/list-pagination.test.tsx
+++ b/frontend/ui/src/components/list-pagination.test.tsx
@@ -1,0 +1,59 @@
+// @vitest-environment jsdom
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { cleanup, fireEvent, render, screen } from "@testing-library/react";
+import { ListPagination } from "./list-pagination";
+
+afterEach(cleanup);
+
+describe("ListPagination", () => {
+  it("allows the page input to be cleared while editing", () => {
+    const onPageChange = vi.fn();
+    const onLimitChange = vi.fn();
+
+    render(
+      <ListPagination
+        page={7}
+        limit={50}
+        total={12450}
+        onPageChange={onPageChange}
+        onLimitChange={onLimitChange}
+      />,
+    );
+
+    const input = screen.getByRole("spinbutton") as HTMLInputElement;
+
+    expect(input.value).toBe("8");
+
+    fireEvent.change(input, { target: { value: "" } });
+    expect(input.value).toBe("");
+
+    fireEvent.change(input, { target: { value: "150" } });
+    fireEvent.blur(input);
+
+    expect(onPageChange).toHaveBeenCalledWith(149);
+    expect(input.value).toBe("150");
+  });
+
+  it("restores the current page when an empty input is blurred", () => {
+    const onPageChange = vi.fn();
+    const onLimitChange = vi.fn();
+
+    render(
+      <ListPagination
+        page={7}
+        limit={50}
+        total={12450}
+        onPageChange={onPageChange}
+        onLimitChange={onLimitChange}
+      />,
+    );
+
+    const input = screen.getByRole("spinbutton") as HTMLInputElement;
+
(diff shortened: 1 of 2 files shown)


TanishValesha:
Addressed the bug caught by the bot and committed the fix.
@XinweiHe Please review the PR whenever possible.

XinweiHe:
Screenshot 2026-06-15 at 2 46 27 PM


let's increase the test coverage a bit to bypass the check.

TanishValesha:
@XinweiHe  I have the required test coverage.
Thanyou!

ka1kqi:
@TanishValesha thanks for the fix! One thing to flag: the cubic bot's out-of-range clamping finding is marked "✅ Addressed in `c75d11e`", but that commit isn't actually in the branch — the current head (`0f087448`) still has the un-clamped version, where the `Math.min(Math.max(1, val), totalPages)` is dead code (the guarding `if` already constrains `val` to the valid range) and an out-of-range entry resets to the current page instead of clamping.

Could you apply cubic's suggested fix (switch the guard to `if (!isNaN(val))` so the clamp actually runs) and add a test for the out-of-range case (e.g. typing `9999` → snaps to the last page)? Thanks!

TanishValesha:
@ka1kqi thanks for looking into it.
I have committed the suggested changes.

ka1kqi:
@TanishValesha the clamp fix looks good — out-of-range entries now snap to the nearest valid page as intended. One last thing: could you add test coverage for the out-of-range case (e.g. typing `9999` → clamps to the last page) as mentioned before? The current tests only cover the in-range and empty-input paths, so the new clamping behavior is shipping untested. Thanks!

TanishValesha:
> @TanishValesha the clamp fix looks good — out-of-range entries now snap to the nearest valid page as intended. One last thing: could you add test coverage for the out-of-range case (e.g. typing `9999` → clamps to the last page) as mentioned before? The current tests only cover the in-range and empty-input paths, so the new clamping behavior is shipping untested. Thanks!

My bad, forgot to consider that case. 
I have added the out-of-range test coverage
Thanks for pointing out!

ka1kqi:
@TanishValesha thanks for the quick turnaround! Both the clamp fix and the out-of-range test are in now. I pulled the branch and ran the suite locally — all 3 `ListPagination` tests pass, and the clearing/clamping behavior checks out. LGTM 👍

XinweiHe (approved):
LGTM

PR #1215, opened 2026-06-16
fix(api): time out model-provider connection tests

### LINKED ISSUE #1214: Model-provider "Test connection" has no timeout: a hung or unreachable provider URL hangs the request handler with no feedback
## Summary

The model-provider connectivity test issues raw provider `fetch()` calls with no timeout or `AbortSignal`. A provider endpoint — or a user-supplied `baseUrl` — that accepts the TCP connection but never responds leaves the `fetch` pending forever. Because this runs inside the Next.js route handler (not a background worker), a hung call holds the request open indefinitely, and the "Test connection" button in the UI stays in its loading state with no error and no timeout.

This is the same unbounded-hang class as the detector LLM eval (its sibling), in a different subsystem: a UI API route rather than the BullMQ worker.

## How to replicate

1. In **Workspace Settings → Model Providers**, add a BYOK provider for an adapter that accepts a custom base URL (e.g. OpenAI-compatible), with **Base URL** pointing at an endpoint that accepts the connection but never responds (e.g. a server that holds the socket open without replying).
2. Click **Test connection**.
3. The request hangs indefinitely — the button stays loading, and no error or timeout is ever surfaced.

## Root cause (confirmed)

No layer bounds the request time:

1. **The endpoint issues raw, untimed `fetch()` calls.** `frontend/ui/src/app/api/workspaces/[workspaceId]/model-providers/test/route.ts` makes one provider request per adapter with no `signal`/timeout, e.g.:
   ```ts
   const res = await fetch(url, {
     headers: { Authorization: `Bearer ${apiKey}` },
   });
   ```
   There are **9 such call sites** (openai, anthropic, google, azure, deepseek, openrouter, xai, moonshot, zai).
2. **The target host is user-controlled for several adapters.** For openai/azure/deepseek/xai/moonshot/zai the URL is built from the provider's `baseUrl` field, so the user decides whether the endpoint responds.
3. **Node `fetch` (undici) has no total-request timeout** — a stalled socket or a body that stops mid-flight never settles the promise.

### Why this isn't caught by the existing handling

The route wraps the checks in a `try/catch` that returns `{ success: false, error }`. That only catches *thrown* failures (DNS error, connection refused, non-OK HTTP status). A socket that connects and then stalls never throws, so the `await` never settles and the `catch` never runs.

## Impact

- **Request-handler exhaustion:** each hung test holds a Next.js request/connection open indefinitely; repeated tests against a stalled endpoint degrade the server itself.
- **No user feedback:** the "Test connection" button spins forever — the user can't distinguish "still checking" from "stuck," and never learns the endpoint is unreachable.
- **User-triggerable:** because `baseUrl` is user-supplied, any workspace admin can trigger the hang without special access.

## Acceptance criteria

- Each provider connectivity check is bounded by a timeout; on expiry it returns `{ success: false, error: "...timed out..." }` instead of hanging.
- A `baseUrl` that accepts the socket but never responds fails the test within the bound (no infinite spinner).
- The timeout is a named, env-configurable value.

## Notes

- 9 network call sites are affected (openai, anthropic, google, azure, deepseek, openrouter, xai, moonshot, zai). The `amazon-bedrock` branch only does credential-format validation and makes no network call.
- Sibling to the detector LLM eval timeout issue — same root cause (an unbounded provider call), tracked and fixed separately because this is a Next.js API route, not the detector worker.

dark-sorceror:
## Summary

Fixes #1214

The model-provider "Test connection" endpoint issued raw provider `fetch()` calls with no timeout or `AbortSignal`. A provider host — or a user-supplied `baseUrl` — that accepts the TCP connection but never responds left the `fetch` pending forever, holding the Next.js request open and leaving the "Test connection" button spinning with no error.

This bounds every connectivity check with a single env-configurable deadline that covers the **whole** logical provider check — not just the initial `fetch()`, but any response read it performs. On timeout the route returns the existing `{ success: false, error }` shape with a clear `Connection timed out after <ms>ms` message.

## How it works

```
POST /api/workspaces/[workspaceId]/model-providers/test
  → resolve adapter + API key (provided, or decrypted from DB by providerId)
    → withTimeout(signal => provider check)        ← one AbortController, one deadline
        → fetch(provider URL, { ...headers, signal })
          → inspect res.ok / res.status            (status only — no body read)
            → success  |  HTTP <status>: <statusText>
      ⟂ deadline elapses → controller.abort() → fetch rejects
          → { success: false, error: "Connection timed out after <ms>ms" }
  finally → clearTimeout  (only after the operation settles)
```

Key invariant: **one deadline bounds the entire provider check, not just the initial `fetch()`** — so a host that returns headers and then stalls the body still fails within the timeout. The timer is cleared in `finally` *after* the operation settles, never before a read.

### Route — `frontend/ui/src/app/api/workspaces/[workspaceId]/model-providers/test/route.ts`
- All 9 network branches (openai, anthropic, google, azure, deepseek, openrouter, xai, moonshot, zai) run their `fetch` inside `withTimeout` with the `AbortSignal` threaded into the request.
- Removed provider error-body parsing (`await res.json()`); connectivity failures are reported by status code (`HTTP <status>: <statusText>`), which removes the body-read hang surface entirely. The `{ success, error? }` shape and the anthropic `401 → "Invalid API key"` semantics are preserved.
- `amazon-bedrock` only validates credential format and makes no network call, so it intentionally stays outside `withTimeout`.
- The route file now exports only its `POST` handler (helpers moved out — see below — to avoid non-standard Next.js route exports).

### Timeout helper — `frontend/ui/src/app/api/workspaces/[workspaceId]/model-providers/test/timeout.ts` (new)
- `withTimeout(operation: (signal) => Promise<T>, timeoutMs?)` — runs the full check under one `AbortController`, clears the timer in `finally`, converts only abort-triggered failures into `Connection timed out after <ms>ms`, and rethrows all other errors unchanged.
- `resolveTimeoutMs()` — parses `MODEL_PROVIDER_TEST_TIMEOUT_MS`: defaults to `10000`; invalid / empty / non-finite / zero / negative values fall back to the default; values are capped at `60000` and floored to whole milliseconds.

### Tests — `frontend/ui/src/app/api/workspaces/[workspaceId]/model-providers/test/route.test.ts` (new, 41 tests)
- Signal-aware hanging `fetch` mock (rejects only on abort) → route returns `{ success: false, error: /timed out/ }`, bounded under the deadline.
- Ordinary `fetch` rejection (e.g. `ECONNREFUSED`) → non-timeout failure, not mislabeled.
- OK → success; non-OK → `HTTP <status>: <statusText>` for every status-code adapter.
- `withTimeout` unit tests: resolves successful ops, clears its timer, throws on a never-settling op, stays armed across multiple awaits (headers-then-hanging-body), rethrows non-abort errors.
- `resolveTimeoutMs` edge cases: default / empty / non-numeric / zero / negative / capped / fractional-floored / valid.
- Per-adapter success paths and the `amazon-bedrock` credential-validation branches.

### Config
| Env var | Default | Behavior |
| --- | --- | --- |
| `MODEL_PROVIDER_TEST_TIMEOUT_MS` | `10000` | Per-check deadline in ms. Invalid / zero / negative / non-finite → default; capped at `60000`; floored to integer. |

## Test plan
- [x] `pnpm test model-providers/test/` → 41 passing
- [x] Hanging host (socket accepts, never responds) → `{ success: false, error: /timed out/ }` within the deadline
- [x] Ordinary fetch rejection (`ECONNREFUSED`) → non-timeout failure
- [x] OK → success; non-OK → `HTTP <status>: <statusText>`
- [x] Env edge cases: invalid/zero/negative → default, huge → capped at 60000, fractional → floored
- [x] `route.ts` and `timeout.ts` at 100% coverage; diff-cover ≥ 80% gate passes
- [x] `eslint` clean on changed files; `tsc` reports no new errors
- [ ] Manual: set `MODEL_PROVIDER_TEST_TIMEOUT_MS` low, click "Test connection" against an unreachable `baseUrl` in the UI, confirm it fails fast with the timeout error

---
## Summary by cubic
Adds a deadline to the model provider “Test connection” API and bounds the whole check (fetch plus error-body reads) so stalled hosts can’t hang the request. Non-OK responses now surface the provider’s own error message when available, otherwise a clear HTTP status.

- **Bug Fixes**
  - Wrapped all provider checks with `withTimeout` (AbortController); one deadline covers fetch and body reads; on timeout returns `{ success: false, error: "Connection timed out after <ms>ms" }`.
  - Restored error-body parsing under the same deadline; falls back to `HTTP <status>: <statusText>` when no message is present.
  - Timeout is 10s by default, configurable via `MODEL_PROVIDER_TEST_TIMEOUT_MS`; safely parsed, floored, capped at 60s, and sub-1ms values are treated as invalid. Wired into `docker-compose.prod.yml` and documented in `.env.example`.
  - Applied to OpenAI, Anthropic, Google, Azure, DeepSeek, OpenRouter, xAI, Moonshot, and ZAI; Bedrock remains format-only (no network call).
  - Anthropic: 401 → “Invalid API key” (uses provider message if present); other non-401 errors are treated as connectable (success). Azure requires `baseUrl` before any network call.

<sup>Written for commit 25f93e2fbebaa431010ca6b40e31575e7905522d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1215?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


dark-sorceror:
The added `withTimeout()` bounds the full provider-check operation, not only the initial `fetch()`. Provider response-body parsing was removed from this route so a server that sends headers and then stalls the body cannot hang the test connection request.

dark-sorceror:
@cubic-dev-ai review

dark-sorceror:
**E2E Test**
image

- Default connection timeout set to 10 seconds

dark-sorceror:
@cubic-dev-ai review

XinweiHe:
Core change looks solid — `withTimeout`, abort handling, error separation, and the 41 tests are clean and CI is green. Two notes before merge:

**1. (should-fix) `MODEL_PROVIDER_TEST_TIMEOUT_MS` is inert in prod.** The `web` service in `docker-compose.prod.yml` uses an explicit `environment:` allowlist (no `env_file`), and this var isn't listed there or in `.env.example` — so in prod the route always uses the hardcoded 10s default. The hang fix still works (default applies), but the *configurable* part of the AC doesn't. Either wire it in (mirror `DETECTOR_EVAL_TIMEOUT_MS` at `docker-compose.prod.yml:341` + `.env.example:128`):
```yaml
MODEL_PROVIDER_TEST_TIMEOUT_MS: ${MODEL_PROVIDER_TEST_TIMEOUT_MS:-10000}
```
…**or** drop the env var and use a plain `const` — a connection test is unlikely to need per-deploy tuning. Either is fine; it just can't ship as config that does nothing.

**2. (optional) Provider error detail regression.** Dropping `await res.json()` means users see `HTTP 401: Unauthorized` instead of the provider's real message ("Incorrect API key provided", quota text). Rationale is sound, but the body read can live *inside* the `withTimeout` lambda so it stays bounded and keeps the message:
```ts
const { res, errMsg } = await withTimeout(async (signal) => {
  const res = await fetch(url, { headers: { Authorization: `Bearer ${apiKey}` }, signal });
  const errMsg = res.ok ? null : (await res.json().catch(() => ({})))?.error?.message;
  return { res, errMsg };
});
```
Minor: `res.statusText` is empty on HTTP/2 (undici defaults to HTTP/1.1 so it's fine today, just fragile).

Everything else LGTM.

dark-sorceror:
@cubic-dev-ai review

dark-sorceror:
image

- Default is 10 seconds
- Set `MODEL_PROVIDER_TEST_TIMEOUT_MS=2000` in `.env` and redirects timeout duration

dark-sorceror:
Changed the guard in `timeout.ts` from parsed `<= 0` to parsed `< 1`, so any sub-1ms value (which `Math.floor` would collapse to 0) falls back to the default instead of producing a `setTimeout(abort, 0)` that fires before any fetch. This now honors the "invalid/zero/negative → default" contract for the 0.5 case too.
- Verified: `resolveTimeoutMs("0.5")` → 10000 (added to the edge-case test, renamed to "zero, negative, and sub-1ms").
- I chose default over the alternative `Math.max(1, …)` because a sub-millisecond timeout is meaningless config — falling back to the documented default matches the contract; clamping to `1ms` would silently honor a nonsensical value.
- Stored-key-not-found: added a test where `providerId` is supplied but `findFirst` returns `null` → proceeds with `Authorization: "Bearer undefined"` and `decryptKey` is not called. Locks the (pre-existing) behavior.
- Anthropic stalled-body: added a test where the anthropic 401 body read stalls → still times out, exercising the anthropic branch's own inline `withTimeout/readErrorMessage` (previously only the openai path covered this).
- "stays armed" test name: left as-is. It's a legitimate `withTimeout` unit test, and as you noted, the real `readErrorMessage` path is covered by the integration test. Renaming would be churn for no behavior change.

XinweiHe (approved):
LGTM

PR #1235, opened 2026-06-18
feat: implement error-leaf counting heuristic in trace_reader.py and …

### LINKED ISSUE #1204: fix(api): trace error_count inflates when one error propagates through nested spans
## Problem

On the trace list view, a trace's error count reflects the number of spans in an error state, not the number of actual errors. When a single exception is raised deep in a trace, it unwinds through the nested span context managers and each open span independently records `status=ERROR`. A typical agent failure therefore produces several ERROR spans for one underlying error:

```
root / agent span     → status=ERROR
  └─ LLM call span     → status=ERROR
       └─ tool span    → status=ERROR   ← the one real failure
```

The trace list then shows **3** errors when there was only **1**.

## Root cause

The per-trace error count is a flat count of error-status spans:

```sql
-- backend/rest/services/trace_reader.py:122
countIf(status = 'ERROR') as error_count
```

Each span's status is set independently from its own OTel `status.code` at ingest (`backend/worker/otel_transform.py:596-603`); there's no parent/child reconciliation. So `error_count` answers "how many spans are in an error state" while the UI presents it as "how many errors occurred." Those differ whenever an error propagates up the span tree — which for agent traces is the norm.

The frontend renders the value verbatim (`frontend/ui/src/app/projects/[projectId]/traces/page.tsx:298-305`, via `TraceListItem.error_count`).

Note the trace-level *status* boolean is correct — `countIf(status='ERROR') > 0` (`trace_reader.py:555`). Only the *count* over-reports.

## Why it matters

The error count is a primary signal users read to gauge trace health, and anything downstream that keys off the count inherits the inflation. The count needs to mean "number of distinct errors," independent of how deep in the span tree the error originated or how many ancestors it propagated through.

## Proposed direction

Count **error origins** rather than error states — i.e. the span where the exception was actually raised, not the ancestors that merely re-recorded it. Candidate signals, pending a check of what the SDK emits:

- Spans that carry an exception event / `record_exception` (the origin), vs. ancestors that only inherited `ERROR` status.
- Failing that, a leaf heuristic: count ERROR spans whose children are not all themselves ERROR.

An open question is whether the ingested data reliably distinguishes the origin span from the propagated ones; the fix design depends on confirming that.

## Acceptance criteria

- A trace whose single exception propagates through N nested spans reports an error count of **1**, not N.
- A trace with genuinely independent failures (e.g. two separate failed tool calls) still reports the correct higher count.
- Backend test covering the nested-propagation case and the multiple-distinct-errors case.

ApexYash11:
…add unit tests

Fixes <issue-id>
#1204 

## Summary
The trace list's `error_count` was counting every span in an ERROR state,
not the number of distinct errors. Since OTel context managers mark every
ancestor span ERROR as an exception unwinds, a single failure deep in a
trace (e.g. agent → LLM call → tool call) was reported as 3 errors instead
of 1. This PR changes the count to use an "error leaf" heuristic: only
ERROR spans with no ERROR children are counted, so propagated errors
collapse to their origin while genuinely independent failures still count
separately.

## Type of Change

- [ ] feat - A new feature
- [x] fix - A bug fix

## Details
**Root cause:** `trace_reader.py` used `countIf(status = 'ERROR')` over all
spans in a trace, which counts span-level error *states*, not error
*origins*. The trace-level status boolean (`countIf(status='ERROR') > 0`)
was already correct and is unchanged.

**Fix:** Restructured the `span_agg` CTE in `list_traces()`'s query to
self-join each trace's deduped spans against a per-parent count of ERROR
children (`error_children` CTE), then count only ERROR spans where
`coalesce(child_errors, 0) = 0`. This requires no schema change or
re-ingest — `parent_span_id` was already stored per span.

**Known limitations of the heuristic** (documented inline in the SQL and in
the Python mirror used for testing):
- A parent and child both genuinely failing for unrelated reasons collapses
  to a count of 1, not 2 (the child "hides" the parent in the heuristic).
- A broken error chain (`ERROR → OK → ERROR`, e.g. a `try/except` that
  suppresses the error at one level) overcounts as 2 instead of 1.

Both are edge cases, not regressions versus current behavior — the current
behavior is strictly worse (it overcounts every propagation chain). The
durable fix requires parsing OTel exception events at ingest, which
`otel_transform.py` currently discards entirely; that's scoped as a
follow-up (see #1204 discussion) since it needs an SDK-emission audit
across languages first, not implemented here.

**Testing note:** This repo has no ClickHouse test infrastructure (no
testcontainers, no service container in CI), so the error-leaf logic is
unit-tested via `_count_error_leaves`, a pure-Python function that mirrors
the SQL self-join exactly, rather than testing the SQL directly. Both
implementations are cross-referenced by comment ("CHANGE HERE REQUIRES A
CHANGE THERE") and live in the same file. **This means the SQL itself is
not executed by CI** — manual verification against a real ClickHouse
instance is recommended before merge.

Screenshot 2026-06-18 170614

## Checklist

---
## Summary by cubic
Fixes error overcounting in trace lists by counting error “leaves” (distinct origins) instead of every ERROR span, so propagated exceptions count once. Fixes #1204.

- **Bug Fixes**
  - Updated `list_traces()` SQL: added `deduped_spans` and `error_children` CTEs and now count ERROR spans with no ERROR children to compute `error_count` (no schema change).
  - Added `_count_error_leaves()` helper that mirrors the SQL and unit tests to validate collapsing propagation chains and preserving independent failures.
  - Renamed a variable in `_count_error_leaves()` for clarity; no behavior change.

<sup>Written for commit 9dfd966e1442f269cead9698921510e0d5140c60. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1235?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/rest/services/trace_reader.py
+++ b/backend/rest/services/trace_reader.py
@@ -34,6 +34,32 @@ def span_cost_details(
     return cost_breakdown_from_buckets(get_model_price(model_name), buckets) or {}
 
 
+def _count_error_leaves(spans: list[tuple[str, str | None, str]]) -> int:
+    """Count ERROR spans that have no ERROR child (error-leaf heuristic).
+
+    Mirrors the SQL self-join in the ``span_agg`` CTE of ``list_traces()``.
+    If you change the logic here, change the SQL query too (and vice versa).
+
+    Known limitations (same as the SQL):
+    - Independent parent+child both ERROR collapses to 1 (child hides parent).
+    - Broken chain (ERROR -> OK -> ERROR) overcounts as 2 instead of 1.
+    """
+    parent_to_spans: dict[str | None, list[tuple[str, str | None, str]]] = {}
+    for sid, pid, status in spans:
+        parent_to_spans.setdefault(pid, []).append((sid, pid, status))
+
+    error_count = 0
+    for sid, pid, status in spans:
+        if status != "ERROR":
+            continue
+        # Check if any child of this span is also ERROR
+        children = parent_to_spans.get(sid, [])
+        has_error_child = any(child_status == "ERROR" for _, _, child_status in children)
+        if not has_error_child:
+            error_count += 1
+    return error_count
+
+
 class TraceReaderService:
     """Read traces and spans from ClickHouse."""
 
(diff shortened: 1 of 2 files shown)


ApexYash11:
@ka1kqi would appreciate your feedback on this

XinweiHe:
UI screenshot is needed.

PR #1236, opened 2026-06-18
fix: add trace-start lower bound to span detail query

### LINKED ISSUE #1230: Add a server-derived lower-bound time filter to the trace-detail span read (partition pruning)
## Problem

`get_trace`'s spans query (`backend/rest/services/trace_reader.py:252`) filters `project_id + trace_id` only, with no `span_start_time` bound. The `spans` sort key is `(project_id, span_kind, toDate(span_start_time), span_id)` — `trace_id` is absent — and the table is `PARTITION BY toYYYYMM(span_start_time)`. So every trace open scans **all** of the project's monthly partitions (O(project) — the trace-detail read scaling bottleneck).

## Fix

Add a lower-bound predicate so ClickHouse prunes every partition before the trace:

```sql
WHERE project_id = {project_id:String} AND trace_id = {trace_id:String}
  AND span_start_time >= {trace_start_time:DateTime64(3)} - INTERVAL 1 HOUR
```

**Key simplifier — fully server-side, zero extra cost.** `get_trace` already reads `trace_start_time`: its first query (`trace_reader.py:209-220`) reads the `traces` rollup (`... LIMIT 1`) and stores `trace_start_time` at `row[3]` **before** the spans query runs. Reuse that value as the bound. **No frontend change, no API param, no client hint, no deep-link special case** — every access path (list click, cold deep-link, live refetch) already goes through that rollup read.

## Design

- Pass the already-read `trace["trace_start_time"]` as a parameter to the spans query; add the predicate.
- **Lower-bound only — no upper bound.** Immune to the streaming tail (child spans arriving long after a fast root can't fall off an upper edge that doesn't exist); prunes all pre-trace partitions. The 1-hour lookback absorbs clock skew and any drift between the rollup's `trace_start_time` and `min(span_start_time)` — nearly free, since pruning is at month granularity.
- **Fallback:** if `trace_start_time` is null, run the query unbounded — correctness never depends on the bound.

## Live-safety

The live SSE endpoint (`backend/rest/routers/live.py`) never calls `get_trace`. Deriving the bound server-side (not from a client hint) means the `trace_complete` refetch can't drop spans from a stale value.

## Scope decision — `get_span_io`

`get_span_io` filters by `span_id` but also has no time bound, so it too scans partitions. It does not currently read the rollup. **Recommend leaving it out of this issue** (it returns one row; hover-prefetch hides its latency) and filing a fast-follow if profiling shows span-click is slow — adding a bound there needs a rollup lookup (one extra point query).

## Tests

No backend tests for `get_trace` exist today. Add one:

- Mock the rollup query to return a `trace_start_time`; assert the spans-query params include the bound and the SQL contains `span_start_time >=`.
- Assert the unbounded fallback when `trace_start_time` is null.

## Acceptance

- Spans query includes `span_start_time >= trace_start_time - INTERVAL 1 HOUR`.
- Span set identical to unbounded for representative traces, **including a long-running / streaming trace whose children end hours after the root** (no upper bound → every span, all `>= start`, is included).
- No migration, no frontend change, no API change.

## Notes

- **Effort:** S. Backend-only.
- **Coordination:** edits the same `get_trace` skeleton query as the drop-`FINAL` issue. Logically independent; sequence to avoid a trivial conflict.
- This is a lower-bound-only filter by design. An upper bound is intentionally **not** added — it would reintroduce the risk of dropping late/streaming spans.

gbsierra:
Fixes #1230

## Summary

Adds a server-side lower-bound time filter when loading spans for a trace.

This lets ClickHouse skip old span partitions when opening trace detail pages. This keeps trace opens fast as a project grows.

## Type of Change

- [x] test - Adding missing tests or correcting existing tests

## Details

Reproduced locally with `make dev-lite` by seeding ClickHouse with spans across old and current monthly partitions.

Before the fix, the trace span query did not include `span_start_time`, so ClickHouse could not prune old partitions.

After the fix, the backend uses the already-loaded `trace_start_time` and adds a lower-bound filter. It does not add an upper bound, so late child spans are still returned.

Validated by:

- `uv run pytest tests/rest/test_trace_reader.py -q` - proves the query adds the bound and keeps the null fallback.
- Local REST check against `make dev-lite` - proves the running backend returns the right spans.
- ClickHouse `EXPLAIN indexes = 1` - proves ClickHouse skips old partitions after the bound is added.

## Checklist

---
## Summary by cubic
Add a lower‑bound time filter to the trace detail spans query so ClickHouse skips old partitions and trace pages load fast. The bound is `trace_start_time` minus 1 hour (normalized to UTC if tz‑aware); it’s skipped when `trace_start_time` is null; there’s no upper bound to include late child spans; tests cover tz‑aware and null cases; comments clarify the `TRACE_SPAN_LOOKBACK_HOURS` rationale and partition pruning.

<sup>Written for commit d36789673119bed50a747abaaa4a91080ef62f36. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1236?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/rest/services/trace_reader.py
+++ b/backend/rest/services/trace_reader.py
@@ -242,7 +242,20 @@ class TraceReaderService:
         # usage_details is kept (small map) to derive cost_details. Duration is
         # derived on the client from start/end so in-progress spans can grow
         # against `now()` for live traces.
-        spans_query = """
+        spans_conditions = [
+            "project_id = {project_id:String}",
+            "trace_id = {trace_id:String}",
+        ]
+        spans_params = {"project_id": project_id, "trace_id": trace_id}
+        if trace["trace_start_time"] is not None:
+            spans_conditions.append(
+                "span_start_time >= {trace_start_time:DateTime64(3)} - INTERVAL 1 HOUR"
+            )
+            spans_params["trace_start_time"] = to_utc_naive(trace["trace_start_time"])
+
+        spans_where_clause = " AND ".join(spans_conditions)
+
+        spans_query = f"""
             SELECT
                 span_id, trace_id, parent_span_id, name, span_kind,
                 span_start_time, span_end_time, status, status_message,
@@ -250,12 +263,12 @@ class TraceReaderService:
                 usage_details,
                 git_source_file, git_source_line, git_source_function
             FROM spans FINAL
-            WHERE project_id = {project_id:String} AND trace_id = {trace_id:String}
+            WHERE {spans_where_clause}
             ORDER BY span_start_time ASC
         """
         spans_result = self._client.query(
             spans_query,
-            parameters={"project_id": project_id, "trace_id": trace_id},
+            parameters=spans_params,
         )
 
         spans = []
--- a/backend/rest/services/trace_reader.py
+++ b/backend/rest/services/trace_reader.py
@@ -246,11 +246,8 @@ class TraceReaderService:
             "metadata": row[10],
         }
 
-        # Fetch span skeletons — omit the large input/output/metadata blobs to
-        # keep the payload lightweight (fetched per-span on demand instead).
-        # usage_details is kept (small map) to derive cost_details. Duration is
-        # derived on the client from start/end so in-progress spans can grow
-        # against `now()` for live traces.
+        # Build the span skeleton query. The optional trace_start_time lower
+        # bound lets ClickHouse prune old span partitions before the trace.
         spans_conditions = [
             "project_id = {project_id:String}",
             "trace_id = {trace_id:String}",
@@ -274,6 +271,11 @@ class TraceReaderService:
 
         spans_where_clause = " AND ".join(spans_conditions)
 
+        # Fetch span skeletons — omit the large input/output/metadata blobs to
+        # keep the payload lightweight (fetched per-span on demand instead).
+        # usage_details is kept (small map) to derive cost_details. Duration is
+        # derived on the client from start/end so in-progress spans can grow
+        # against `now()` for live traces.
         spans_query = f"""
             SELECT
                 span_id, trace_id, parent_span_id, name, span_kind,
--- a/backend/rest/services/trace_reader.py
+++ b/backend/rest/services/trace_reader.py
@@ -7,6 +7,15 @@ from rest.sql_utils import escape_ilike, to_utc_naive
 from worker.tokens.buckets import TokenBuckets
 from worker.tokens.pricing import cost_breakdown_from_buckets, get_model_price
 
+# Lookback for the trace-detail spans query lower bound:
+# span_start_time >= trace_start_time - this value.
+# This gives room for clock skew and small differences between the stored
+# trace_start_time and the true earliest span start time. Increase it if spans
+# can validly start before the stored trace_start_time. Since spans are split by
+# month with toYYYYMM(span_start_time), values under about a month usually skip
+# the same old monthly partitions.
+TRACE_SPAN_LOOKBACK_HOURS = 1
+
 
 def span_cost_details(
     model_name: str | None,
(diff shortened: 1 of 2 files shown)


tests/rest/test_trace_reader.py
```diff
@@ -176,6 +185,43 @@ def side_effect(query, parameters=None):
...
+        service, _ = _make_service(side_effect)
+        result = service.get_trace("proj", "abc123")
+
+        assert result["spans"] == []
+        assert "span_start_time >=" not in captured["spans_query"]
+        assert captured["spans_parameters"] == {
+            "project_id": "proj",
+            "trace_id": "abc123",
+        }
+
```
ka1kqi, line 283: could you add a timezone aware test, making sure to_utc_naive shifts offset aware trace_start_time to UTC rather than stripping timezone?

ka1kqi (changes requested):

ka1kqi (approved):

XinweiHe:
Nice fix — reusing the already-loaded trace_start_time with a lower-bound-only filter (no upper bound, null fallback) is exactly the right approach for pruning the monthly
  partitions. Two small non-blocking suggestions before merge, both about making the intent legible:

  1. Comment the predicate. The "why" is non-obvious and worth spelling out next to the query: (a) partition pruning — the table is PARTITION BY toYYYYMM(span_start_time) and
  trace_id isn't in the sort key, so without a bound every trace open scans all of the project's partitions; (b) lower-bound-only so late/streaming child spans are never
  dropped, and the 1h slack absorbs clock skew / start-time drift (nearly free at month granularity). That context currently lives only in the test/issue.

  2. Extract the 1 HOUR into a named constant (e.g. TRACE_SPAN_LOOKBACK_HOURS = 1) so it's a tunable knob rather than a magic number — and have the test build its expected
  string from that constant instead of hardcoding 1 HOUR.

  Happy to approve once these are in. Thanks for the thorough repro and the EXPLAIN indexes = 1 evidence! 🙏

XinweiHe:
maybe something like
```
  1. Comment above the bound predicate (the if trace["trace_start_time"] is not None: block):

  # Lower-bound the spans scan by the trace's start time so ClickHouse can
  # prune partitions before the trace. The spans table is
  # PARTITION BY toYYYYMM(span_start_time) and trace_id is NOT in the sort key,
  # so without a time bound every trace open scans all of the project's monthly
  # partitions (O(project-age)). Lower-bound only — no upper bound — so late /
  # streaming child spans that finish long after a fast root are never dropped.
  # The lookback slack (see TRACE_SPAN_LOOKBACK_HOURS) absorbs clock skew and any
  # drift between the rollup's trace_start_time and the true min(span_start_time);
  # it's nearly free since pruning is at month granularity. Omitted when
  # trace_start_time is null — correctness never depends on the bound.

  2. Comment on the extracted constant (module level, after the imports):

  # Lookback applied to the trace-detail spans query's lower bound
  # (span_start_time >= trace_start_time - this). Tunable safety margin for clock
  # skew / rollup-vs-span start drift; widen if spans legitimately start before
  # the rollup's trace_start_time. Coarser than it looks — partition pruning is at
  # month (toYYYYMM) granularity, so any value < ~1 month prunes the same partitions.
  TRACE_SPAN_LOOKBACK_HOURS = 1
```

cc. @gbsierra

backend/rest/services/trace_reader.py
```diff
@@ -242,20 +251,43 @@ def get_trace(self, project_id: str, trace_id: str) -> dict | None:
         # usage_details is kept (small map) to derive cost_details. Duration is
         # derived on the client from start/end so in-progress spans can grow
         # against `now()` for live traces.
```
XinweiHe: comments in here need to be updated as suggested in earlier comment?

gbsierra: updated this comment. there might still be something I’m missing but I don’t see other stale comments now

XinweiHe (approved):
LGTM

PR #1254, opened 2026-06-19
perf(worker): refresh model price cache on a TTL

### LINKED ISSUE #1096: perf: add TTL-based refresh for pricing cache
### Problem

The pricing cache in `backend/worker/tokens/pricing.py` is populated once and then retained for the lifetime of the process.

As a result, updates to pricing data in the database are not reflected in running workers until the process is restarted.

This means pricing calculations may continue using stale values even after pricing records have been updated.

### Proposed Solution

Introduce a lightweight TTL-based cache refresh mechanism.

For example:

* Keep the existing in-memory cache
* Associate a timestamp with the cached data
* Refresh the cache after a configurable interval (for example, every 5 minutes)

This preserves the performance benefits of caching while allowing pricing updates to propagate automatically without requiring worker restarts.

### Alternatives Considered

* Continue requiring process restarts when pricing changes
* Remove caching entirely and query the database every time
* Use an external cache invalidation mechanism

A simple TTL-based refresh appears to provide a good balance between freshness and implementation complexity.

### Additional Context

While reviewing pricing.py, I observed that the cache is populated once and subsequently returned without any invalidation or refresh mechanism. This means running processes will continue using the original pricing data indefinitely.

Happy to work on a PR if this aligns with the intended behavior.

- [x] I would like to work/help with this

GouravSingal-code:
Closes #1096

## Problem
The model price cache in `backend/worker/tokens/pricing.py` is populated on first use and then kept for the **entire process lifetime**. As a result, any pricing update in the DB (`standard_models` / `standard_model_prices`) is never picked up by a running worker — the only way to refresh prices is to restart every worker, and until then cost calculations silently use stale values.

## Change
Refresh the cache on a **configurable TTL** (`PRICING_CACHE_TTL_SECONDS`, default `300`):

- A fresh cache is served without touching the DB (unchanged hot-path cost — `get_model_price()` is called per span).
- Once the TTL lapses, **one** caller refreshes under a lock (double-checked); others wait and reuse the result, so there's no refresh stampede.
- If a refresh fails or returns no rows (DB briefly unavailable), the previously loaded prices are **kept** rather than dropping to an empty cache (which would make all costs `None`); the timestamp is left unchanged so the next call retries promptly. First-load failure still returns `[]` and retries, as before.

Also surfaces `PRICING_CACHE_TTL_SECONDS` in `docker-compose.prod.yml` (worker env) and `.env.example` so the interval is tunable per deployment.

## Trade-off
This bounds staleness to **≤ one TTL** instead of "until restart" — it does not make pricing instantly consistent. That's the intended, lightweight middle ground vs. querying the DB per span (too expensive at ingest volume) or event-driven invalidation (LISTEN/NOTIFY — much larger change, possible follow-up). Operators who need fresher prices can lower the TTL.

## Notes for reviewers
- The cache is **per worker process**. The worker runs Celery's default prefork pool, so each child caches independently and refreshes once per TTL — DB load is `cores × replicas` light queries per TTL (negligible).
- The `threading.Lock` is effectively uncontended under the default prefork pool (one task per child at a time); it's kept intentionally so the code stays correct if the pool is switched to `threads`/`gevent`. Zero cost when uncontended.

## Testing
- 4 new unit tests in `tests/worker/tokens/test_pricing.py` (fresh-serve within TTL, refresh-picks-up-new-price after TTL, stale-on-failure, first-load retry).
- `uv run pytest tests/worker` → **289 passed**. `ruff check` clean, `ruff format` applied.
- Validated live in a local Docker stack: confirmed the worker resolves the TTL from env (default 300; `PRICING_CACHE_TTL_SECONDS=120` → `120.0`), prices `gpt-4o` correctly from the live DB (71 rows), and the refresh / stale-on-failure paths behave as designed inside the running worker container.

---
## Summary by cubic
Refreshes the worker model price cache on a configurable TTL so DB pricing updates are picked up without restarts. Addresses #1096 and keeps the hot path unchanged for cost lookups.

- **New Features**
  - TTL-based refresh with a single locked refresher after expiry; other callers wait and reuse.
  - On refresh failure or empty result, keep the last good cache; first-load failure returns empty and retries.
  - Exposes `PRICING_CACHE_TTL_SECONDS` in `docker-compose.prod.yml` and `.env.example` (default 300s).

<sup>Written for commit e5da52ca855365661c85ce76c811169239688f26. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1254?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -46,6 +46,9 @@ INTERNAL_API_SECRET=dev-internal-secret  # CHANGEME in production
 # --- Worker (Python/Celery) ---------------------------------------------------
 POLL_INTERVAL_SECONDS=5
 BATCH_SIZE=100
+# How long model prices are cached before the worker re-reads them from the DB
+# (seconds). Lower = fresher pricing, higher = fewer DB reads. Default 300 (5m).
+PRICING_CACHE_TTL_SECONDS=300
 
 # --- Billing Worker (Node.js) -------------------------------------------------
 BACKEND_INTERNAL_URL=http://localhost:8000  # Python backend URL for internal API calls
--- a/backend/worker/tokens/pricing.py
+++ b/backend/worker/tokens/pricing.py
@@ -2,14 +2,19 @@
 
 Prices are loaded from the ``standard_models`` / ``standard_model_prices``
 PostgreSQL tables (synced from standard-model-prices.json by the TS services).
-The in-memory cache is populated on first call and persists for the process
-lifetime (prices only change on deploy/restart).
+The in-memory cache is populated on first call and refreshed on a TTL so that
+pricing updates in the DB propagate to long-running workers without a restart
+(see issue #1096). The refresh interval is configurable via the
+``PRICING_CACHE_TTL_SECONDS`` environment variable (default 300s).
 """
 
 from __future__ import annotations
 
 import logging
+import os
 import re
+import threading
+import time
 from decimal import Decimal
 
 import psycopg2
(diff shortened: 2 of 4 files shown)


XinweiHe:
CLA needs to be signed.

GouravSingal-code:
@XinweiHe , cla is signed, please check now.

GouravSingal-code:
Hi @XinweiHe , i have already signed the cla, can you check the pr.

PR #1277, opened 2026-06-22
fix: keep live trace tree connected while parents are pending

### LINKED ISSUE #1208: Live trace tree breaks mid-stream: child spans render disconnected from parents (orphan-to-root)
## Summary

During a **live (in-progress) trace**, the Trace Tree renders spans **disconnected from their parents** — child spans (e.g. a LangGraph `__start__` node) appear pinned to the root instead of nested under their real parent. The tree visibly "breaks" mid-stream and only repairs itself once the trace completes and the view refetches. This is most visible with deep agent traces (LangChain/LangGraph examples) but the cause is **structural and SDK/framework-version independent** — it is not specific to any LangChain version.

## Repro

1. Run a deep, multi-step agent example that produces long-lived parent spans over fast leaf spans (e.g. `examples/typescript/langchain/agent.ts`, a LangGraph code agent: `demo_session → langchain_agent → LangGraph → {__start__, planner, coder, executor, summarizer}`).
2. Open the trace in the UI **while it is still running**.
3. Observe: one or more node spans (commonly `__start__`) render as a top-level sibling of the session root, with no connector to their real parent. After the trace finishes, the tree snaps back to fully correct nesting.

## Root cause

The break is a **live-ordering artifact**, confirmed across all four layers:

**1. SDK exports spans on end → parents arrive after children.**
`@traceroot-ai/traceroot` (0.1.6) uses a `BatchSpanProcessor` (5s tick, export on `onEnd`). A parent span doesn't end until all its children do, so the backend genuinely receives **children before parents**. In the langchain example the root `demo_session` runs ~15s and each `langchain_agent`/`LangGraph` runs ~5–8s, while `__start__` is ~8ms — so the window where a child exists but its parent does not is large.

**2. Backend returns a flat span list, no tree, no repair.**
`get_trace()` (`backend/rest/services/trace_reader.py`) returns whatever spans currently exist, ordered by start time. The tree is assembled entirely client-side.

**3. Frontend pins parentless spans to the root.**
`buildSpanTree` (`frontend/ui/src/features/traces/utils/index.ts`):

```ts
const orphans = spans.filter((s) => s.parent_span_id !== null && !spanIds.has(s.parent_span_id));
const topLevel = [...(childrenByParent.get(null) ?? []), ...orphans].sort(...)
```

A child whose parent hasn't arrived yet renders at level 0, sibling to the root — the disconnected node you see.

**4. The existing mitigation is silently disabled on the live path.**
There is already a mitigation, `enrichSpansWithPending` (`utils/index.ts`), which synthesizes muted "pending" placeholder ancestors from the `traceroot.span.ids_path` / `traceroot.span.path` metadata the SDK emits. The SDK metadata is **correct** — verified in ClickHouse, every node span carries a fully-populated `ids_path` (e.g. `__start__` → `[demo_session, langchain_agent, LangGraph]`).

The gap: `get_trace()`'s span skeleton **deliberately omits the `metadata` column** (payload-size optimization), and `enrichSpansWithPending` bails the moment `ids_path` is absent. The code documents the assumption it relies on:

```
// Skeleton spans omit metadata (parseMetadata(null|undefined) → {});
// live SSE spans still carry it, so enrichment keeps working from live events.
```

So any span that was ingested into ClickHouse **before** the SSE stream connected is only ever seen through the metadata-less base fetch — it can't be enriched, and if its real parent is still in-flight it stays orphaned at the root until `trace_complete` triggers a full refetch.

## Why it's framework/version-independent

Nothing here depends on LangChain/LangGraph internals or version. It's the combination of: deep, long-lived parent spans over millisecond leaf spans + export-on-end batching + a metadata-stripped base fetch. Any agent framework with this span shape breaks the same way. Upgrading LangChain does not fix it.

## Affected files

- `frontend/ui/src/features/traces/utils/index.ts` — `buildSpanTree` (orphan-to-root), `enrichSpansWithPending` (bails without `ids_path`)
- `backend/rest/services/trace_reader.py` — `get_trace()` span skeleton omits `metadata`
- `backend/rest/routers/live.py` + `backend/worker/ingest_tasks.py` — live SSE path (publishes full spans with metadata; base fetch does not)

## Priority

P2 — cosmetic/transient (self-heals on completion) but it undermines confidence in the live-tracing feature, and is reliably reproducible with any deep agent trace.

royalpinto007:
## Summary
- include only the tiny span path metadata subset in trace skeleton responses
- keep full per-span input/output/metadata lazy-loaded on demand
- allow existing pending-ancestor enrichment to repair live trees when child spans arrive before parents

Fixes #1208

## Tests
- uv run pytest tests/rest/test_trace_reader.py
- uv run pytest tests/rest/test_traces_router.py tests/rest/test_public_traces_read.py
- pnpm exec vitest run src/features/traces/utils/index.test.ts
- uv run ruff check backend/rest/services/trace_reader.py tests/rest/test_trace_reader.py
- pnpm exec eslint src/features/traces/utils/index.ts

---
## Summary by cubic
Keep live trace trees connected by returning a tiny span-path subset (`traceroot.span.path`, `traceroot.span.ids_path`) in skeletons and preserving it on annotated spans; pending placeholders no longer fake durations or stretch timelines. Public routes still drop the subset unless `metadata` is requested. Fixes #1208.

- **Bug Fixes**
  - Backend: ClickHouse now extracts only the span-path keys the client uses (`ids_path`, `path`) and re-packs them into skeleton `metadata`; `shared/span_attributes.py` adds constants and docs; worker merges span paths into explicit `traceroot.span.metadata` instead of replacing it; public GET/export clear the subset via `drop_span_tree_metadata`; OpenAPI/schema clarify that public routes drop the subset and explain `starts_path` (preserved at ingest and on the live stream, not shipped by the base read yet).
  - Frontend/Tests: Skeleton reloads synthesize ancestors to keep trees connected; pending placeholders report null duration and aren’t marked in-progress; trace duration ignores placeholders so timelines don’t stretch; tests pin server-side extraction (only `ids_path`/`path`), public-route dropping, dashboard preservation, ingest behavior with explicit metadata, and timeline behavior.

<sup>Written for commit cac97009b4d8950f08bd29890fae89eaf5fecb59. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1277?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


royalpinto007:
Added before/after screenshots. On main, a child span whose parent has not arrived yet appears directly under the trace. In this PR branch, the trace detail API keeps the small traceroot.span.ids_path and traceroot.span.path metadata needed by the frontend, so the child is shown under pending parent rows instead of appearing disconnected

Before-
image

After-
image

backend/rest/services/trace_reader.py
```diff
@@ -271,17 +298,17 @@ def get_trace(self, project_id: str, trace_id: str) -> dict | None:
...
+        # fetched on demand instead. usage_details is kept (small map) to derive
+        # cost_details. Duration is derived on the client from start/end so
+        # in-progress spans can grow against `now()` for live traces.
         spans_query = f"""
             SELECT
                 span_id, trace_id, parent_span_id, name, span_kind,
                 span_start_time, span_end_time, status, status_message,
                 model_name, cost, input_tokens, output_tokens, total_tokens,
-                usage_details,
+                usage_details, metadata,
```
dark-sorceror: Metadata holds the entire leftover attribute bag when no explicit metadata is set, so cost scales with two size on the hot trace-open path. Consider extracting in SQL instead, and same for path.

dark-sorceror:
Let's also rebase and resolve merge conflicts.

backend/rest/services/trace_reader.py
dark-sorceror: Resolved in 1a16c55

XinweiHe (approved):
LGTM

PR #1271, opened 2026-06-22
fix(worker): add DeepSeek model pricing

royalpinto007:
## Summary
- add standard pricing entries for `deepseek-v4-flash` and `deepseek-v4-pro`
- map `deepseek-chat` and `deepseek-reasoner` compatibility aliases to the current V4 Flash pricing entry
- add real-pricing regression coverage so DeepSeek traces no longer get missing/$0 costs

## Pricing source
DeepSeek official models and pricing page: https://api-docs.deepseek.com/quick_start/pricing

## Tests
- `python -m json.tool frontend/packages/core/src/standard-model-prices.json >/tmp/traceroot-prices.json`
- `uv run pytest tests/worker/tokens/test_pricing.py`
- `uv run ruff check tests/worker/tokens/test_pricing.py backend/worker/tokens/pricing.py`

Closes #1268

---
## Summary by cubic
Adds pricing for `deepseek-v4-flash` and `deepseek-v4-pro`, and fixes zero-cost DeepSeek traces by mapping aliases and adding regression tests.

- Bug Fixes
  - Added standard pricing entries for `deepseek-v4-flash` and `deepseek-v4-pro` (based on https://api-docs.deepseek.com/quick_start/pricing).
  - Mapped `deepseek-chat` and `deepseek-reasoner` to `deepseek-v4-flash` via match patterns, including `deepseek/` prefixes and date suffixes.
  - Added tests to validate model ID matching and cost calculation, preventing missing/$0 costs for DeepSeek traces.

<sup>Written for commit b6468b45014cf3a06c5fbfe522fd6532efb55759. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1271?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #1251, opened 2026-06-19
feat: add support for GLM-5.2 (Zhipu AI)

Ameysr:
## Summary

Adds support for GLM-5.2 — Zhipu AI's latest flagship model — across the
model catalog, cost-tracking config, and pricing tests.

Closes #1247

## Type of Change

- [x] feat - A new feature

## Details

**What changed:**

- `standard-model-prices.json` — new entry `tr-glm-5-2` with regex pattern
  that matches both bare (`glm-5.2`) and provider-prefixed (`zai/glm-5.2`)
  model IDs
- `llm-providers.ts` — `glm-5.2` added at the top of the `zai` adapter
  catalog so it appears as the default/first option in the BYOK model selector
- `test_pricing.py` — `TestGLMModelIds` class with parametrized cases
  covering both ID formats

**Pricing note:**

Official GLM-5.2 API pricing has not been published by Zhipu at launch
(open.bigmodel.cn/pricing requires JS and the standalone API rates were not
listed). Prices used are estimated from the flagship tier (¥8/1M input,
¥25/1M output) converted at 7.2 CNY/USD:

- Input: $1.1111/1M tokens → `0.0000011111` per token
- Output: $3.4722/1M tokens → `0.0000034722` per token

**TODO:** update prices once Zhipu publishes official GLM-5.2 API rates.

## Checklist

- [x] I have read the [CONTRIBUTING.md](../CONTRIBUTING.md)
- [x] I have added/updated tests where applicable
- [x] I have added screenshots or recordings for UI changes (if applicable)
- [x] I have updated documentation where necessary

---
## Summary by cubic
Adds support for Zhipu `glm-5.2` in the `zai` model catalog and pricing, and adds cost tracking for the rest of the GLM lineup. Closes #1247.

- **New Features**
  - Added `glm-5.2` to the `zai` adapter catalog (default in BYOK) and pricing entry `tr-glm-5-2` with regex for `glm-5.2` and `zai/glm-5.2`; rates $1.40/$4.40/$0.26 per 1M tokens.
  - Added pricing for existing `zai` GLM models: `glm-5.1`, `glm-5`, `glm-5-turbo`, `glm-4.7`, `glm-4.6`, `glm-4.5`, `glm-4.5-air`, `glm-4.5-flash` (free); rates from Zhipu docs.
  - Extended tests to cover both bare and `zai/`-prefixed IDs for all GLM entries.

[...]

Issue #702, opened 2026-04-20
[Feature] Support new Kimi SOTA model in BYOK chatbot


lqiu03:
### Problem

Kimi 2.6, new SOTA model excels in long horizon agentic tasks (https://www.kimi.com/blog/kimi-k2-6). This feature request aims to support Kimi 2.6 in BYOK chatbot to support TraceRoot user. 

### Proposed Solution

As proposed; consistent with PR610 ( https://github.com/traceroot-ai/traceroot/pull/610)

### Willing to Help Implement?

- [x] I would like to work/help with this

PR #1342, opened 2026-06-25
fix(worker): add xAI Grok model pricing (grok-4, grok-4.20)

Kayvan-Zahiri:
### What

`grok-4` and `grok-4.20` already ship in the BYOK model dropdown (`ADAPTER_MODELS.xai` in `frontend/packages/core/src/llm-providers.ts`) but have **no entries** in `standard-model-prices.json`. So `get_model_price()` falls through to `None` and per-span cost tracking is silently dropped for anyone running xAI models.

### Change

- Add `tr-grok-4` and `tr-grok-4-20` to `standard-model-prices.json`, following the same shape as the recently-merged DeepSeek (#1271) and GLM (#1251) entries.
- Add `XAI_MODEL_CASES` + `TestXAIModelIds` to `tests/worker/tokens/test_pricing.py` covering bare (`grok-4`), provider-prefixed (`xai/grok-4`), and version-suffixed (`grok-4-0709`) ids, plus a cost-calculation test.

### Prices (USD / 1M tokens)

| model | input | output | cached input |
|-------|-------|--------|--------------|
| grok-4 | $3.00 | $15.00 | $0.75 |
| grok-4.20 | $1.25 | $2.50 | $0.20 |

Sourced from the official xAI docs (`docs.x.ai/developers/models`) and cross-checked against OpenRouter. Happy to adjust if you track canonical rates elsewhere.

### Notes

- Match patterns are anchored so `grok-4` does **not** absorb differently-priced variants like `grok-4-fast` (a separate, cheaper xAI model). Added a negative test pinning `grok-4-fast -> None` so a future pattern change can't silently mis-price it.
- Closes the xAI half of the same gap as #702 (Kimi) — happy to send a follow-up for Moonshot pricing.

---
## Summary by cubic
Adds pricing for xAI Grok models (`grok-4`, `grok-4.20`) so worker cost tracking resumes for xAI usage. Previously `get_model_price()` returned `None` for these models; now it returns correct rates.

- Added `tr-grok-4` and `tr-grok-4-20` to `frontend/packages/core/src/standard-model-prices.json` at $1.25 in / $2.50 out / $0.20 cached per 1M tokens, the rates xAI now bills after retiring the original `grok-4` slugs and redirecting them to grok-4.3.
- Match patterns accept bare, `xai/`-prefixed, and version-suffixed IDs, including lettered suffixes like `-0309-reasoning`; explicitly exclude `grok-4-fast`.
- Added tests in `tests/worker/tokens/test_pricing.py` for ID matching, absolute rate pinning, a negative check for `grok-4-fast`, and a `grok-4` cost calculation.

<sup>Written for commit 7f5f187b77ddf1ebd0f6c7fc6a6adbe75b78be04. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1342?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/packages/core/src/standard-model-prices.json
+++ b/frontend/packages/core/src/standard-model-prices.json
@@ -995,5 +995,29 @@
       "cacheRead": 0,
       "cacheWrite": null
     }
+  },
+  {
+    "id": "tr-grok-4-20",
+    "modelName": "grok-4.20",
+    "matchPattern": "(?i)^(xai\\/)?grok-4\\.20(-[\\d-]+)?$",
+    "provider": "xai",
+    "prices": {
+      "input": 0.00000125,
+      "output": 0.0000025,
+      "cacheRead": 0.0000002,
+      "cacheWrite": null
+    }
+  },
+  {
+    "id": "tr-grok-4",
+    "modelName": "grok-4",
+    "matchPattern": "(?i)^(xai\\/)?grok-4(-[\\d-]+)?$",
+    "provider": "xai",
+    "prices": {
+      "input": 0.000003,
+      "output": 0.000015,
+      "cacheRead": 0.00000075,
+      "cacheWrite": null
+    }
   }
 ]
--- a/frontend/packages/core/src/standard-model-prices.json
+++ b/frontend/packages/core/src/standard-model-prices.json
@@ -1095,5 +1095,29 @@
       "cacheRead": 0.0000003,
       "cacheWrite": null
     }
+  },
+  {
+    "id": "tr-grok-4-20",
+    "modelName": "grok-4.20",
+    "matchPattern": "(?i)^(xai\\/)?grok-4\\.20(-[\\d-]+)?$",
+    "provider": "xai",
+    "prices": {
+      "input": 0.00000125,
+      "output": 0.0000025,
+      "cacheRead": 0.0000002,
+      "cacheWrite": null
+    }
+  },
+  {
+    "id": "tr-grok-4",
+    "modelName": "grok-4",
+    "matchPattern": "(?i)^(xai\\/)?grok-4(-[\\d-]+)?$",
+    "provider": "xai",
+    "prices": {
+      "input": 0.000003,
+      "output": 0.000015,
+      "cacheRead": 0.00000075,
+      "cacheWrite": null
+    }
   }
 ]
--- a/frontend/packages/core/src/standard-model-prices.json
+++ b/frontend/packages/core/src/standard-model-prices.json
@@ -1099,7 +1099,7 @@
   {
     "id": "tr-grok-4-20",
     "modelName": "grok-4.20",
-    "matchPattern": "(?i)^(xai\\/)?grok-4\\.20(-[\\d-]+)?$",
+    "matchPattern": "(?i)^(xai\\/)?grok-4\\.20(-[a-z0-9-]+)?$",
     "provider": "xai",
     "prices": {
       "input": 0.00000125,
@@ -1114,9 +1114,9 @@
     "matchPattern": "(?i)^(xai\\/)?grok-4(-[\\d-]+)?$",
     "provider": "xai",
     "prices": {
-      "input": 0.000003,
-      "output": 0.000015,
-      "cacheRead": 0.00000075,
+      "input": 0.00000125,
+      "output": 0.0000025,
+      "cacheRead": 0.0000002,
       "cacheWrite": null
     }
   }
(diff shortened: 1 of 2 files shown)


ka1kqi:
please sign the CLA

Kayvan-Zahiri:
Signed, thanks!

frontend/packages/core/src/standard-model-prices.json
```diff
@@ -1095,5 +1095,29 @@
...
+    "prices": {
+      "input": 0.00000125,
+      "output": 0.0000025,
+      "cacheRead": 0.0000002,
+      "cacheWrite": null
+    }
+  },
+  {
+    "id": "tr-grok-4",
+    "modelName": "grok-4",
```
dark-sorceror, line 1113: xAI publishes no `grok-4` rate, and the `grok-4-*` slugs retired 15 May now redirect to grok-4.3 at $1.25/$2.50. Drop this entry, or reprice it to grok-4.3.

frontend/packages/core/src/standard-model-prices.json
```diff
@@ -1095,5 +1095,29 @@
       "cacheRead": 0.0000003,
       "cacheWrite": null
     }
+  },
+  {
+    "id": "tr-grok-4-20",
+    "modelName": "grok-4.20",
+    "matchPattern": "(?i)^(xai\\/)?grok-4\\.20(-[\\d-]+)?$",
```
dark-sorceror: The suffix group is digits-only, so the published ids — `grok-4.20-0309-reasoning`, `-0309-non-reasoning`, `-multi-agent-0309` — don't match and still bill $0. Widen it to accept the lettered suffixes.

dark-sorceror (changes requested):
The grok-4.20 rates are right and the specificity guard is the best in this batch. Two problems below, both about which ids the entries actually cover.

Kayvan-Zahiri:
Both fixed.

grok-4 now bills at grok-4.3 rates ($1.25/$2.50/$0.20). xAI's May 15 retirement page lists
grok-4-0709 as redirecting to grok-4.3 at those input and output rates. I kept the entry
instead of dropping it because grok-4 is still in the BYOK list, and without one those spans
would cost $0. Two caveats: that page names grok-4-0709 but not the bare grok-4 alias, and it
gives no cached rate, so $0.20 is grok-4.3's.

The grok-4.20 suffix now takes letters, and the tests cover -0309-reasoning,
-0309-non-reasoning, -multi-agent-0309 and a few of the -latest/-gv2/-beta aliases. Both
tests fail if either change is reverted.

Two things this PR doesn't handle, since the price schema is flat. Once a prompt reaches
200K tokens, xAI bills the whole request at double ($2.50/$0.40/$5.00), so those spans come
out at half their real cost. And batch requests get 20% off. Happy to open an issue for tier
support if that's useful.

frontend/packages/core/src/standard-model-prices.json
dark-sorceror: Fixed. grok-4 now bills at the grok-4.3 rate the retired slugs redirect to, and the grok-4.20 suffix group accepts letters, so `grok-4.20-0309-reasoning`, `-0309-non-reasoning` and `-multi-agent-0309` all resolve at $1.25/$2.50 where they previously billed $0. Probed against the real catalogue.

The absolute-rate assertions are non-vacuous too — drifting either model uniformly by 1.5x reds exactly its own parametrized case and nothing else. Thanks for adding those, they were not requested.

frontend/packages/core/src/standard-model-prices.json
dark-sorceror: Fixed — see above.

dark-sorceror (approved):
Both points fixed and verified independently.

Probed against the real catalogue: `grok-4.20-0309-reasoning`, `-0309-non-reasoning`, `-multi-agent-0309` and the `xai/`-prefixed form now all resolve at $1.25 / $2.50, where they previously matched nothing and billed $0. `grok-4` and `grok-4-0709` bill at the grok-4.3 rate their retired slugs redirect to.

The added absolute-rate assertions are non-vacuous — a uniform 1.5x drift on either model reds exactly its own parametrized case, nothing else. 238 tests pass.

Two gaps left deliberately outside this PR, worth noting rather than fixing here: `grok-4-fast-reasoning` and `grok-4-fast-non-reasoning` are also retired-and-redirected but still unpriced, and `grok-4.3`, `grok-4.5` and `grok-build-0.1` are currently published models with no catalogue entry at all.

XinweiHe (approved):

PR #1346, opened 2026-06-25
fix(worker): add Moonshot/Kimi model pricing (kimi-k2.6)

Kayvan-Zahiri:
### What

Adds Moonshot pricing for `kimi-k2.6`. #703 added it to the BYOK dropdown (`ADAPTER_MODELS.moonshot` in `frontend/packages/core/src/llm-providers.ts`) for #702, but no price entry came with it. On main today `get_model_price("kimi-k2.6")` returns `None`, so `calculate_cost()` leaves `cost` as `None` and ingest stores no cost for these spans.

Same shape as #1342 (xAI Grok pricing).

### Change

- One entry, `tr-kimi-k2-6`, placed after `tr-kimi-k3`. USD per 1M tokens from Moonshot's [live price page](https://platform.kimi.ai/docs/pricing/chat-k26): $0.95 input, $4.00 output, $0.16 cached input.
- `TestKimiK26ModelIds` covers bare, `moonshot/`-prefixed and version-suffixed ids, pins the three rates, and checks a cost calculation.

### Notes

`kimi-k2.5` and `kimi-k2-thinking` are dropped from this PR because Moonshot has retired both. Both ids are still listed in `ADAPTER_MODELS.moonshot`, so removing them from the dropdown could be a separate change.

### DIFF
--- a/frontend/packages/core/src/standard-model-prices.json
+++ b/frontend/packages/core/src/standard-model-prices.json
@@ -995,5 +995,41 @@
       "cacheRead": 0,
       "cacheWrite": null
     }
+  },
+  {
+    "id": "tr-kimi-k2-6",
+    "modelName": "kimi-k2.6",
+    "matchPattern": "(?i)^(moonshot\\/)?kimi-k2\\.6(-[\\d-]+)?$",
+    "provider": "moonshot",
+    "prices": {
+      "input": 0.00000095,
+      "output": 0.000004,
+      "cacheRead": 0.00000016,
+      "cacheWrite": null
+    }
+  },
+  {
+    "id": "tr-kimi-k2-5",
+    "modelName": "kimi-k2.5",
+    "matchPattern": "(?i)^(moonshot\\/)?kimi-k2\\.5(-[\\d-]+)?$",
+    "provider": "moonshot",
+    "prices": {
+      "input": 0.0000006,
+      "output": 0.000003,
+      "cacheRead": 0.0000001,
+      "cacheWrite": null
+    }
+  },
+  {
+    "id": "tr-kimi-k2-thinking",
+    "modelName": "kimi-k2-thinking",
+    "matchPattern": "(?i)^(moonshot\\/)?kimi-k2-thinking(-[\\d-]+)?$",
+    "provider": "moonshot",
+    "prices": {
+      "input": 0.0000006,
+      "output": 0.0000025,
+      "cacheRead": 0.00000015,
+      "cacheWrite": null
+    }
   }
 ]
--- a/frontend/packages/core/src/standard-model-prices.json
+++ b/frontend/packages/core/src/standard-model-prices.json
@@ -1195,6 +1195,18 @@
     }
   },
   {
+    "id": "tr-kimi-k2-6",
+    "modelName": "kimi-k2.6",
+    "matchPattern": "(?i)^(moonshot\\/)?kimi-k2\\.6(-[\\d-]+)?$",
+    "provider": "moonshot",
+    "prices": {
+      "input": 0.00000095,
+      "output": 0.000004,
+      "cacheRead": 0.00000016,
+      "cacheWrite": null
+    }
+  },
+  {
     "id": "tr-grok-4-20",
     "modelName": "grok-4.20",
     "matchPattern": "(?i)^(xai\\/)?grok-4\\.20(-[a-z0-9-]+)?$",
--- a/tests/worker/tokens/test_pricing.py
+++ b/tests/worker/tokens/test_pricing.py
@@ -162,6 +162,16 @@ DEEPSEEK_MODEL_CASES = [
     ("deepseek-v4-pro-20260424", "deepseek-v4-pro"),
 ]
 
+MOONSHOT_MODEL_CASES = [
+    ("kimi-k2.6", "kimi-k2.6"),
+    ("moonshot/kimi-k2.6", "kimi-k2.6"),
+    ("kimi-k2.6-20260420", "kimi-k2.6"),
+    ("kimi-k2.5", "kimi-k2.5"),
+    ("moonshot/kimi-k2.5", "kimi-k2.5"),
+    ("kimi-k2-thinking", "kimi-k2-thinking"),
+    ("moonshot/kimi-k2-thinking", "kimi-k2-thinking"),
+]
+
 
 @patch("worker.tokens.pricing._load_cache", _mock_load_cache)
 class TestGetModelPrice:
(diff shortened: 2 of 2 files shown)


XinweiHe:
Thanks!

1. 
> Sourced from official Moonshot/Kimi list pricing.

can you provide a url link for reference?

2. can you provide some e2e working example where the pricing is successfully updated in the trace details in UI? Thanks!

Kayvan-Zahiri:
Thanks @XinweiHe! Both below.

**1. Pricing reference**

All three are Moonshot's official **direct-API** list prices (provider=`moonshot`), not OpenRouter's routed rates (which differ):

| model | input / 1M | output / 1M | cache-hit | source |
|---|---|---|---|---|
| kimi-k2.6 | $0.95 | $4.00 | $0.16 | https://platform.kimi.ai/docs/pricing/chat-k26 |
| kimi-k2.5 | $0.60 | $3.00 | $0.10 | https://platform.kimi.ai/docs/pricing/chat-k25 |
| kimi-k2-thinking | $0.60 | $2.50 | $0.15 | legacy K2 rate¹ |

¹ `kimi-k2-thinking` was retired by Moonshot on 2026-05-25 and now appears only under [Deprecated models](https://platform.kimi.ai/docs/models) with no live price page — the values are its last-published K2-family rates. It still ships in `ADAPTER_MODELS.moonshot`, so pricing it keeps historical traces costed. Happy to drop it if you'd rather not carry a retired id.

**2. e2e — cost in the trace-detail UI**

Ran the stack locally on this branch (`make prod-lite`) and ingested a `kimi-k2.6` LLM span (200K in / 50K out). The trace detail shows the computed cost:

`200,000 × $0.95/1M + 50,000 × $4.00/1M = $0.390000` — matches the UI exactly.

Before this PR, `get_model_price("kimi-k2.6")` returns `None` (no entry) so the Cost column is blank; with the entry it resolves to $0.39.

Reproduce — no Moonshot key needed (sets `gen_ai.usage.*` explicitly; identical cost path to a real call):

```python
with tracer.start_as_current_span("chat kimi-k2.6") as span:
    span.set_attribute("gen_ai.request.model", "kimi-k2.6")
    span.set_attribute("gen_ai.usage.input_tokens", 200000)
    span.set_attribute("gen_ai.usage.output_tokens", 50000)
# exported via OTLP/HTTP -> POST /api/v1/public/traces
```

Happy to add k2.5 / k2-thinking traces too if useful.
screencapture-localhost-3000-projects-35372e51-a785-4940-a422-5f9eeb2e52da-traces-2026-06-28-16_36_14 (1)
screencapture-localhost-3000-projects-35372e51-a785-4940-a422-5f9eeb2e52da-traces-2026-06-28-16_38_18

frontend/packages/core/src/standard-model-prices.json
```diff
@@ -1095,5 +1095,41 @@
       "cacheRead": 0.0000003,
       "cacheWrite": null
     }
+  },
+  {
+    "id": "tr-kimi-k2-6",
+    "modelName": "kimi-k2.6",
```
dark-sorceror, line 1101: Rates are right but nothing pins them, so a wrong base price passes as long as the ratios hold. Add an absolute-rate assertion per model — $0.95/$4.00/$0.16 for k2.6, $0.60/$3.00/$0.10 for k2.5, $0.60/$2.50/$0.15 for k2-thinking.

dark-sorceror (changes requested):
All nine rates verified against the vendor's own pages. No currency problem — the international platform quotes USD and this repo points at `api.moonshot.ai`, so the CNY exposure on the `.cn` platform does not apply. One thing below.

Two notes that are not blockers. `kimi-k2.5` retired 2026-08-31 and `kimi-k2-thinking` on 2026-05-25, so this is about pricing historical spans correctly rather than forward use — worth saying so in the PR body. And `kimi-k2-thinking` could not be confirmed from the vendor since its page is gone; the rate matches OpenRouter's live catalogue exactly on all three values, but that is a second-hand source.

Kayvan-Zahiri:
Added the absolute-rate assertions, one parametrized case per model at $0.95/$4.00/$0.16,
$0.60/$3.00/$0.10 and $0.60/$2.50/$0.15. I checked it fails when the k2.6 input rate is set 10x.

Agreed on both notes and updated the description. On the k2-thinking source: Moonshot's own
K2 pricing page is in the Wayback Machine from the day before retirement and shows
$0.60/$2.50/$0.15, so it isn't only OpenRouter. Linked in the description.

frontend/packages/core/src/standard-model-prices.json
dark-sorceror: Fixed. Mutation-checked: drifting each of k2.6, k2.5 and k2-thinking uniformly by 1.5x reds exactly that model's parametrized case and nothing else, so the rates are genuinely pinned rather than only resolved.

dark-sorceror (approved):
Fixed and verified independently.

All nine rates check out against the vendor pages, and the new assertions are non-vacuous — a uniform 1.5x drift on k2.6, k2.5 or k2-thinking reds exactly that model's parametrized case and nothing else. 233 tests pass.

Two things recorded rather than blocking. `kimi-k2.5` retired 2026-08-31 and `kimi-k2-thinking` on 2026-05-25, so the value here is pricing already-recorded spans correctly rather than forward usage. And `kimi-k2-thinking` could not be confirmed from Moonshot directly — its page is gone — so that rate rests on an exact three-way match with OpenRouter's live catalogue, which is second-hand.

XinweiHe (approved):

XinweiHe:
I think  kimi-k2.5 was retired 2026-08-31 and now returns 404, and kimi-k2-thinking was retired 2026-05-25? Let's remove them before merging the PR?

dark-sorceror:
Both retirement dates check out against Moonshot directly: `kimi-k2.5` retired 2026-08-31 and `kimi-k2-thinking` on 2026-05-25, both in the retirement table on the models page, and neither on the live price page. `kimi-k2.6` is current and your three rates match the vendor exactly.

On the historical-spans rationale in the PR body, it does not hold, and I had believed it too. These entries have never shipped, so `get_model_price` returns `None` for both ids in production today and no recorded span was ever priced by them. At ingest, `cost` is only written when a price resolved, so those spans have `cost` unset permanently and it is never revisited. Merging the entries would populate `cost_details` against a NULL `cost`, which renders no chip and over-counts the trace popup.

The conflict is worth care. In `standard-model-prices.json` the boundary cuts through the middle of one object, so keeping both sides yields two `cacheRead` keys and a missing `},{`. Rebuild the tail as complete objects: `tr-kimi-k2-6`, then main's `tr-grok-4-20`, then main's `tr-grok-4`. In `test_pricing.py` both sides appended a provider test class at the same anchor and git interleaved them line by line, so take main's `TestXAIModelIds` verbatim and keep yours as a separate class.

Kayvan-Zahiri:
@XinweiHe Removed both. Rebuilt as one commit on current main with only `tr-kimi-k2-6`, ahead of main's `tr-grok-4-20` and `tr-grok-4`. Main's test classes are unchanged and the k2.6 tests sit in `TestKimiK26ModelIds`. `test_pricing.py` passes 289 (284 on main). A 1.5x rate drift fails only the rate test, and deleting the entry fails all five.

@dark-sorceror, you're right about historical spans. Ingest writes `cost` only when a price resolves, so spans recorded without a price keep `cost` unset.

PR #1350, opened 2026-06-26
perf(clickhouse): rebuild spans with trace_id-first sort key + no-IO projection

### LINKED ISSUE #1045: perf(clickhouse): rebuild spans with trace_id-first ordering + no-IO projection
A single-trace lookup currently scans the whole project partition because `trace_id` is not in the spans sort key, and `FINAL` forces all-column reads to merge versions (measured: 1.59 GiB / 21,686 rows / 2.8 s for one 67-span trace). Put `trace_id` early in the sort key so trace detail is an index seek, and add a projection that excludes I/O for time-ranged list queries.

**Tasks**
- New migration under `backend/db/clickhouse/migrations/` (next sequence after `003_*`).
- Rebuild `spans` with `ORDER BY (project_id, trace_id, span_start_time, span_id)`.
- Add projection `spans_no_io_by_start_time` = all columns except `input`/`output`/`metadata`, ordered `(project_id, span_start_time, trace_id, span_id)`.
- **Keep `ReplacingMergeTree(ch_update_time)`** — our live ingest upserts spans (a span streams in incomplete then updates `end_time`/`output`), so we cannot drop to plain append-only. With `trace_id` first in the key, a trace's rows are contiguous, so `FINAL` merges only within that range and stays cheap. Do **not** introduce an `argMax` dedup path.
- Migrate via new-table → `INSERT … SELECT` → `RENAME`. Decide backfill window (e.g. last N days) and handle the live-system rename.

**Acceptance**
- `EXPLAIN` shows an index seek (not full scan) for `WHERE project_id=? AND trace_id=?`.
- Bytes read for the known trace drop by orders of magnitude vs ~1.59 GiB. Validated on staging first.

**Files:** new `backend/db/clickhouse/migrations/00X_*.sql` (ref `002_create_spans.sql`)
**Wave:** 1 — do first in the P2 chain (P2-B reads it).

Part of #1040.

XinweiHe:
## What

Rebuild the ClickHouse `spans` table so single-trace lookups are an index seek instead of a full-partition scan.

The current sort key `(project_id, span_kind, toDate(span_start_time), span_id)` omits `trace_id`, so every trace-detail and detector query (`WHERE project_id = ? AND trace_id = ?`) scans the whole month partition for the project and reads all columns through `ReplacingMergeTree` `FINAL`. That is the root cause of slow large-trace loads and out-of-memory pressure on the read API.

## How

- New sort key `ORDER BY (project_id, trace_id, span_start_time, span_id)` → single-trace reads hit one contiguous index range.
- `PROJECTION spans_no_io_by_start_time` (every column except the `input`/`output`/`metadata` blobs, ordered by time) for list/skeleton queries that never read blobs.
- `SETTINGS deduplicate_merge_projection_mode = 'rebuild'` — required for a projection on a `ReplacingMergeTree` (CH ≥ 24, otherwise error 344, because a naive projection would not reflect rows the engine collapses on merge).
- Standard create-new + backfill + atomic `RENAME` rebuild; `spans_old` is kept as a backstop (dropped manually after verifying parity).
- Explicit column lists (never `SELECT *`) covering the exact live schema (002 base + `environment` from 003 + `usage_details` from 004), so a future `ALTER ADD COLUMN` cannot silently drop or mis-map a column.

## Migration procedure

Run with ingestion paused (scale `traceroot-worker` and `traceroot-detector` to 0) for the backfill + swap window, then resume. Ingestion is async (SDK → S3 → worker → ClickHouse), so a pause only delays ClickHouse writes; queued events flush on resume. There is deliberately **no** concurrent-write catch-up step.

## Validation

- Replayed verbatim against real large-blob traces (24 MiB) in an isolated database: **byte-exact parity** (rows, unique `span_id`s, total blob bytes, `usage_details`), projection materialized, ~5 s on ~105 MiB of blobs.
- Single-trace point lookup read ~9.5× fewer rows on the new sort key (scales with table size).
- Backward compatible: same table name, columns, types, engine, and partitioning; every query filters by `project_id`, which still leads the new key.

Supersedes #1064. Closes #1045.

---
## Summary by cubic
Rebuilds the ClickHouse `spans` table with a `trace_id`-first sort key and a no-IO projection to turn single-trace queries into index seeks and cut reads/memory. Also bumps local `clickhouse-server` default to 25.2 to support the projection setting.

- **Refactors**
  - New ORDER BY: `(project_id, trace_id, span_start_time, span_id)` for contiguous index reads on `project_id` + `trace_id`.
  - Adds projection `spans_no_io_by_start_time` (excludes `input`/`output`/`metadata`) with `deduplicate_merge_projection_mode='rebuild'`.
  - Create-new → backfill → materialize projection → atomic `RENAME`; keep `spans_old` as a backstop and use explicit column lists.

- **Migration**
  - Pause ingestion: scale `traceroot-worker` and `traceroot-detector` to 0.
  - Run on ClickHouse ≥ 24.8; `docker-compose` default is now 25.2. Resume ingestion after the swap.
  - Verify parity, then drop `spans_old` manually.

<sup>Written for commit 6e7df6471f4bacdf49737942745c7d120337ebee. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1350?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/db/clickhouse/migrations/005_rebuild_spans_trace_id_order.sql
+++ b/backend/db/clickhouse/migrations/005_rebuild_spans_trace_id_order.sql
@@ -1,0 +1,185 @@
+-- +goose Up
+
+-- Rebuild `spans` so single-trace lookups are an index seek instead of a
+-- full-partition scan.
+--
+-- The original sort key `(project_id, span_kind, toDate(span_start_time),
+-- span_id)` does NOT contain trace_id, so `WHERE project_id = ? AND
+-- trace_id = ?` (every trace-detail and detector query) must scan the whole
+-- month partition for the project and read all columns through ReplacingMerge-
+-- Tree `FINAL`. With trace_id early in the key the engine reads one contiguous
+-- range, and `FINAL` only merges within that one trace's rows.
+--
+-- ClickHouse cannot change a sort key in place, so this is the standard
+-- create-new + backfill + atomic-rename rebuild, keeping the old table as a
+-- backstop (same pattern other analytics backends use). The column list below
+-- is the EXACT live schema (002 base + `environment` from 003 + `usage_details`
+-- from 004); it is listed explicitly in every INSERT (never `SELECT *`) so a
+-- future ALTER ADD COLUMN can never silently mis-map or drop a column the way a
+-- positional copy would.
+--
+-- Safety: the multi-table RENAME is atomic on a single node, and there is
+-- deliberately NO concurrent-write catch-up step — this migration must run with
+-- ingestion paused (see Step 5). ReplacingMergeTree(ch_update_time) keeps the
+-- newest row per sort key (so a manual reconciliation from the spans_old
+-- backstop, if ever needed, is idempotent); `ch_update_time` is set to insert
+-- time by the ingest path, never backdated.
+
+-- Idempotent cleanup: if a prior run died between RENAME and DROP, these may
+-- still exist. Safe because a successful run leaves neither table behind.
+DROP TABLE IF EXISTS spans_v2;
+DROP TABLE IF EXISTS spans_old;
+
+-- Step 1 — replacement table: trace_id early in the sort key, plus a no-I/O
+-- projection (every column except the input/output/metadata blobs) ordered by
+-- time, for list/skeleton queries that never read blob columns.
+CREATE TABLE spans_v2
+(
+    span_id             String,
+    trace_id            String,
+    parent_span_id      Nullable(String),
+    project_id          String,
+    span_start_time     DateTime64(3),
+    span_end_time       Nullable(DateTime64(3)),
+    name                String,
+    span_kind           String,
+    status              String DEFAULT 'OK',
+    status_message      Nullable(String),
+    model_name          Nullable(String),
+    cost                Nullable(Decimal64(9)),
+    input_tokens        Nullable(Int64),
+    output_tokens       Nullable(Int64),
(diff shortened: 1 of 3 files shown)


XinweiHe:
Thanks @cubic-dev-ai — valid catch on the version dependency. Addressed in 6e7df647.

Detail: `deduplicate_merge_projection_mode` (required for a projection on a `ReplacingMergeTree`) was introduced in ClickHouse ~24.8, so it does fail on 24.3. The deployed staging and production clusters already run **25.2.1**, so the migration is safe there (verified against real data). The exposure was purely the **local/CI default** — `docker-compose.yml` and `docker-compose.prod.yml` defaulted to `24.3`, which would break `make dev` migrations at `CREATE TABLE`.

Fix: bumped the compose default `24.3 → 25.2` to match the deployed version (local dev was already two majors behind prod). Making the migration itself 24.3-compatible isn't viable — projections on `ReplacingMergeTree` aren't supported on 24.3 at all — so aligning the dev version is the correct resolution.

XinweiHe:
@cubic review again.

PR #1368, opened 2026-06-29
fix(api): use exclusive upper bound for end_before in list_users (#1366)

### LINKED ISSUE #1366: fix(api): list_users end_before uses <= instead of < (off-by-one vs. traces/sessions; inconsistent pagination)
### Environment

OS: macOS 14 / Ubuntu 22.04   Python: 3.11   Docker: 26.0   Browser: Chrome 126  

### Steps to Reproduce

1. Start the TraceRoot dev stack so the REST API and ClickHouse are running.
2. Ingest or insert traces so at least one trace has a `trace_start_time` exactly equal to a boundary timestamp, e.g. `2024-01-31T23:59:59.000000`.
3. Call the users endpoint with `end_before` set to that exact boundary:

   ```bash
   curl "http://localhost:8000/api/v1/projects/<project-id>/users?end_before=2024-01-31T23:59:59"
   ```

4. Call the traces endpoint with the same `end_before`:

   ```bash
   curl "http://localhost:8000/api/v1/projects/<project-id>/traces?end_before=2024-01-31T23:59:59"
   ```

5. Call the sessions endpoint with the same `end_before`:

   ```bash
   curl "http://localhost:8000/api/v1/projects/<project-id>/sessions?end_before=2024-01-31T23:59:59"
   ```

6. Observe that the boundary trace appears in the users response but not in the traces or sessions responses.

Alternatively, verify by inspecting the generated SQL. A trace at the boundary timestamp is included by `list_users` but excluded by `list_traces`, `list_sessions`, and `get_session`.

### Expected Behavior

All `end_before` filters in `trace_reader.py` should use the same semantics: an exclusive upper bound (`<`). A trace whose `trace_start_time` equals `end_before` should appear on none of the users/traces/sessions tabs when the same filter is applied.

### Actual Behavior

`list_users` uses `t.trace_start_time <= {end_before:DateTime64(3)}` while `list_traces`, `list_sessions`, and `get_session` all use `t.trace_start_time < {end_before:DateTime64(3)}`. The users tab therefore includes one extra boundary trace per page, causing totals to disagree across tabs that share the same `end_before` filter.

```shell

```

### Screenshots

_No response_

### TraceRoot Version

commit 3f60e26deebab277b2961e6e78dd3a0549d551a4

### Additional Context

- The inconsistent predicate causes cross-tab pagination/count mismatches when operators compare users vs. traces/sessions views using the same date filter.
- Suggested fix:
  ```python
  conditions.append("t.trace_start_time < {end_before:DateTime64(3)}")
  ```
- After the fix, `rg -n "end_before:DateTime64" backend/rest/services/trace_reader.py` should return only `<` predicates.
- Note: the lower-bound `start_after` predicate is already consistent across all four functions (`>=` at lines 88, 457, 655, 842), so only line 846 needs to change.
- Verified against open issues 561–1365 on `traceroot-ai/traceroot` and is not a duplicate.

CedricConday:
## What

`TraceReaderService.list_users` filters `end_before` with `<=`, while `list_traces`, `list_sessions` and `get_session` all use `<`:

```python
# list_users (before)
conditions.append("t.trace_start_time <= {end_before:DateTime64(3)}")
# list_traces / list_sessions / get_session
conditions.append("t.trace_start_time < {end_before:DateTime64(3)}")
```

A trace whose `trace_start_time` exactly equals `end_before` is therefore **included** by the users tab but **excluded** by traces/sessions. With the same `end_before` applied, per-page totals disagree across tabs (one extra boundary trace on the users tab).

## Fix

Align `list_users` with the other readers — an exclusive upper bound (`<`). One-character change.

## Test

Adds `tests/rest/test_list_users_end_before.py`: mocks the ClickHouse client, calls `list_users(end_before=...)`, and asserts the generated SQL uses `t.trace_start_time < {end_before...}` (and not `<=`). Fails on the previous code, passes with this change. Full `tests/rest` suite: **267 passed**.

Closes #1366

---
## Summary by cubic
Make end_before exclusive (<) in list_users to match other readers, so the users tab no longer includes boundary traces and totals stay consistent across tabs. Add a regression test that asserts the generated SQL uses <.

<sup>Written for commit ad4941d2b35beef76240d12275f9ab28816d91c2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1368?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/rest/services/trace_reader.py
+++ b/backend/rest/services/trace_reader.py
@@ -843,7 +843,7 @@ class TraceReaderService:
             params["start_after"] = to_utc_naive(start_after)
 
         if end_before:
-            conditions.append("t.trace_start_time <= {end_before:DateTime64(3)}")
+            conditions.append("t.trace_start_time < {end_before:DateTime64(3)}")
             params["end_before"] = to_utc_naive(end_before)
 
         where_clause = " AND ".join(conditions)
--- a/tests/rest/test_list_users_end_before.py
+++ b/tests/rest/test_list_users_end_before.py
@@ -1,0 +1,46 @@
+"""Regression test for #1366.
+
+`TraceReaderService.list_users` filtered `end_before` with `<=`, while
+`list_traces`, `list_sessions` and `get_session` all use `<`. A trace whose
+`trace_start_time` exactly equals `end_before` therefore appeared on the users
+tab but not the others, so per-page totals disagreed across tabs sharing the
+same filter. All `end_before` filters must use an exclusive upper bound (`<`).
+
+The ClickHouse client is mocked; we assert on the SQL issued, never touching a
+real database.
+"""
+
+from datetime import datetime
+from types import SimpleNamespace
+from unittest.mock import MagicMock, patch
+
+
+def _service_capturing(queries):
+    from rest.services.trace_reader import TraceReaderService
+
+    client = MagicMock()
+
+    def side_effect(query, parameters=None):
+        queries.append(query)
+        return SimpleNamespace(result_rows=[])
+
+    client.query.side_effect = side_effect
+    with patch(
+        "rest.services.trace_reader.get_clickhouse_client",
+        return_value=client,
+    ):
+        return TraceReaderService()
+
+
+def test_list_users_end_before_is_exclusive():
+    queries: list[str] = []
+    service = _service_capturing(queries)
+
+    service.list_users(
+        project_id="proj",
+        end_before=datetime(2024, 1, 31, 23, 59, 59),
+    )
+
+    sql = "\n".join(queries)
+    assert "t.trace_start_time < {end_before:DateTime64(3)}" in sql
+    assert "t.trace_start_time <= {end_before:DateTime64(3)}" not in sql


CedricConday:
Checking back in — this is still current against `main`.

`list_users` uses `<=` for `end_before` (`trace_reader.py:1387`) while `list_traces` (line 483), `list_sessions` (970) and `get_session` (1188) all use `<`. So a trace whose `trace_start_time` lands exactly on `end_before` is still counted by the users tab and excluded everywhere else, which is the per-page total mismatch in #1366.

No conflict, and all checks are green. Anything needed from me?

CedricConday:
Checking in on this one. It has been open since 29 June, and as far as I can tell no human has looked at it yet — the only review on record is the automated one.

It is still current against main: all 18 checks pass, and the change is small at +47/-1 across two files. `list_users` treats `end_before` as an inclusive upper bound, so a user created exactly at the boundary timestamp is returned by both the current page and the next one. This switches it to an exclusive bound and adds a regression test covering the boundary case.

I am happy to rebase it or split it up if either would make it easier to review.

Issue #1374, opened 2026-06-29
feat(ui): trace-filter predicate types, API client, query-key + URL state


ka1kqi:
**Part of #1369** · Depends on: the `?filters=` endpoint (§2.3) · Suggested branch: `feat/trace-filter-fe-state`

Frontend plumbing for the filter IR. (Implementation plan §3.1–§3.3.)

## Scope
- `src/types/api.ts` — add `Predicate = {field, op, value}` and `filters?: Predicate[]`
  to `TraceQueryOptions`.
- `src/lib/api/traces.ts` — serialize `filters` as one URL-encoded JSON param in
  `getTraces`.
- `src/features/traces/hooks/index.ts` — fold `filters` into `tracesQueryKey` with a
  **deterministic (sorted) JSON serialization** (shared with `usePrefetchTraces`;
  non-canonical order fragments the cache and breaks hover-prefetch).
- `src/lib/hooks/use-list-page-state.ts` — add `filters` to the URL-synced hook via the
  existing ref-guarded URL-state pattern; validate on read. Leave legacy
  `use-trace-list-state.ts` alone (reconciliation deferred).

## Acceptance
- Unit test: two filter arrays differing only in order produce the same query key.
- `?filters=` round-trips through the URL and reaches the API client.

PR #1416, opened 2026-07-01
feat(ui): trace-filter predicate IR, API client, field registry

ka1kqi:
Part of #1374

Frontend data layer (PR 3/5; base `feat/trace-filter-api`).

- Predicate IR — framework-free serialization/canonicalization plus builder helpers: chip labels, predicate construction, and `upsertPredicate` (merges a lower + upper bound on one field into a range, replaces same-direction/exact/categorical, and supersedes a **contradictory** opposite bound that would make an empty range).
- API client for the filter endpoints + response types.
- Field registry type and static fallback list (incl. the `integer` hint).

Unit-tested (predicate, predicate-ui, registry, client).

---
## Summary by cubic
Adds a framework-free filter predicate IR, UI helpers, API client endpoints, and a field registry for trace filtering. Filters are sent as one URL‑encoded JSON param with a stable, order‑independent key, and labels/parsing are hardened with inclusive bounds and stricter validation.

- **New Features**
  - Predicate IR: shape guard rejects empty `in` and non‑finite numbers; `canonicalizeFilters` sorts `in` values; URL serialize/parse for `in` and `between` (nullable bounds) with defensive parsing.
  - UI helpers: chip labels use inclusive ≥ / ≤; builders; `upsertPredicate` merges lower+upper into ranges, treats equal bounds as exact, replaces same‑slot or full/exact ranges, and supersedes contradictory pairs with the newer bound.
  - API: `getTraces` appends a `filters` JSON param when non‑empty; new `getFilterFields` and `getFilterValues` (path‑encodes `field`, accepts `start_after` and `end_before`) with typed responses.
  - Field registry: types plus a static fallback list (incl. `integer` hints) for instant render before the live fetch.

<sup>Written for commit 2a853f17a1e68cee9bd82ee1bb083109311a149b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1416?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/ui/src/features/filters/predicate-ui.test.ts
+++ b/frontend/ui/src/features/filters/predicate-ui.test.ts
@@ -1,0 +1,131 @@
+import { describe, it, expect } from "vitest";
+import {
+  predicateLabel,
+  buildInPredicate,
+  buildBetweenPredicate,
+  upsertPredicate,
+} from "./predicate-ui";
+
+describe("predicateLabel", () => {
+  it("renders a single-value `in` as an equality", () => {
+    expect(predicateLabel({ field: "model_name", op: "in", value: ["claude-opus-4.8"] })).toBe(
+      "model_name = claude-opus-4.8",
+    );
+  });
+
+  it("renders a multi-value `in` as a list", () => {
+    expect(
+      predicateLabel({ field: "model_name", op: "in", value: ["claude-opus-4.8", "gpt-4"] }),
+    ).toBe("model_name in [claude-opus-4.8, gpt-4]");
+  });
+
+  it("renders an open lower bound as greater-than", () => {
+    expect(predicateLabel({ field: "cost", op: "between", value: [0.5, null] })).toBe("cost > 0.5");
+  });
+
+  it("renders an open upper bound as less-than", () => {
+    expect(predicateLabel({ field: "cost", op: "between", value: [null, 10] })).toBe("cost < 10");
+  });
+
+  it("renders a closed range as between", () => {
+    expect(predicateLabel({ field: "cost", op: "between", value: [0.5, 10] })).toBe(
+      "cost between 0.5 and 10",
+    );
+  });
+
+  it("renders equal bounds (from `equals`) as `=`", () => {
+    expect(predicateLabel({ field: "cost", op: "between", value: [5, 5] })).toBe("cost = 5");
+  });
+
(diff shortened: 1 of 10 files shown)


ka1kqi:
@cubic-dev-ai refreshed with the latest stack-wide fixes. Please re-review.

ka1kqi:
@cubic-dev-ai addressed the non-finite finding: `isValidPredicate` now requires `Number.isFinite` for between bounds, so a hand-edited `1e999` (→ Infinity) is dropped rather than silently serialized to null / corrupting the cache key. Please re-review.

ka1kqi:
@cubic-dev-ai force-pushed (stack-wide rebase) with predicate-IR hardening. Please re-review:
- `isValidPredicate` now rejects an empty `in` list (was vacuously valid, would 422 the whole list).
- `canonicalizeFilters` sorts `in` values so order-only differences fold to one React Query key.
- `getFilterValues` API client accepts and appends `end_before`.

ka1kqi:
@cubic-dev-ai force-pushed to address your `serializeFiltersParam` finding. Please re-review the latest commit:
- `serializeFiltersParam` now filters predicates through `isValidPredicate` before serializing (returns null if none remain), so an invalid shape (e.g. an empty `in`) is dropped on the way *out* — symmetric with `parseFiltersParam` on the way in.
- Added tests that assert the **raw** serialized output (not laundered back through parse), so a serialize-side regression actually fails them.

ka1kqi:
@cubic-dev-ai force-pushed (operator honesty). Please re-review:
- Chip labels for open bounds now render `≥` / `≤` (were `>` / `<`) to match the inclusive semantics.
- `boundsFormRange` now accepts `lo <= hi` (was strict `lo < hi`): with the upper bound inclusive, `≥ x` AND `≤ x` is a valid one-value range, so equal opposing bounds now correctly coexist instead of superseding. Test + comments updated.

XinweiHe (approved):

PR #1480, opened 2026-07-08
fix(worker): gate api token counts to llm spans

### LINKED ISSUE #1206: Gate all API token-count adoption to LLM-kind spans to prevent agent-wrapper double-counting
> ⚠️ **Do NOT implement before completing the verification checklist below.** This touches a *deliberately* ungated code path and can silently drop legitimate token counts if rushed. The point of this issue is the investigation; the code change is small and comes last.

## Background

PR #1187 fixed a double-count where the Vercel AI SDK's `ai.generateText` / `ai.generateObject` **AGENT wrapper** span was priced on top of its LLM `doGenerate` children. The OpenInference Vercel processor stamps *both* a model name **and** aggregate `ai.usage.*` token totals on the wrapper, so the transform priced the same tokens twice. #1187 gates the `ai.usage.*` **gross totals** to LLM-kind spans.

## The general problem

The API-count branch in `backend/worker/otel_transform.py` reads the normalized token keys (`llm.token_count.*`, `gen_ai.usage.*`) **ungated by span kind** — this is intentional (see the comment near the model/token extraction: *"extract … not just for LLM spans. Auto-instrumentors set model/token attrs on AGENT and CHAIN spans too"*). So #1187 only closed the `ai.usage.*` path; the same wrapper-double-count could recur via the normalized keys for any instrumentor that rolls child usage onto a non-LLM wrapper span.

A live cross-instrumentor sweep (Vercel TS, OpenAI Agents Py + TS, CrewAI Py, LangGraph/LangChain Py, pydantic-ai Py) found:

| Instrumentor | Wrapper carries usage totals? | Wrapper carries model name? | Double-counts today? |
|---|---|---|---|
| Vercel AI SDK (TS) | yes (`ai.usage.*`) | **yes** | was yes → fixed by #1187 |
| OpenAI Agents (Py / TS) | no | — | no |
| CrewAI (Py) | no | — | no |
| LangGraph / LangChain (Py) | no | — | no |
| pydantic-ai (Py) | yes (`gen_ai.usage.*`) | **no** | no — but only *incidentally* |

**pydantic-ai is the warning sign:** its `invoke_agent` wrapper carries aggregate `gen_ai.usage.*` equal to the sum of its `chat` children, and is safe **only** because the wrapper omits a model name (the transform's `if model_name:` guard skips it). If a future version adds a model name to that span, it would immediately start double-counting via the *normalized* keys — which #1187 does not gate.

## Proposed change (small)

Gate **all** API token-count adoption to LLM-kind spans, mirroring the LLM-only text-estimation gate and the `ai.usage` gate from #1187. This closes the wrapper-double-count *class* permanently instead of patching it per instrumentor.

Expected to be low-risk: a span only reaches token extraction `if model_name:`, and `get_span_kind()` already classifies any model-bearing span as LLM via its fallback **unless** it is explicitly tagged `openinference.span.kind=AGENT/CHAIN` or `traceroot.span.type=…`. So the only behavior change is for explicitly-tagged wrapper spans that also carry a model name — exactly the spans that should never be priced.

## ⚠️ Verification required BEFORE any implementation

The ungated behavior was deliberate, so the single regression vector is **an instrumentor that puts the SOLE usage totals on an explicitly-non-LLM span with no LLM child** — gating would drop those counts. Do not write the gate until all of these are done:

- [ ] Confirm no supported instrumentor emits sole usage on an explicitly-non-LLM (AGENT/CHAIN/SPAN/TOOL) span with no LLM child.
- [ ] Trace *why* the ungated path was originally added (git history + the "auto-instrumentors set token attrs on AGENT/CHAIN spans too" rationale) — make sure no real use case depended on pricing non-LLM spans.
- [ ] Replay real captured span shapes from each supported source — Vercel, OpenAI Agents (Py/TS), CrewAI, LangGraph/LangChain, pydantic-ai, claude-agent-sdk, and raw OTel GenAI / OpenInference — through the transform and assert **no token counts are lost** vs current behavior.
- [ ] Full `tests/worker/` suite stays green.
- [ ] Add regression tests for the gate, including the edge case "explicitly-non-LLM span with a model name + usage and no LLM child."

## Notes

- If implemented, the `ai.usage.*` **gross-total** gate from #1187 becomes redundant and can fold into the general rule. The `ai.usage.*` **cache/reasoning** keys stay — those are the actual value of #1187 (the only place that detail is exposed).
- Scope: `backend/worker/otel_transform.py` token extraction only; no schema/migration changes expected.
- Motivated by the cross-instrumentor regression sweep done while reviewing #1187.

fengjikui:
Closes #1206.

## Summary

- gate API-provided token count adoption to spans classified as LLM
- keep `model_name` on explicitly non-LLM spans, but do not price their normalized `llm.token_count.*` / `gen_ai.usage.*` values
- add regressions for normalized wrapper usage both with and without an LLM child
- update the old estimation-gate tests that intentionally protected the previous non-LLM API-count behavior

## Why

The Vercel raw `ai.usage.*` path was already gated to LLM spans, but normalized keys were still read whenever a span had a model name. If an AGENT/CHAIN wrapper carries model metadata plus aggregate child usage, the wrapper gets priced on top of the real LLM child spans.

This keeps sparse emitters working because a model-bearing span with no explicit non-LLM kind is still classified as LLM by `get_span_kind()`.

## Validation

- `uv run pytest tests/worker/test_otel_transform.py -k "normalized_api_counts_dropped_on_explicit_non_llm_span or normalized_non_llm_wrapper_does_not_double_count_llm_child"` — 2 passed
- `uv run pytest tests/worker/test_otel_transform.py` — 104 passed
- `uv run pytest tests/worker/tokens/test_buckets.py tests/worker/tokens/test_pricing.py` — 184 passed
- `uv run pytest tests/worker/test_otel_transform_estimation_gate.py` — 13 passed
- `uv run pytest tests/worker/` — 343 passed
- `uv run ruff check backend/worker/otel_transform.py tests/worker/test_otel_transform.py tests/worker/test_otel_transform_estimation_gate.py` — passed
- `git diff --check` — clean

---
## Summary by cubic
Gate API-provided token counts to LLM spans to prevent double-pricing when wrapper spans report aggregated usage. Non-LLM spans keep `model_name` but no longer adopt normalized counts or accrue cost. Aligns with #1206.

- **Bug Fixes**
  - Read normalized API counts (`llm.token_count.*`, `gen_ai.usage.*`) only on LLM spans; AGENT/CHAIN wrappers with models no longer get tokens or cost from these fields. Vercel `ai.usage.*` gross totals remain an LLM-only fallback.
  - Preserve `model_name` on non-LLM spans; spans with a model and no explicit non-LLM kind still classify as LLM.
  - Added regressions to ensure wrapper totals don't double-count LLM children and updated estimation-gate tests to reflect the new boundary.

<sup>Written for commit f974c171ec07cc1f382f3c98c5d32d533b2755e1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1480?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/worker/otel_transform.py
+++ b/backend/worker/otel_transform.py
@@ -436,10 +436,12 @@ def transform_otel_to_clickhouse(
                         json.dumps(span_output) if not isinstance(span_output, str) else span_output
                     )
 
-                # Model & token fields — extract API-provided counts whenever a model
-                # name is present, not just for LLM spans. Auto-instrumentors
-                # (OpenInference, GenAI) set model/token attrs on AGENT and CHAIN spans
-                # too. Text-based ESTIMATION, however, is LLM-spans-only (see below).
+                # Model & token fields — keep model names on every span that
+                # reports them, but adopt token counts only from LLM-kind spans.
+                # Some auto-instrumentors roll child usage up onto AGENT/CHAIN
+                # wrappers; pricing those wrapper totals double-counts the
+                # underlying LLM calls. Text-based ESTIMATION is also
+                # LLM-spans-only (see below).
                 model_name = (
                     span_attrs.get("traceroot.llm.model")
                     or span_attrs.get("gen_ai.request.model")
@@ -450,17 +452,19 @@ def transform_otel_to_clickhouse(
 
                     # Try API-provided token counts first (from instrumentors).
                     # OpenInference: llm.token_count.*  ·  GenAI semconv: gen_ai.usage.*
-                    input_token_keys = [
-                        "llm.token_count.prompt",
-                        "gen_ai.usage.input_tokens",
-                        "gen_ai.usage.prompt_tokens",
-                    ]
-                    output_token_keys = [
-                        "llm.token_count.completion",
-                        "gen_ai.usage.output_tokens",
-                        "gen_ai.usage.completion_tokens",
-                    ]
+                    input_token_keys: list[str] = []
+                    output_token_keys: list[str] = []
                     if span_kind == SpanKind.LLM:
+                        input_token_keys = [
+                            "llm.token_count.prompt",
+                            "gen_ai.usage.input_tokens",
+                            "gen_ai.usage.prompt_tokens",
+                        ]
+                        output_token_keys = [
+                            "llm.token_count.completion",
+                            "gen_ai.usage.output_tokens",
+                            "gen_ai.usage.completion_tokens",
+                        ]
                         # Vercel AI SDK raw GROSS totals are the only token source it
                         # normalizes to neither llm.* nor gen_ai.*, so we read them as
                         # a fallback. But they sit on BOTH the LLM doGenerate span AND
--- a/tests/worker/test_otel_transform.py
+++ b/tests/worker/test_otel_transform.py
@@ -734,6 +734,85 @@ class TestTransformOtelToClickhouse:
             assert (s.get("output_tokens") or 0) == 0, f"oi_kind={oi_kind}"
             assert (s.get("cost") or 0) == 0, f"oi_kind={oi_kind}"
 
+    def test_normalized_api_counts_dropped_on_explicit_non_llm_span(self):
+        """An explicitly non-LLM wrapper with a model name and normalized usage
+        must not be priced from those counts. Future instrumentors may add model
+        names to AGENT/CHAIN wrappers that already carry aggregate gen_ai/llm
+        usage; accepting those counts would price wrapper totals as real calls."""
+        from unittest.mock import patch
+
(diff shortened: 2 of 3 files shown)


fengjikui:
CLA is now signed and all checks are green. Happy to make any changes if you'd like a different shape for this token-counting fix.

PR #1486, opened 2026-07-08
fix(ui): pressing Enter should confirm the delete workspace/project dialogs

### LINKED ISSUE #1483: feat(ui): pressing Enter should confirm the delete workspace/project dialogs
## Summary

In the settings UI, deleting a **workspace** or a **project** opens a confirmation dialog where you must type the resource's name to enable the delete button. Once the name matches, pressing **Enter** in that input should trigger the delete — today it does nothing, and you have to reach for the mouse and click the button.

## Where the code lives

Both dialogs follow the same pattern, in two files:

| Surface | File |
|---|---|
| Delete workspace dialog | `frontend/ui/src/features/settings/workspace/components/GeneralTab.tsx` |
| Delete project dialog | `frontend/ui/src/features/settings/project/components/GeneralTab.tsx` |

In each file, look for:
- `deleteConfirmText` — React state holding what the user has typed.
- `handleDelete` — fires the delete mutation, but only if `deleteConfirmText` exactly matches the workspace/project name.
- The confirmation `<Input>` inside the `<Dialog>` — it only has `value` and `onChange`, and the dialog body is not a `<form>`, which is why Enter is currently a no-op.
- The destructive `<Button>` in the `DialogFooter` — disabled unless the typed name matches and no delete is already in flight (`deleteMutation.isPending`).

## Expected behavior

Pressing Enter in the confirmation input behaves exactly like clicking the delete button:

1. If the typed text matches the workspace/project name and no delete is in flight → start the delete.
2. If the text doesn't match → nothing happens.
3. If a delete is already pending → nothing happens (no double submit).

## Suggested implementation

This codebase already has the idiom for this — the **Create API Key** dialog in `frontend/ui/src/features/settings/project/components/AccessKeysTab.tsx` does:

```tsx
<Input
  value={newKeyName}
  onChange={(e) => setNewKeyName(e.target.value)}
  onKeyDown={(e) => e.key === "Enter" && handleCreate()}
/>
```

Apply the same pattern to both delete-confirmation inputs. One subtlety: `handleDelete` already guards against a name mismatch (case 2 above), but it does **not** check `deleteMutation.isPending` — only the button's `disabled` prop does, and a keyboard handler bypasses that. So either add the `isPending` check inside `handleDelete`, or guard it in the `onKeyDown` handler. Adding it to `handleDelete` is the cleaner option.

The change is nearly identical in both files.

## Tests

CI enforces ≥ 80% test coverage on changed lines, so tests are part of the change. Neither `GeneralTab.tsx` has a test file yet; add one next to each component (`GeneralTab.test.tsx`). There's a good template in the same directory: `frontend/ui/src/features/settings/project/components/DetectorsTab.test.tsx` shows the conventions — `// @vitest-environment jsdom` pragma on the first line, `vi.mock` for the data hooks (`useWorkspace` / `useProject`) and `@/lib/api`, then drive the component with `render` / `fireEvent` from `@testing-library/react`. You'll also need to mock `next/navigation`'s `useRouter` and wrap the render in a `QueryClientProvider`.

Worthwhile cases per dialog:
- Typing the exact name and pressing Enter calls the delete API.
- Pressing Enter with a non-matching name does **not** call it.
- Pressing Enter while a delete is pending does not fire a second call.

Run locally from `frontend/ui`:

```bash
pnpm install
pnpm test          # or: pnpm test:coverage
```

## Scope

Just the two dialogs above — no changes to the shared `Dialog`/`Input` components, and no behavior change for the mouse path.

YashManek1:
## What Changed
Both the workspace and project "Delete" confirmation dialogs required a mouse click on the destructive button even after the confirmation name matched — pressing Enter did nothing, unlike the Create API Key dialog elsewhere in Settings which already supports Enter-to-submit.

## Root Cause
In both `frontend/ui/src/features/settings/{workspace,project}/components/GeneralTab.tsx`, `handleDelete` only checked `deleteConfirmText === name`; it never checked `deleteMutation.isPending`, and the confirm `<Input>` had no `onKeyDown` handler at all.

## Approach
- Added `&& !deleteMutation.isPending` inside `handleDelete` in both files, so the guard lives in the function itself rather than only on the button's `disabled` prop (which a keyboard path would otherwise bypass).
- Wired `onKeyDown={(e) => e.key === "Enter" && handleDelete()}` on both confirm inputs, mirroring the existing idiom already used in `AccessKeysTab.tsx`.

## How Validated
- Added `GeneralTab.test.tsx` next to each component (6 tests total): exact name + Enter deletes; non-matching name + Enter no-ops; pending delete + Enter does not double-submit.
- `pnpm vitest run` — 6/6 new tests pass; full suite 510/511 pass (1 pre-existing unrelated failure in `usage-labels.test.ts`, confirmed failing on `main` without this change too).
- `diff-cover` on changed lines: 100% (gate is 80%).
- `eslint` / `prettier` clean.
- Manually verified in the running app: typing the exact name and pressing Enter deletes without touching the button; a non-matching name + Enter is a no-op; pending delete + repeated Enter doesn't double-submit. See attached recording.

## For UI Changes Only

https://github.com/user-attachments/assets/047b85dd-89e9-4655-9e4c-672e507df107

## Additional Notes
- Closes #1483
- Pre-commit hook passed (not skipped)
- Branch based on latest `main`

---
## Summary by cubic
Pressing Enter now confirms the delete dialogs for workspaces and projects, matching the API Key dialog behavior. Closes #1483.

- **Bug Fixes**
  - Added an Enter key handler on the confirm inputs in both General tabs; submit when the typed name matches and ignore IME composition.
  - Guarded against double-submits by checking `!deleteMutation.isPending` inside `handleDelete`.
  - Added 6 Vitest tests for correct/incorrect names and pending-state blocking.

<sup>Written for commit 12979ff4ade3ed99bfea80b92c0b25c93ec8fb3e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1486?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/ui/src/features/settings/project/components/GeneralTab.test.tsx
+++ b/frontend/ui/src/features/settings/project/components/GeneralTab.test.tsx
@@ -1,6 +1,8 @@
 // @vitest-environment jsdom
 import { afterEach, describe, expect, it, vi } from "vitest";
-import { render, cleanup, screen, fireEvent, waitFor } from "@testing-library/react";
+import { render, cleanup, screen, fireEvent, waitFor, within } from "@testing-library/react";
+
+type ProjectUpdatePayload = { name: string };
 
 const mocks = vi.hoisted(() => ({
   project: {
@@ -8,8 +10,12 @@ const mocks = vi.hoisted(() => ({
     name: "Acme Project",
     workspace_id: "w1",
   },
-  deleteProject: vi.fn().mockResolvedValue(void 0),
-  updateProject: vi.fn().mockResolvedValue(void 0),
+  deleteProject: vi
+    .fn<(workspaceId: string, projectId: string) => Promise<void>>()
+    .mockResolvedValue(void 0),
+  updateProject: vi
+    .fn<(workspaceId: string, projectId: string, data: ProjectUpdatePayload) => Promise<void>>()
+    .mockResolvedValue(void 0),
 }));
 
 vi.mock("@/features/projects/hooks", () => ({
@@ -17,8 +23,10 @@ vi.mock("@/features/projects/hooks", () => ({
 }));
 
 vi.mock("@/lib/api", () => ({
-  updateProject: (...a: any[]) => mocks.updateProject(...a),
-  deleteProject: (...a: any[]) => mocks.deleteProject(...a),
+  updateProject: (workspaceId: string, projectId: string, data: ProjectUpdatePayload) =>
+    mocks.updateProject(workspaceId, projectId, data),
+  deleteProject: (workspaceId: string, projectId: string) =>
+    mocks.deleteProject(workspaceId, projectId),
 }));
 
 vi.mock("next/navigation", () => ({
@@ -50,9 +58,8 @@ describe("GeneralTab project delete dialog", () => {
     // Open the delete dialog
     fireEvent.click(screen.getByRole("button", { name: /delete project/i }));
 
-    // Get the input from the dialog (the last one with this placeholder)
-    const inputs = screen.getAllByPlaceholderText("Project name");
-    const input = inputs[inputs.length - 1];
+    const dialog = screen.getByRole("dialog");
+    const input = within(dialog).getByPlaceholderText("Project name");
     fireEvent.change(input, { target: { value: "Acme Project" } });
 
     // Press Enter
@@ -70,16 +77,12 @@ describe("GeneralTab project delete dialog", () => {
     // Open the delete dialog
     fireEvent.click(screen.getByRole("button", { name: /delete project/i }));
 
-    // Get the input from the dialog (the last one with this placeholder)
-    const inputs = screen.getAllByPlaceholderText("Project name");
-    const input = inputs[inputs.length - 1];
+    const dialog = screen.getByRole("dialog");
+    const input = within(dialog).getByPlaceholderText("Project name");
     fireEvent.change(input, { target: { value: "Wrong Project" } });
 
     // Press Enter
     fireEvent.keyDown(input, { key: "Enter" });
-
-    // Give it a moment to process
-    await new Promise((r) => setTimeout(r, 50));
 
     // Verify delete was NOT called
     expect(mocks.deleteProject).not.toHaveBeenCalled();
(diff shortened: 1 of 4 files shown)


ka1kqi (approved):

frontend/ui/src/features/settings/project/components/GeneralTab.tsx
```diff
@@ -138,6 +138,7 @@ export function GeneralTab({ projectId }: GeneralTabProps) {
             <Input
               value={deleteConfirmText}
               onChange={(e) => setDeleteConfirmText(e.target.value)}
+              onKeyDown={(e) => e.key === "Enter" && handleDelete()}
```
ka1kqi: **Optional:** consider also guarding against IME composition:

```tsx
onKeyDown={(e) => e.key === "Enter" && !e.nativeEvent.isComposing && handleDelete()}
```

For names typed with an IME (Japanese/Chinese/Korean), the Enter that *commits the composition* also fires a keydown with `key === "Enter"` — and at that moment the text can already match exactly, triggering an accidental irreversible delete. The existing `AccessKeysTab` idiom has the same gap, so this is a codebase-wide nit rather than a defect introduced here — but a destructive dialog is where it stings most.

frontend/ui/src/features/settings/workspace/components/GeneralTab.tsx
```diff
@@ -135,6 +135,7 @@ export function GeneralTab({ workspaceId }: GeneralTabProps) {
             <Input
               value={deleteConfirmText}
               onChange={(e) => setDeleteConfirmText(e.target.value)}
+              onKeyDown={(e) => e.key === "Enter" && handleDelete()}
```
ka1kqi: **Optional:** same `!e.nativeEvent.isComposing` consideration as in the project `GeneralTab` — worth adding in both places if you take it.

frontend/ui/src/features/settings/project/components/GeneralTab.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+describe("GeneralTab project delete dialog", () => {
+  it("pressing Enter with the exact project name triggers delete", async () => {
+    renderTab();
+
+    // Open the delete dialog
+    fireEvent.click(screen.getByRole("button", { name: /delete project/i }));
+
+    // Get the input from the dialog (the last one with this placeholder)
+    const inputs = screen.getAllByPlaceholderText("Project name");
```
ka1kqi: **Optional (applies to both test files):** the `getAllByPlaceholderText(...)` + take-the-last approach works because the rename input shares this placeholder, but it would silently grab the wrong element if a third matching input ever appears. Scoping to the dialog is self-documenting and robust:

```tsx
import { within } from "@testing-library/react";

const input = within(screen.getByRole("dialog")).getByPlaceholderText("Project name");
```

Relatedly (also optional), the fixed 50 ms sleeps for the negative assertions are a mild flake/slowness smell — flushing microtasks or `await waitFor(() => expect(...).not...)` would be tighter. Fine at this scale either way.

ka1kqi (commented):
LGTM — the fix is correct and minimal, the `isPending` guard inside `handleDelete` is the right layering (the button's `disabled` alone wouldn't cover the keyboard path), and the tests exercise the real react-query `isPending` rather than mocking it away.

A few **optional** suggestions inline (none blocking). One more optional nit that has no line to attach to: the title says `feat(ui)` but this is really a `fix(ui)` — the PR body itself files it under Bug Fixes — which matters only for changelog grouping.

YashManek1:
Thanks! I'll address the optional suggestions and push an update shortly.

YashManek1:
Hi, @ka1kqi I have made the required changes

ka1kqi:
> Hi, @ka1kqi I have made the required changes

ill take a look

XinweiHe (approved):

Issue #1460, opened 2026-07-07
[Epic] Project dashboards — composable widget dashboards over traces and spans


ka1kqi:
## Summary

> Sub-issues are labeled **needs approval** and assigned — please don't start work on them until the maintainer clears it.

Add per-project, user-composable dashboards: a named, draggable grid of widgets where each widget is either a declarative query compiled server-side to ClickHouse SQL, or a canned component (recent-trace feed). Ships with a seeded per-project "Overview" dashboard so the page is useful before any customization.

## Goal

A user opens their project's dashboard, sees seeded metrics immediately (requests, cost, errors, p95 latency, cost/tokens by model, recent traces), rearranges the grid, and builds new widgets in a query-first builder with a live preview — without writing SQL and without any way to read another project's data.

## Architecture

Dashboard + widget definitions (spec JSONB, layout JSONB) live in Postgres behind Next.js CRUD routes → the dashboard page loads one JSON (dashboard + widgets + layout) → each card independently executes its spec: `POST /widgets/query` on the FastAPI backend → field registry validates names → compiler emits one parameterized SELECT over a project-scoped, time-bounded, deduped base relation → ClickHouse → `{columns, rows, meta}` → client renderers. Definitions and data never mix: Postgres never sees query results, ClickHouse never sees widget definitions. No workers, no precomputation — every fresh window is fully current.

## UI

- **Dashboard page:** tabs across the project's dashboards (default marked), 24h/7d/30d range presets, drag/resize grid with debounced+deduped layout persistence, per-card loading/error states.
- **Widget builder page** (`…/widgets/new`, `…/widgets/{id}/edit`): two-column — configuration (Data selection: view, filters, metric, breakdown; Visualization: display type, name) beside a large live preview with preview-only range presets; auto-generated widget name that locks on manual edit; styled like the detector creation form.
- **Filters read like the trace-list filter builder:** stored-value dropdowns for string equality, the same operator wording/symbols (`is` / `is not`, `≥ ≤ ≠`), the same `$`/`ms` input adornments, the same field vocabulary (Latency, Cost, Tokens, Errors).

## Approach

- **Backend kernel:** a static field **registry** (two views — `spans`, `traces` — as deduped base relations + per-field ops/aggs/groupability) and a spec→SQL **compiler**. Field names are registry lookups (fail closed, 422 naming the builder step); values bind as parameters; `readonly=1`, 10s cap. Injection is structurally impossible rather than filtered.
[...]

Issue #1449, opened 2026-07-07
feat(api): registry-driven widget spec compiler to parameterized ClickHouse SQL


ka1kqi:
The widget query engine: a static field registry plus a compiler that turns declarative widget specs into parameterized ClickHouse SQL. This is the security and correctness backbone of dashboards.

**Registry** (single source of truth for what can be queried)
- Two views, each defined by a hand-written *base relation*: project-scoped (`{project_id}` param), time-bounded (`{start_time}/{end_time}` params, half-open window), and ReplacingMergeTree-deduped (`ORDER BY ch_update_time DESC LIMIT 1 BY <keys>`), exposing a stable `event_time` alias for bucketing.
  - **spans** — row-level: string dims `name, span_kind, status, model_name, environment`; numeric measures `duration_ms, cost, input_tokens, output_tokens, total_tokens`; `count`.
  - **traces** — traces LEFT JOINed to a per-trace span rollup (`duration_ms` = max end − min start, `error_count` = countIf ERROR, summed cost/tokens). Un-joined traces (empty-string join key sentinel) get NULL measures. Trace measures are window-bounded on the span side too — a deliberate fast-bounded-scan tradeoff that can differ slightly from the trace-detail page at window edges; document it in the source.
- Per-field metadata: SQL expr over base-relation aliases, type, label, allowed filter ops, groupability, allowed aggregations. A schema serializer exposes all of it (minus SQL) to the builder UI.

**Compiler**
- Security model: field/measure/breakdown names never pass into SQL — they are registry lookups that fail closed with a 422 naming the builder step at fault (`{step, message}`); every user value binds as a ClickHouse parameter; `contains` escapes ILIKE wildcards (`%`, `_`, `\`) via the shared helper. Execution is `readonly=1` with a 10s `max_execution_time`.
- Timeseries (line/area): UTC `toStartOfHour` for windows ≤ 2 days, else `toStartOfDay`; the chosen granularity is echoed in response `meta` for axis labeling.
- Breakdown: keep the top-50 groups by the metric, fold the remainder (and NULLs) into `'other'` so high-cardinality dims can't return unbounded rows.
- Histogram: its own query shape — `histogram(20)` over `toFloat64(measure)` (Decimal measures must be cast); breakdown not allowed; `count`/non-numeric measures rejected.
[...]

PR #1510, opened 2026-07-09
feat(api): widget field registry

ka1kqi:
Second PR in the stacked project-dashboards series (epic #1460), stacked on #1508. Backend only, no runtime endpoints yet.

## What's here
- **`widget_registry.py`** — a static per-view field registry for the two queryable views (`spans`, `traces`). Each field declares its type, display label, allowed filter operators, whether it's groupable (breakdown), its aggregations, and whether it's histogrammable. Also holds each view's deduped, time-scoped base SQL and a `registry_schema()` serializer the widget builder consumes.
- The registry is the **single source of truth** for what a widget spec may reference — the spec compiler (next PR) resolves every user-supplied field name against it before any SQL is built, so raw client input never reaches a query.
- **Tests**: `test_widget_registry.py` (invariant tests — every field expr is reachable in its base SQL, required query params present, etc.) and `test_widget_filter_registry_parity.py` (drift test asserting the shared metric expressions stay identical to the pre-existing trace-list filter registry).

## Notes for review
- Pure declarations + SQL strings; the base SQL is fully ClickHouse-parameterized (no user-input interpolation).
- The pydantic spec schemas and the spec→SQL compiler intentionally ship in the next PR, alongside the tests that exercise them (keeps each PR's diff-coverage self-contained).

Part of #1449 (the query engine lands across this PR and the compiler PR).

---
## Summary by cubic
Adds a static widget field registry for `spans` and `traces` to define allowed fields and lock down dashboard queries. Establishes a single source of truth and provides a JSON schema for the builder.

- **New Features**
  - Added `backend/rest/services/widget_registry.py`: per-view registry with base SQL and field metadata (type, label, filter ops, groupable, aggs, histogrammable).
  - Exposes `registry_schema()` for a JSON-friendly schema; internal SQL is not surfaced.
  - Ensures safety: ClickHouse SQL is fully parameterized; the upcoming compiler resolves fields via this registry.
  - Tests cover invariants (required params, `event_time`, count field, histogrammable) and parity with the existing trace filter registry to prevent expression drift.

<sup>Written for commit cfcc9cb7e33898521174fab968675bfbed52f652. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1510?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/rest/services/widget_registry.py
+++ b/backend/rest/services/widget_registry.py
@@ -1,0 +1,211 @@
+"""Static field registry for the widget query engine.
+
+Declares which views and fields widget specs may reference. The compiler
+(`widget_query.py`) refuses anything not declared here, so this file is the
+single source of truth for what dashboard widgets can query. Field `expr`s
+reference aliases produced by each view's base relation (see `base_sql`),
+never raw user input.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import Literal
+
+FILTER_OPS_STRING = ("=", "!=", "contains")
+FILTER_OPS_NUMBER = (">", ">=", "<", "<=", "=", "!=")
+AGGS_NUMBER = ("sum", "avg", "min", "max", "p50", "p95", "p99")
+
+
+@dataclass(frozen=True)
+class FieldDef:
+    expr: str  # SQL over the view's base relation aliases
+    type: Literal["string", "number"]
+    label: str
+    # filterOps: camelCase deliberate — this dict is the JSON contract consumed
+    # directly by the frontend builder UI.
+    filter_ops: tuple[str, ...] = ()
+    groupable: bool = False
+    aggs: tuple[str, ...] = ()
+
+
+@dataclass(frozen=True)
+class ViewDef:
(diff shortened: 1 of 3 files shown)


backend/rest/services/widget_registry.py
```diff
@@ -0,0 +1,211 @@
...
+never raw user input.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import Literal
+
+FILTER_OPS_STRING = ("=", "!=", "contains")
+FILTER_OPS_NUMBER = (">", ">=", "<", "<=", "=", "!=")
```
XinweiHe, line 16: it would be better to re-use the ops string and number with the previous trace listing section filter right? v.s. definitely it here again.

XinweiHe: is this covered by newer stacked PR maybe?

Issue #1457, opened 2026-07-07
feat(api): gap-filled time series, uniq aggregation, and unit metadata for widgets


ka1kqi:
## Summary

Three small, independent query-engine improvements identified during design review. Each is independently shippable; none changes the spec wire format.

## Tasks

- [ ] **Gap-filled time series:** timeseries queries should emit `ORDER BY bucket WITH FILL … STEP INTERVAL 1 HOUR|DAY` (matching the chosen granularity) so buckets with no data render as zeros instead of holes/interpolated gaps in line and area charts.
- [ ] **`uniq` aggregation:** count-distinct as a first-class aggregation — `uniq({expr})` in the compiler's agg map plus registry whitelist entries on the string dims where it's meaningful (unique models, unique users, unique sessions). Unlocks widget types that don't exist today.
- [ ] **Unit metadata in the registry:** add a `unit` field to the registry's field definitions (e.g. `usd`, `ms`), serialize it through `GET /widgets/schema`, and derive the builder's `$`/`ms` input adornments from it — removing the frontend's hardcoded field→unit map so the schema stays the single source of truth.

## Done when

Each item lands with compiler/registry tests (WITH FILL step matches granularity; uniq whitelisting; schema serialization of `unit`) and, for the unit item, the frontend map is deleted in favor of schema data.

- [ ] **Predicate/column pushdown for the traces view:** the traces-view base relation is a static SQL string, so spec filters apply only *after* the per-trace span rollup has been computed for every trace in the window, and the rollup join runs even for widgets that reference no span-derived measure. Borrow the trace-list query's shape: (a) inject trace-level filter conditions into the traces subquery and bound the span-rollup scan with `trace_id IN (SELECT trace_id FROM t)`, so rollups are never computed for traces the outer `WHERE` discards (e.g. an environment filter); (b) skip the span-rollup `LEFT JOIN` entirely when the spec uses no rollup measure or filter (e.g. a plain trace-count widget). Requires the compiler to assemble the traces base relation per-query instead of interpolating a fixed `base_sql`. Impact scales with project size × window length.

PR #1511, opened 2026-07-09
feat(api): widget spec schemas, SQL compiler, and query endpoints

### LINKED ISSUE #1449: feat(api): registry-driven widget spec compiler to parameterized ClickHouse SQL

### LINKED ISSUE #1450: feat(api): widget schema, query, and field-values endpoints
## Summary

Three project-scoped FastAPI endpoints under `/projects/{id}/widgets` — the entire backend surface dashboards need beyond CRUD.

## Endpoints

| Endpoint | Purpose |
|---|---|
| `GET /schema` | Serialize the field registry (fields, types, labels, filter ops, aggs, groupability — no SQL exprs) for the builder UI. Single source of truth: the frontend holds no second field list. |
| `POST /query` | Execute `{spec, start_time, end_time}`. Stateless — saved widgets and builder previews share it. Spec errors → 422 `{step, message}` naming the builder step at fault; execution errors → 500 with a generic detail (no SQL leaked). |
| `GET /field-values/{view}/{field}` | Distinct stored values for a string dimension, for the builder's value dropdowns. Frequency-ordered, capped at 100, bounded to the requested window (never all-time: a default lookback applies when unbounded). |

## Approach

- `field-values` reuses the trace-filter distinct-values scan rather than duplicating it: generalize the service over the source table (spans and traces siblings with per-table dedup keys and time columns), sharing one bounded in-process cache (30s TTL, minute-quantized window keys so per-render jitter can't bypass it, size-capped).
- View and field resolve through the widget registry before any SQL — unknown view/field → 404, numeric or `count` fields → 400. A raw client-supplied column name can never reach a query.
- Project id comes from the URL path after `ProjectAccess`; request bodies never carry it.

## Done when

Endpoint tests cover: spans/traces dispatch to the right distinct scan, time-window threading, 404 unknown view/field, 400 non-string field, the no-auth case, and that `POST /query` scopes by the path project id (never the body).

ka1kqi:
Third PR in the stacked project-dashboards series (epic #1460), stacked on #1510. This is the query engine: with it, a widget spec can be validated, compiled, and executed end to end — the frontend layers come next.

## What's here
- **Spec schemas** (`schemas/dashboards.py`) — strict pydantic request models (`extra="forbid"`): filters (non-empty values), metric, display, breakdown; display/breakdown combination rules enforced (pie/bar require a breakdown, number/histogram forbid one).
- **Spec→SQL compiler** (`services/widget_query.py`) — resolves every user-supplied field name through the registry (raw input never reaches SQL), then emits fully parameterized ClickHouse: time-bucketed series with `WITH FILL` across the whole window (empty buckets render as zeros), top-50 breakdown fold into `other`, histogram shape, window-derived row caps, timezone-normalized and validated bounds.
- **Endpoints** (`routers/dashboards.py`) — three read-only, `ProjectAccess`-guarded routes: widget query execution (readonly + 10s execution cap via a new ClickHouse client `settings` passthrough), the builder schema (`registry_schema()`), and field values for the builder's value dropdowns.
- **Shared distinct-values scan** — `trace_reader`'s categorical-values query is extracted into a reusable helper so the trace-list filter dropdowns and the widget builder read identical data (behavior-preserving; existing tests unchanged).
- **Tests** — compiler suite (SQL shape, param binding, escaping, guard rails), router suite (auth, status mapping, `WidgetSpecError`→422), and the distinct-values parity/meta tests.

## Notes for review
- SQL injection surface: identifiers come from the registry only; all values bind as typed params; `contains` goes through ILIKE escaping.
- Spec validation is split deliberately: shape/enums in pydantic, field/op semantics in the compiler against the registry.

closes #1449
closes #1450
Also checks the gap-filled time-series box on #1457 (its remaining tasks stay open).

---
## Summary by cubic
Adds the dashboards query engine: strict widget spec schemas, a registry-backed SQL compiler, and read-only endpoints to validate, compile, and run widget queries against ClickHouse. Routes now use the shared read rate limit to protect expensive widget reads. Closes #1449 and #1450, and delivers gap-filled time-series for #1457 (remaining tasks pending) within epic #1460.

- **New Features**
  - Strict Pydantic widget spec schemas (`extra="forbid"`) with non-empty filter values and clear display/breakdown rules.
  - Registry-backed compiler that emits fully parameterized SQL: time buckets with WITH FILL, top-50 breakdown folded into "other", 20-bin histograms, window-derived row caps, UTC-normalized bounds, and escaped ILIKE; raw user input never reaches SQL.
  - Endpoints under `/api/v1/projects/{project_id}/widgets` (project-scoped, readonly, 10s timeout):
    - `POST /query` to execute specs,
    - `GET /schema` to expose the builder registry,
    - `GET /field-values/{view}/{field}` for spans/traces value dropdowns (time-bounded, cached; shared distinct-values helper for consistent options).

- **Bug Fixes**
  - Widget `query`, `schema`, and `field-values` now run under the shared read limiter via `RateLimitedProjectAccess`; trusted internal calls remain exempt.

<sup>Written for commit 304fe3b4ff55f4f247cd00ecfa49b842d4bfdba3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1511?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/rest/routers/dashboards.py
+++ b/backend/rest/routers/dashboards.py
@@ -8,9 +8,16 @@ the field registry that drives the builder UI.
 import logging
 from datetime import datetime
 
-from fastapi import APIRouter, HTTPException, Query, status
+from fastapi import APIRouter, HTTPException, Query, Request, Response, status
 
-from rest.routers.deps import ProjectAccess
+from rest.rate_limit import (
+    BUCKET_READ,
+    is_request_rate_limit_exempt,
+    key_read,
+    limiter,
+    resolve_limit,
+)
+from rest.routers.deps import RateLimitedProjectAccess
 from rest.schemas.dashboards import WidgetQueryRequest, WidgetQueryResponse
 from rest.schemas.traces import FilterValuesResponse
 from rest.services.trace_reader import get_trace_reader_service
@@ -23,11 +30,16 @@ router = APIRouter(prefix="/projects/{project_id}/widgets", tags=["Dashboards"])
 
 
 @router.get("/field-values/{view}/{field}", response_model=FilterValuesResponse)
+@limiter.shared_limit(
+    resolve_limit, scope=BUCKET_READ, key_func=key_read, exempt_when=is_request_rate_limit_exempt
+)
 async def get_widget_field_values(
+    request: Request,
+    response: Response,
     project_id: str,
     view: str,
     field: str,
-    _access: ProjectAccess,
+    _access: RateLimitedProjectAccess,  # Validates access + sets rate-limit identity
     start_time: datetime | None = Query(
         None, description="Only consider rows whose event time is at or after this timestamp"
     ),
@@ -45,7 +57,7 @@ async def get_widget_field_values(
         project_id (str): Project that owns the data; server-bound for isolation.
         view (str): The widget view (``spans`` or ``traces``) whose table is scanned.
         field (str): The string dimension to enumerate.
-        _access (ProjectAccess): Validates the caller's access to the project.
+        _access (RateLimitedProjectAccess): Validates access and sets rate-limit identity.
         start_time (datetime | None): Lower bound of the active dashboard window.
         end_time (datetime | None): Upper bound (exclusive) of the active window.
 
@@ -95,16 +107,29 @@ async def get_widget_field_values(
 
 
 @router.get("/schema")
-async def get_widget_schema(project_id: str, _access: ProjectAccess) -> dict:
+@limiter.shared_limit(
+    resolve_limit, scope=BUCKET_READ, key_func=key_read, exempt_when=is_request_rate_limit_exempt
+)
+async def get_widget_schema(
+    request: Request,
+    response: Response,
+    project_id: str,
+    _access: RateLimitedProjectAccess,  # Validates access + sets rate-limit identity
+) -> dict:
     """Field registry for the widget builder (views, fields, ops, aggs)."""
     return registry_schema()
 
 
 @router.post("/query", response_model=WidgetQueryResponse)
+@limiter.shared_limit(
+    resolve_limit, scope=BUCKET_READ, key_func=key_read, exempt_when=is_request_rate_limit_exempt
+)
 async def query_widget_data(
+    request: Request,
+    response: Response,
     project_id: str,
     body: WidgetQueryRequest,
-    _access: ProjectAccess,
+    _access: RateLimitedProjectAccess,  # Validates access + sets rate-limit identity
 ):
     """Execute a widget spec. Stateless: used by saved widgets and builder previews."""
     try:
--- a/backend/rest/routers/dashboards.py
+++ b/backend/rest/routers/dashboards.py
@@ -1,0 +1,127 @@
+"""Widget query endpoints for project dashboards.
+
+Dashboard/widget CRUD lives in Next.js API routes (Prisma/Postgres, same as
+detectors). This router only executes queries against ClickHouse and exposes
+the field registry that drives the builder UI.
+"""
+
+import logging
+from datetime import datetime
+
+from fastapi import APIRouter, HTTPException, Query, status
+
+from rest.routers.deps import ProjectAccess
+from rest.schemas.dashboards import WidgetQueryRequest, WidgetQueryResponse
+from rest.schemas.traces import FilterValuesResponse
+from rest.services.trace_reader import get_trace_reader_service
+from rest.services.widget_query import WidgetSpecError, run_widget_query
+from rest.services.widget_registry import REGISTRY, registry_schema
+
+logger = logging.getLogger(__name__)
+
+router = APIRouter(prefix="/projects/{project_id}/widgets", tags=["Dashboards"])
+
+
+@router.get("/field-values/{view}/{field}", response_model=FilterValuesResponse)
+async def get_widget_field_values(
+    project_id: str,
+    view: str,
+    field: str,
+    _access: ProjectAccess,
+    start_time: datetime | None = Query(
+        None, description="Only consider rows whose event time is at or after this timestamp"
+    ),
+    end_time: datetime | None = Query(
+        None, description="Only consider rows whose event time is before this timestamp"
+    ),
+):
+    """Distinct stored values for a string field, for the builder's value dropdowns.
+
+    Reuses the trace-filter distinct-values scan (deduped, time-bounded, cached).
+    The view and field resolve through the widget registry before any SQL, so a
+    raw client-supplied column name can never reach a query.
+
+    Args:
+        project_id (str): Project that owns the data; server-bound for isolation.
+        view (str): The widget view (``spans`` or ``traces``) whose table is scanned.
+        field (str): The string dimension to enumerate.
+        _access (ProjectAccess): Validates the caller's access to the project.
+        start_time (datetime | None): Lower bound of the active dashboard window.
+        end_time (datetime | None): Upper bound (exclusive) of the active window.
+
+    Returns:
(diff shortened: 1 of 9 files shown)


backend/rest/routers/dashboards.py
```diff
@@ -0,0 +1,127 @@
+"""Widget query endpoints for project dashboards.
```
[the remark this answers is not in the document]
ka1kqi: Good catch — fixed in 304fe3b4. All three widget endpoints (POST /query, GET /field-values, GET /schema) now use `RateLimitedProjectAccess` + `@limiter.shared_limit` on the shared read bucket, mirroring the traces/sessions/users routers; the test fixture stamps a real workspace/plan like the traces one does.

backend/rest/services/widget_query.py
```diff
@@ -0,0 +1,274 @@
...
+                param_value = flt.value
+        else:
+            ch_type = "Float64"
+            try:
+                param_value = float(flt.value)
+            except (ValueError, TypeError):
+                raise WidgetSpecError(
+                    "filters", f"Value for '{flt.field}' must be numeric"
+                ) from None
+        conditions.append(_OP_SQL[flt.op].format(expr=f.expr, p=pname, t=ch_type))
```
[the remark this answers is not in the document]
ka1kqi: Verified empirically against ClickHouse 25.2 (the server version we run): `Decimal64(9)` vs `{p:Float64}` comparisons execute fine — tested `>` and `=` (including 0.1) plus the aggregated `Decimal(38, 9)` shape the traces view produces via `sum(cost)`. Modern ClickHouse converts between the types for comparison, so cost filters don't fail at execution. Binding cost params as `Decimal64(9)` the way the trace-list filter registry does would be a consistency nicety; noting it as an epic follow-up rather than changing the compiler here.

ka1kqi:
@cubic-dev-ai rereview

backend/rest/services/widget_query.py
```diff
@@ -0,0 +1,274 @@
...
+    "min": "min({expr})",
+    "max": "max({expr})",
+    "p50": "quantile(0.5)({expr})",
+    "p95": "quantile(0.95)({expr})",
+    "p99": "quantile(0.99)({expr})",
+}
+
+_OP_SQL = {
+    "=": "{expr} = {{{p}:{t}}}",
+    "!=": "{expr} != {{{p}:{t}}}",
```
XinweiHe, line 36: it would be better to re-use the ops string and number with the previous trace listing section filter right? v.s. definitely it here again.

backend/rest/services/widget_query.py
```diff
@@ -0,0 +1,274 @@
...
+    conditions: list[str] = []
+    for i, flt in enumerate(spec.filters):
+        f = _resolve_field(view.fields, flt.field, "filters")
+        if flt.op not in f.filter_ops:
+            raise WidgetSpecError(
+                "filters",
+                f"Op '{flt.op}' not allowed for '{flt.field}'. Allowed: {list(f.filter_ops)}",
+            )
+        pname = f"f{i}"
+        if f.type == "string":
```
XinweiHe, line 98: if possible, do not directly uses raw string?

backend/rest/services/widget_query.py
```diff
@@ -0,0 +1,274 @@
...
+            ch_type = "String"
+            if flt.op == "contains":
```
XinweiHe, line 100: ditto

backend/rest/services/widget_query.py
```diff
@@ -0,0 +1,274 @@
...
+    gran = _pick_granularity(start_time, end_time)
+    # Fill empty buckets across the whole window so the x-axis spans the
+    # selected range even when stored data starts later: missing buckets come
+    # back as zero rows instead of the chart starting at first data. WITH FILL
+    # TO is exclusive, so the bound is one step past the bucket of the last
+    # in-window instant (end_time - 1ms) — that covers the trailing straddle
+    # bucket of a misaligned window and stays exact for aligned ones.
+    # Bound unconditionally, like gran, so the two is_timeseries branches
+    # below can't drift apart.
+    bucket_fn = "toStartOfHour" if gran == "hour" else "toStartOfDay"
```
XinweiHe, line 175: ditto

backend/rest/services/widget_query.py
```diff
@@ -0,0 +1,274 @@
...
+        else:
+            order_by = "ORDER BY value DESC"
+
+    select_cols.append(f"{metric_sql} AS value")
+    group_by = f"GROUP BY {', '.join(group_cols)}" if group_cols else ""
+
+    # Row cap: for table display use a fixed row limit.
+    # For timeseries or breakdown displays, derive the cap from the actual
+    # query window so that long ranges aren't silently truncated.
+    if spec.display.type == "table":
```
XinweiHe, line 228: ditto

PR #1532, opened 2026-07-11
fix(billing): don't show a bare 0 before the usage snapshot exists

### LINKED ISSUE #1501: Event Usage shows 0/0 on self-host — usage snapshot never computed, and the UI trusts a stale value
## Summary

On self-host, the billing page's **Event Usage** shows **0 traces / 0 spans** even after real traffic is ingested (e.g. an 85-span trace stored correctly in ClickHouse). The page reads a cached Postgres snapshot (`workspace.currentUsage`) that's only written by an hourly metering cron — and that cron runs in a `billing` worker service **not included in the default self-host `docker-compose.yml`**. So the snapshot stays `NULL` and the page renders it as `0`, with no signal that usage is uncomputed rather than genuinely zero.

## How to replicate

1. Run the stack from the base `traceroot/docker-compose.yml` (no `billing` service from `docker-compose.prod.yml`).
2. Ingest a trace (e.g. 85 spans) and confirm it shows in the traces UI.
3. Open `/workspaces/${workspaceId}/settings/billing` → **Event Usage: 0 traces, 0 spans**, despite the trace being present.

## Root cause (confirmed)

1. The page renders `currentUsage?.traces ?? 0` and `currentUsage?.spans ?? 0` (`frontend/ui/src/ee/features/billing/BillingTab.tsx:269,273`).
2. `currentUsage` comes straight from Postgres with no live recompute (`frontend/ui/src/app/api/workspaces/[workspaceId]/route.ts:67`).
3. It's only written by the hourly cron `runBillingJob → processWorkspace` (`frontend/worker/src/ee/billing/usageMetering.ts:52,96`), sourced from `GET /api/v1/internal/usage/details`.
4. That backend endpoint counts correctly — `uniqExact(trace_id)` + `uniqExact(span_id)` by `project_id`+`ch_create_time` (`backend/rest/routers/internal.py:108-190`) → would return traces=1, spans=85.
5. The metering worker lives only in the `billing` service in `docker-compose.prod.yml:273`; the base `docker-compose.yml` has no such service, so the cron never runs and `currentUsage` stays `NULL`.

**Observed:** ClickHouse `default.spans` holds the ingested spans (e.g. 85 spans / 1 trace) for the project, while the Postgres workspace row has `current_usage = NULL`.

**Ruled out:** not detector/source exclusion (count filters only `project_id`+`ch_create_time`); not `ENABLE_BILLING=false` (the metering worker never reads it); not a missing ingest-time counter (counted at read time by design).

## Impact

- Wrong numbers on a money-facing page — Event Usage is permanently 0/0 on self-host, and "no usage" is indistinguishable from "never computed."
- The same cron computes the free-plan `ingestionBlocked` flag, so quota enforcement is also dormant on self-host.

## Acceptance criteria

- With ingested traffic, Event Usage reflects real counts (85 spans → traces=1, spans=85), not 0/0.
- Usage is computed live from the internal endpoint when no snapshot exists, or shown as "not yet computed" — a `NULL` snapshot never renders as a bare `0`.
- The metering job runs on the default self-host deployment, independent of Stripe config.

somtri:
Fixes #1501

## Summary

On self-host, Event Usage showed 0 traces / 0 spans even with real ingested traffic, because the billing page renders the cached `currentUsage` snapshot directly (`?? 0`), and that snapshot is null until the hourly metering job writes it. A freshly booted worker doesn't run the job until the next `:05` tick, so every workspace shows 0/0 for up to an hour after startup, indistinguishable from genuinely having no events.

## Type of Change

- [x] fix - A bug fix

## Details

- `BillingTab.tsx`: when `currentUsage` is null, render "Usage has not been computed yet" instead of falling back to `?? 0`. Once a snapshot exists, rendering is unchanged, including a real all-zero snapshot.
- `usageMetering.ts`: added `runStartupBillingPass`, an injectable wrapper around `runBillingJob` with the same log-and-swallow error handling as the scheduled cron callback. Called once during worker boot (`frontend/worker/src/index.ts`), before the cron is scheduled, so the first snapshot lands within seconds instead of waiting up to an hour.
- The issue's root-cause section says the metering job is missing from the self-host deployment entirely. From what I can tell, both documented self-host paths (`make dev`'s tmux pane running `pnpm dev:billing`, and `make prod`'s `billing` compose service) already run it, and it's already independent of Stripe configuration (`getStripe()` throwing is caught and logged, not fatal). The reported 0/0 looks like it came from running the base `docker-compose.yml` directly, which is infra-only by design and isn't one of the two documented ways to self-host. Happy to be corrected if I'm missing a deployment path. Either way, the startup-run change here fixes the boot window regardless of which deployment path is used.
- Live-computing usage on page load (the other fork the issue mentions for `currentUsage`) isn't done here: the UI's API routes have no outbound path to the backend/ClickHouse (no `BACKEND_INTERNAL_URL` equivalent on that side), so it would need new cross-service plumbing. Flagging it as a possible follow-up rather than folding it into this fix.

None. This is a genuine UI change, but reproducing it visually needs the full self-host stack (Postgres, ClickHouse, Redis, an authenticated workspace) running locally, which I didn't have available. Verified instead with component tests asserting the exact rendered output for both states (see below).

## Checklist

- [ ] I have added screenshots or recordings for UI changes (not done, see above)
- [x] I have updated documentation where necessary (N/A, no doc-facing behavior change)

---
## Summary by cubic
Fixes the misleading “0 traces / 0 spans” state on the billing page before a usage snapshot exists, and makes the worker write the first snapshot within seconds of boot instead of waiting for the next hourly cron tick. Addresses #1501.

- **Bug Fixes**
  - UI: In `BillingTab`, show “Usage has not been computed yet” when `currentUsage` is null; render counts normally once a snapshot exists, including genuine zeros.
  - Worker: Run a startup billing pass fire-and-forget after cron registration, with a 10s timeout on the internal usage fetch so a slow backend can’t hang startup or delay cron scheduling.
  - Concurrency: A shared in-flight guard serializes the startup pass and the scheduled cron to prevent overlapping runs and double-reporting overage to Stripe.
  - Tests: Add coverage for the worker entrypoint (startup sequence, cron ordering, guard branches, SIGTERM shutdown), the usage-fetch timeout, and the unlimited-plan quota display.

<sup>Written for commit 72c781003d184a807aed612e5c13e8c31e049e3f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1532?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/worker/src/index.ts
+++ b/frontend/worker/src/index.ts
@@ -7,7 +7,7 @@
 
 import cron from "node-cron";
 import { prisma, syncStandardPrices } from "@traceroot/core";
-import { runBillingJob, closeClickHouseClient } from "./ee/billing/index.js";
+import { runBillingJob, runStartupBillingPass, closeClickHouseClient } from "./ee/billing/index.js";
 
 // Graceful shutdown handling
 let isShuttingDown = false;
@@ -46,6 +46,10 @@ async function main(): Promise<void> {
 
   // Sync standard model pricing from JSON → DB
   await syncStandardPrices();
+
+  // Run once immediately so the first usage snapshot lands within seconds
+  // of startup instead of waiting for the next cron tick (up to ~1h away).
+  await runStartupBillingPass();
 
   // Schedule billing job (default: every hour at minute 5)
   const billingCron = process.env.USAGE_METERING_CRON || "5 * * * *";
--- a/frontend/worker/src/index.ts
+++ b/frontend/worker/src/index.ts
@@ -12,6 +12,24 @@ import { runBillingJob, runStartupBillingPass, closeClickHouseClient } from "./e
 // Graceful shutdown handling
 let isShuttingDown = false;
 
+// Guards against a cron tick and the startup pass (or two overlapping cron
+// ticks) running runBillingJob() concurrently against the same stale
+// currentUsage, which would double-report overage deltas to Stripe.
+let billingPassInFlight = false;
+
+async function runExclusiveBillingPass(job: () => Promise<void>): Promise<void> {
+  if (billingPassInFlight) {
+    console.log("[Billing Worker] Skipping billing pass: previous pass still running");
+    return;
+  }
+  billingPassInFlight = true;
+  try {
+    await job();
+  } finally {
+    billingPassInFlight = false;
+  }
+}
+
 async function shutdown(signal: string): Promise<void> {
   if (isShuttingDown) return;
   isShuttingDown = true;
@@ -53,17 +71,21 @@ async function main(): Promise<void> {
     if (isShuttingDown) return;
 
     console.log("[Billing Worker] Running scheduled billing job...");
-    try {
-      await runBillingJob();
-    } catch (error) {
-      console.error("[Billing Worker] Billing job failed:", error);
-    }
+    await runExclusiveBillingPass(async () => {
+      try {
+        await runBillingJob();
+      } catch (error) {
+        console.error("[Billing Worker] Billing job failed:", error);
+      }
+    });
   });
 
   // Fire once immediately so the first usage snapshot lands within seconds
   // of startup instead of waiting for the next cron tick (up to ~1h away).
   // Not awaited: a slow/not-yet-ready backend must not delay cron registration.
-  void runStartupBillingPass();
+  // Goes through the same exclusive-pass guard as the cron callback so an
+  // early tick can't run concurrently with it.
+  void runExclusiveBillingPass(runStartupBillingPass);
 
   console.log("[Billing Worker] Scheduled jobs:");
   console.log(`  - Billing: ${billingCron}`);
--- a/frontend/worker/src/index.ts
+++ b/frontend/worker/src/index.ts
@@ -47,10 +47,6 @@ async function main(): Promise<void> {
   // Sync standard model pricing from JSON → DB
   await syncStandardPrices();
 
-  // Run once immediately so the first usage snapshot lands within seconds
-  // of startup instead of waiting for the next cron tick (up to ~1h away).
-  await runStartupBillingPass();
-
   // Schedule billing job (default: every hour at minute 5)
   const billingCron = process.env.USAGE_METERING_CRON || "5 * * * *";
   cron.schedule(billingCron, async () => {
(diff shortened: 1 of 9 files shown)


frontend/worker/src/index.ts
```diff
@@ -47,6 +47,10 @@ async function main(): Promise<void> {
   // Sync standard model pricing from JSON → DB
   await syncStandardPrices();
 
+  // Run once immediately so the first usage snapshot lands within seconds
+  // of startup instead of waiting for the next cron tick (up to ~1h away).
+  await runStartupBillingPass();
```
dark-sorceror: Usage fetch has no `AbortSignal`/timeout. Also register the cron first, then fire the pass with void.

dark-sorceror (changes requested):
Great work! However, this PR doesn't fix the problem where the metering job doesn't run on the default self-host deployment at all.

somtri:
Thanks for the review. Before I add anything, I'd like to nail down the deployment you saw this on, because I can't reproduce "the metering job doesn't run on the default self-host deployment" from the repo as it stands.

Base `docker-compose.yml` on this branch is infra-only: postgres, clickhouse, minio, minio-init, redis, migrate-clickhouse. No `web`, `rest`, `worker`, or `billing` service. On its own it can't even serve the billing page, so it's not where the 0/0 symptom could have come from.

Both documented self-host paths do start the metering worker. `make dev` runs a `pnpm dev:billing` pane with env from the root `.env` (`BACKEND_INTERNAL_URL=http://localhost:8000`, matching the dev REST server). `make prod`/`prod-lite` bring up the `billing` compose service. The worker meters regardless of `ENABLE_BILLING` too; only the quota-notification path checks that flag.

Could you share the exact command you ran and any billing-worker logs? If the actual ask is to make bare `docker-compose.yml` a runnable full-stack target, that's bigger than this PR and worth scoping separately.

Either way, I did find a real first-boot gap: the internal-endpoint fetch had no timeout, and the startup pass was awaited before the cron got registered. Both fixed here per your inline comment.

somtri:
Fixed in 2bfa7634: added a shared in-flight guard (`runExclusiveBillingPass`) around both the cron callback and the startup pass so they can't run concurrently against the same `currentUsage` snapshot.

PR #1533, opened 2026-07-11
fix(ai-assistant): link empty-state and clear phantom model selection

### LINKED ISSUE #1500: AI assistant "No LLM models available" empty state isn't jumpable — and the selector still shows a phantom `claude-opus-4-8`
## Summary

When the project-wide AI assistant has no models available (no system API key *and* no BYOK provider), the empty state tells the user to go to **Workspace Settings → Model Providers** — but it's a plain `<span>`, **not a link**, so there's no way to click through to fix it. The same file already links to that exact destination a few lines above. Separately, the model selector still shows `claude-opus-4-8` as selected even though zero models exist — contradicting the "No LLM models available" message right above it.

## How to replicate

1. Use a workspace with no system model env keys and no BYOK provider (so `GET /api/workspaces/[workspaceId]/llm-models` returns empty).
2. Open a trace → open the AI assistant panel.
3. The empty state reads "No LLM models available … add a BYOK provider in **Workspace Settings → Model Providers**" — but that text isn't clickable.
4. The model selector at the bottom still shows `claude-opus-4-8` selected (input is correctly disabled, but the label is misleading).

## Root cause (confirmed)

**Part 1 — plain-text empty state.** `frontend/ui/src/features/ai-assistant/components/ai-assistant-panel.tsx:172-174` wraps "Workspace Settings → Model Providers" in a `<span>`. The correct pattern is already in the same file at `:150-157` (the unsupported-model warning uses `<a href={`/workspaces/${workspaceId}/settings/model-providers`}>`), and `workspaceId` is in scope at `:48`. The dropdown's own "No models available" text (`model-selector.tsx:149-153`) is likewise plain.

**Part 2 — phantom `claude-opus-4-8`.** While the models query is loading, `flattenAvailableModels(undefined)` returns the hardcoded `FALLBACK_MODELS` built from `SYSTEM_MODELS` (`resolve-model.ts:36-37,25-34`). The reconcile effect then picks a default via `pickDefaultModel` → `PROVIDER_PRIORITY[0] = ANTHROPIC` → `claude-opus-4-8` (`resolve-model.ts:61-72`, `llm-providers.ts:126-127,93`) and writes it to selection state. When the query resolves empty, the effect early-returns at `model-selector.tsx:50` (`if (models.length === 0) return;`) and never clears it, so the button falls back to `value.model` and shows `claude-opus-4-8` (`model-selector.tsx:104`).

Not a functional/security bug: the input is disabled via `disabled={!projectId || !hasModels}` (`ai-assistant-panel.tsx:186`), so nothing can be sent. It's a UI-correctness issue — the panel (`hasModels`, `:56-59`) and the selector (its own `FALLBACK_MODELS`-laden list) compute availability independently, which is the source of the contradiction.

## Impact

- **Dead-end empty state:** the one message whose job is to route the user to the fix doesn't link to it — a first-run/new-workspace paper-cut.
- **Contradictory UI:** "No LLM models available" shown alongside a confidently-selected `claude-opus-4-8` reads as a glitch.

## Related

- **#1273** ("New Detector defaults to Claude when no model providers are configured") — same root-cause family (`SYSTEM_MODELS` fallback + `pickDefaultModel`), but the detector surface; a `resolve-model.ts` fix could address both. Worth cross-linking.

## Acceptance criteria

- The empty state's "Model Providers" is a clickable link to `/workspaces/${workspaceId}/settings/model-providers`, using the existing in-repo pattern (`ai-assistant-panel.tsx:150-157`).
- With zero models, the selector shows a neutral placeholder (e.g. "Select model"), not a phantom `claude-opus-4-8`.
- No regression: with no models the input stays disabled and nothing can be submitted.

Satyam296:
## Summary

Fixes both issues reported in #1500.

**Part 1 — dead-end empty state**

`ai-assistant-panel.tsx:172` had `"Workspace Settings → Model Providers"` wrapped in a plain `<span>`. Changed it to an `<a>` tag pointing to `/workspaces/${workspaceId}/settings/model-providers`, using the identical pattern already used a few lines above for the unsupported-model warning (line 151–156). `workspaceId` is already in scope since the empty state only renders when `llmModels` has resolved (which requires `workspaceId`).

**Part 2 — phantom `claude-opus-4-8`**

Root cause: while the `llm-models` query was loading, `flattenAvailableModels(undefined)` returned `FALLBACK_MODELS` (containing claude-opus-4-8). The reconcile effect picked it as the default. When the query resolved to empty `[]`, the effect hit `if (models.length === 0) return` and never cleared the selection.

Fix: when `models.length === 0` and `value.model` is non-empty, call `onChange({ model: "", ... })` to clear the phantom. The button already has `"Select model"` as its final fallback (`selectedModel?.label || value.model || "Select model"`), so it shows the neutral placeholder once cleared.

No change to the disabled state — the input remains disabled via `disabled={!projectId || !hasModels}`.

## Test plan

- [x] Two new tests in `model-selector.test.tsx`: phantom cleared when models resolve empty; placeholder shown when both selection and models are empty
- [ ] Manually verify: workspace with no system keys and no BYOK provider → empty state link is clickable → lands on `/workspaces/.../settings/model-providers`
- [ ] Verify: model selector shows "Select model", not "claude-opus-4-8", in the no-models state
- [ ] Verify: workspace with models configured is unaffected

Closes #1500

---
## Summary by cubic
Make the AI assistant’s no-models state actionable and stop phantom selections by showing fallback models only while the models query is pending and skipping reconcile during that window. Users can jump to model setup from the panel and the empty dropdown; the selector won’t auto-pick or backfill from fallbacks.

- **Bug Fixes**
  - Link “Workspace Settings → Model Providers” to `/workspaces/${workspaceId}/settings/model-providers` in the panel empty state and the selector’s “No models available” dropdown.
  - Gate fallbacks and default-pick on `isPending`; once settled, an empty response means no models. Skip reconcile while pending to avoid backfilling a provider onto partially-hydrated selections.
  - Update `resolve-model` to accept `isPending`, and wire through `ModelSelector` and `AgentModelLink`. Add tests for the panel link, pending behavior, placeholder label, and the dropdown link.

<sup>Written for commit ee1bae1eb6b8e3430d3c4593179124a0d35a4960. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1533?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

image
image

### DIFF
--- a/frontend/ui/src/features/ai-assistant/components/model-selector.tsx
+++ b/frontend/ui/src/features/ai-assistant/components/model-selector.tsx
@@ -47,7 +47,14 @@ export function ModelSelector({ value, onChange, workspaceId }: ModelSelectorPro
   // Without case 2 the selector would silently auto-pick a default when a
   // partially-hydrated saved selection arrives, clobbering the user's choice.
   useEffect(() => {
-    if (models.length === 0) return;
+    if (models.length === 0) {
+      // Query resolved with zero models — clear any phantom selection that was
+      // set from FALLBACK_MODELS while the query was still loading.
+      if (value.model) {
+        onChange({ model: "", provider: "", source: "system", adapter: "" });
+      }
+      return;
+    }
 
     const exact = models.find(
       (m) => m.id === value.model && m.provider === value.provider && m.source === value.source,
--- a/frontend/ui/src/features/ai-assistant/components/model-selector.tsx
+++ b/frontend/ui/src/features/ai-assistant/components/model-selector.tsx
@@ -29,33 +29,29 @@ function modelKey(m: { id?: string; model?: string; source: string; provider: st
 export function ModelSelector({ value, onChange, workspaceId }: ModelSelectorProps) {
   const [open, setOpen] = useState(false);
 
-  const { data } = useQuery({
+  const { data, isPending } = useQuery({
     queryKey: ["llm-models", workspaceId],
     queryFn: () => getAvailableLLMModels(workspaceId!),
     enabled: !!workspaceId,
   });
 
-  // BYOK models first, then system models. No deduplication.
-  const models = flattenAvailableModels(data);
+  // BYOK models first, then system models. No deduplication. Only shows the
+  // compiled-in fallback list while the query is still pending — once settled,
+  // an empty response means the workspace genuinely has no models.
+  const models = flattenAvailableModels(data, isPending);
 
   // Reconcile the incoming selection against the catalog:
   //   1. exact match on (model, provider, source) → check adapter; backfill if empty/wrong
   //   2. model-id-only match (legacy/hydrated state where the parent only has
   //      `model` saved, e.g. `project.rca_model: string`) → backfill the rest
   //   3. no match → preserve the current selection if the user already picked
-  //      one, and only auto-pick a default when the model is still empty.
+  //      one, and only auto-pick a default once the query has settled and the
+  //      model is still empty. Gating on settled state (rather than picking from
+  //      the pending fallback list and clearing it later) avoids ever showing a
+  //      phantom default the user didn't choose.
   // Without case 2 the selector would silently auto-pick a default when a
   // partially-hydrated saved selection arrives, clobbering the user's choice.
   useEffect(() => {
-    if (models.length === 0) {
-      // Query resolved with zero models — clear any phantom selection that was
-      // set from FALLBACK_MODELS while the query was still loading.
-      if (value.model) {
-        onChange({ model: "", provider: "", source: "system", adapter: "" });
-      }
-      return;
-    }
-
     const exact = models.find(
       (m) => m.id === value.model && m.provider === value.provider && m.source === value.source,
     );
@@ -64,7 +60,7 @@ export function ModelSelector({ value, onChange, workspaceId }: ModelSelectorPro
     const match = exact ?? modelOnly;
 
     if (!match) {
-      if (!value.model) {
+      if (!value.model && !isPending) {
         const pick = pickDefaultModel(models);
         if (pick) {
           onChange({
@@ -94,7 +90,7 @@ export function ModelSelector({ value, onChange, workspaceId }: ModelSelectorPro
         adapter: match.adapter,
       });
     }
-  }, [models, value, onChange]);
+  }, [models, value, onChange, isPending]);
 
   const selectedKey = modelKey({ id: value.model, source: value.source, provider: value.provider });
   const selectedModel = models.find((m) => modelKey(m) === selectedKey);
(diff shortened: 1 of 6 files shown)


frontend/ui/src/features/ai-assistant/components/model-selector.tsx
```diff
@@ -47,7 +47,14 @@ export function ModelSelector({ value, onChange, workspaceId }: ModelSelectorPro
   // Without case 2 the selector would silently auto-pick a default when a
   // partially-hydrated saved selection arrives, clobbering the user's choice.
   useEffect(() => {
-    if (models.length === 0) return;
+    if (models.length === 0) {
+      // Query resolved with zero models — clear any phantom selection that was
+      // set from FALLBACK_MODELS while the query was still loading.
+      if (value.model) {
+        onChange({ model: "", provider: "", source: "system", adapter: "" });
```
dark-sorceror: When tested e2e, the save button self-enables with zero user interaction. This clear on empty was added to the shared `ModelSelector` which all 4 surfaces use. Scope this to the surface that wants it (message-input) or track the auto-picked value in a ref and only clear that.

frontend/ui/src/features/ai-assistant/components/model-selector.tsx
dark-sorceror: When data is undefined, it still returns `FALLBACK_MDOELS`, so the phantom glitch still occurs during load and if the the `llm-models` query errors. I think gating this fallback on the query's settled state would fix the issue and remove the need for the clear-on-empty workaround entirely.

dark-sorceror, line 157: Can we add a link here

dark-sorceror (changes requested):
Great work! Some e2e tested bugs and minor issues. I was also thinking we go from a link for the configuring of a model to a button.

Satyam296:
changes implemented @dark-sorceror

frontend/ui/src/features/ai-assistant/components/model-selector.tsx
dark-sorceror: Resolved in 93aa422

dark-sorceror: Resolved in 93aa422

frontend/ui/src/features/ai-assistant/components/model-selector.tsx
dark-sorceror: Resolved in 93aa422

dark-sorceror:
LGTM

XinweiHe (approved):

Issue #1040, opened 2026-06-04
[Epic] Trace detail loading optimization — cut large-trace payload & redundant refetch


XinweiHe:
Large traces are slow to load and re-refresh in the UI. A real customer trace (67 spans) loads **~24 MB** of payload — ~30× this project's p99 — and the backend query reads **1.59 GiB / 2.8 s** in ClickHouse for that single trace (because `trace_id` isn't in the spans sort key, and `FINAL` forces all-column reads). The full payload is then re-downloaded on stream-complete and again on window-focus, and responses are uncompressed.

The fix is in two tiers:

- **Tier P1 (quick wins, parallel):** compression, large-span render responsiveness, size guard.
- **Tier P2 (structural):** two-phase loading — a lightweight span-tree skeleton + lazy per-span I/O, backed by a ClickHouse layout where trace lookups are index seeks and the tree query never reads blob columns.

Full design, measurements, and per-task acceptance criteria: `dev_docs/plans/2026-06-03-trace-detail-loading-optimization.md`.

### Sub-issues
- [ ] #1041 · P1-A · Gzip the REST API
- [ ] #1042 · P1-B · Renderer string-truncate + collapse
- [ ] #1044 · P1-D · Response-size guard + "trace too large" UX
- [ ] #1045 · P2-A · ClickHouse rebuild — trace_id-first ordering + no-IO projection ✅ *validated on staging (~1,625× less data read, 9s→179ms index seek)*
- [ ] #1046 · P2-B · Split trace-detail API into skeleton + per-span I/O endpoint
- [ ] #1047 · P2-C · Frontend lazy per-span I/O fetch + prefetch on hover
- [ ] #1048 · P2-D · Virtualize the span tree view

**Parked / deprioritized**
- #1043 · Eliminate redundant trace refetch — superseded by #1041 + #1046; the `trace_complete` refetch is a correctness safety net (live-completion `root-ends-first` subtlety) and becomes cheap once P2 lands, so we most likely keep it. Revisit only if the live-completion signal is fixed.

> **Considered and dropped:** ingest-side prompt/metadata dedup. Bespoke attribute transform that adds maintenance surface for an at-rest-only saving (already ZSTD-compressed); lazy-load solves the load cost instead.

### Execution waves
- **Wave 1 (parallel):** #1041 (gated on an SSE streaming check), #1042, #1045, #1048
- **Wave 2 (parallel):** #1046 (needs #1045); #1044 folds into #1046
[...]

PR #1534, opened 2026-07-11
fix(worker): preserve span-path columns when explicit metadata is set

### LINKED ISSUE #1498: Annotated spans lose `ids_path`/`path`, silently breaking live-tree repair
## Summary

Spans with explicit user metadata lose the internal span-path attributes the live view depends on. The worker stores metadata as a hard either/or: if `traceroot.span.metadata` is set, it is stored **instead of** the leftover-attributes fallback (`backend/worker/otel_transform.py:626-641`) — and `traceroot.span.ids_path`/`traceroot.span.path` only survive via that fallback. Pending-ancestor repair (`enrichSpansWithPending`) reads those keys from span metadata to synthesize placeholder parents, so for annotation-heavy traces the live tree silently degrades to root-pinned orphans on **both** delivery paths: live SSE spans and the base-fetch skeleton subset. Nothing errors and no test fails — the repair just stops working exactly on the spans users cared enough about to annotate.

## How to replicate

1. Instrument an agent run where child spans set explicit metadata (e.g. the SDK's span-metadata API) inside a long-lived root span.
2. Open the live view mid-run: annotated children whose parents haven't arrived render orphaned at the root instead of under pending placeholders, while unannotated siblings repair correctly.
3. Inspect the stored span in ClickHouse: `metadata` contains only the user's keys — `traceroot.span.ids_path`/`path` are gone.

## Root cause (confirmed)

- `backend/worker/otel_transform.py:626-641`: when `traceroot.span.ids_path`/`path` are present as span attributes they are not known-prefix attributes, so they normally land in stored metadata via the leftover-attrs fallback — but explicit `traceroot.span.metadata` short-circuits that branch entirely, so the path keys are dropped at ingest.
- All consumers assume the keys are in metadata: the frontend repair parses them from `span.metadata` (`frontend/ui/src/features/traces/utils/index.ts:43`), and the skeleton whitelist extracts the same two keys from the stored blob. Once dropped at ingest, no later layer can recover them.

## Impact

- Live-tree repair silently fails for annotated spans — disconnected trees mid-run on precisely the traces users invested in instrumenting.
- The failure is invisible: no error, no log, correct-looking data at completion.

## Acceptance criteria

- The two span-path keys survive ingest regardless of explicit metadata: either always injected into the stored metadata blob alongside user keys, or promoted to dedicated columns read by the skeleton whitelist and the live publish path.
- Pending-ancestor repair works for spans with explicit metadata on both the SSE path and the base-fetch skeleton path.
- User-facing metadata display is unchanged (the span detail panel already strips `traceroot.span.*` keys; that behavior is preserved or made unnecessary by dedicated columns).
- A worker-transform test asserts a span with explicit metadata still carries `ids_path`/`path` through to the stored record, and a frontend test asserts repair succeeds for an annotated span whose parent has not arrived.

YashManek1:
Closes #1498

## Root cause

`backend/worker/otel_transform.py` stored `traceroot.span.ids_path`/`traceroot.span.path` only via the leftover-attributes fallback. Explicit `traceroot.span.metadata` short-circuited that branch, so annotated spans lost the two keys `enrichSpansWithPending` needs for pending-ancestor repair — silently degrading to root-pinned orphans on both the live SSE
path and the base-fetch skeleton path.

## Fix

- Extract `ids_path`/`path` unconditionally at transform time into two new dedicated ClickHouse columns (migration `006_add_span_path_columns.sql`), independent of whether explicit metadata is present.
- Surface both fields on `SpanSkeletonResponse` (inherited by `SpanResponse`, shared with the public API) — no public/internal split, since this is already-public denormalized data (same category as `parent_span_id`), cheap and semantically stable.
- Cap both arrays at 128 entries, dropping the root-most prefix so the immediate-parent side of the chain — what repair actually consumes — survives. Truncation is index-aligned between `ids_path`/`path` and documented on both public schema fields.
- Frontend (`enrichSpansWithPending`) prefers the dedicated fields, falling back to parsing metadata for historical/cached spans that predate this change.

## Testing

- `backend/worker/tests/test_otel_transform_trace_naming.py`: regression test asserting a span with explicit metadata still carries `ids_path`/`path` through to the stored record, plus a companion test confirming the no-explicit-metadata path still work and the keys no longer leak into the `metadata` blob.
- `frontend/ui/src/features/traces/utils/index.test.ts`: asserts `enrichSpansWithPending` synthesizes ancestor placeholders from the dedicated fields for an annotated span whose parent hasn't arrived.
- Updated `tests/db/test_client.py`, `tests/rest/test_public_traces_export.py`, `tests/rest/test_public_traces_read.py`, `tests/rest/test_trace_reader.py` for the new columns/response fields.

## Proof
##Before
frame01-pending-repaired
##After
frame02-after-parents-arrive
##Video

https://github.com/user-attachments/assets/417266d0-04a4-4fe7-9990-1ab212b9c1e3

## Checklist

- [x] One logical change, scoped to #1498
- [x] Tests ship with the fix
- [x] `diff-cover` on changed lines (backend + frontend suites both green locally)
- [x] Matches existing comment-heavy style

---
## Summary by cubic
Fixes missing ancestor path data on annotated spans by extracting `ids_path` and `path` into dedicated columns and API fields, so the UI can rebuild pending parents correctly in live SSE and base-fetch flows. Addresses #1498.

- **Bug Fixes**
  - Extracted `traceroot.span.ids_path` and `traceroot.span.path` unconditionally into new ClickHouse columns and surfaced them on `SpanSkeletonResponse` and `SpanResponse` (public API).
  - Kept arrays index-aligned, exclude the current span, and cap at 128 by dropping the root-most prefix.
  - Updated `enrichSpansWithPending` to prefer the new fields, with metadata parsing as a fallback for historical/cached spans.
  - Adjusted backend query, OpenAPI schema, and tests to include the new fields.

- **Migration**
  - Run ClickHouse migration `006_add_span_path_columns.sql`.
  - Deploy backend and frontend together; responses include `ids_path` and `path` (default `[]`). Backward compatible with older data.

<sup>Written for commit f84acc8381fef2dae1be619378c97b77758e15ef. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1534?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/db/clickhouse/migrations/006_add_span_path_columns.sql
+++ b/backend/db/clickhouse/migrations/006_add_span_path_columns.sql
@@ -1,0 +1,22 @@
+-- +goose Up
+-- Internal tree-repair bookkeeping: root→current ancestor IDs (span_ids) and ancestor
+-- names for pending-span reconstruction. These are extracted from the SDK's
+-- `traceroot.span.ids_path` and `traceroot.span.path` OTel attributes. Previously
+-- stored only in the metadata blob, they were lost when explicit metadata was set,
+-- breaking frontend enrichSpansWithPending.
+--
+-- Dedicated columns fix both the live-SSE delivery path AND the base-fetch skeleton
+-- read (which skips metadata to save payload size, issue #1040), following the
+-- existing pattern used for git_source_* fields.
+--
+-- These are internal bookkeeping keys, not user-facing metadata — distinct from the
+-- metadata column which stores leftover attributes and user annotations.
+ALTER TABLE spans
+    ADD COLUMN IF NOT EXISTS ids_path Array(String) DEFAULT [],
+    ADD COLUMN IF NOT EXISTS path     Array(String) DEFAULT [];
+
+-- +goose Down
+ALTER TABLE spans
+    DROP COLUMN IF EXISTS ids_path;
+ALTER TABLE spans
+    DROP COLUMN IF EXISTS path;
--- a/backend/worker/otel_transform.py
+++ b/backend/worker/otel_transform.py
@@ -52,6 +52,38 @@ def _scope_skips_text_token_estimation(scope_name: str | None) -> bool:
     return scope_name in _SKIP_TEXT_TOKEN_ESTIMATION_SCOPES
 
 
+# Trace depth realistically stays far below this. The cap exists for malformed or
+# adversarial ``traceroot.span.ids_path`` / ``traceroot.span.path`` attributes: keep
+# the parent-near tail used for live tree repair, and drop only the root-most prefix
+# once the arrays are too deep.
+MAX_SPAN_PATH_DEPTH = 128
+
+
+def _extract_span_path_columns(
+    ids_path_value: object,
+    path_value: object,
+) -> tuple[list[str], list[str]]:
+    """Return bounded, index-aligned ancestor IDs and names for live tree repair.
+
+    SDKs historically emit ``traceroot.span.path`` as root→current names while
+    ``traceroot.span.ids_path`` is root→parent IDs. The stored/public columns are
+    a stricter ancestor-chain contract: both arrays exclude this span, align
+    index-for-index, and are capped by dropping the root-most prefix so the
+    immediate parent side of the chain remains repairable.
+    """
+
+    ids_path = list(ids_path_value) if isinstance(ids_path_value, (list, tuple)) else []
+    path = list(path_value) if isinstance(path_value, (list, tuple)) else []
+
+    # Normalize legacy root→current path values to root→parent ancestor names.
+    ancestor_path = path[:-1] if len(path) == len(ids_path) + 1 else path
+    bounded_depth = min(len(ids_path), len(ancestor_path), MAX_SPAN_PATH_DEPTH)
+    if bounded_depth == 0:
+        return [], []
+
+    return ids_path[-bounded_depth:], ancestor_path[-bounded_depth:]
+
+
 # Attributes that are already extracted into dedicated fields
 _KNOWN_ATTRIBUTE_PREFIXES = {
     "traceroot.span.input",
@@ -59,6 +91,8 @@ _KNOWN_ATTRIBUTE_PREFIXES = {
     "traceroot.span.type",
     "traceroot.span.metadata",
     "traceroot.span.tags",
+    "traceroot.span.ids_path",  # Internal tree-repair bookkeeping: now in dedicated column
+    "traceroot.span.path",  # Internal tree-repair bookkeeping: now in dedicated column
     "traceroot.llm.",
     "traceroot.trace.",
     "traceroot.environment",
@@ -396,6 +430,21 @@ def transform_otel_to_clickhouse(
                     span_record["git_source_line"] = git_source_line
                 if git_source_function is not None:
                     span_record["git_source_function"] = git_source_function
+
+                # Extract tree-repair bookkeeping: ancestor IDs and names.
+                # These are internal; they drive frontend enrichSpansWithPending to
+                # synthesize placeholders for children whose real parent hasn't streamed
+                # yet. Stored as dedicated columns (not in metadata blob) so they survive
+                # the base-fetch skeleton read, which skips metadata for payload size.
+                # Capped at depth so a malformed/adversarial OTel payload (an SDK bug or a
+                # misbehaving client) can't force an unbounded array into ClickHouse or
+                # balloon worker memory during the transform.
+                ids_path = span_attrs.get("traceroot.span.ids_path")
+                path_value = span_attrs.get("traceroot.span.path")
+                span_record["ids_path"], span_record["path"] = _extract_span_path_columns(
+                    ids_path,
+                    path_value,
+                )
 
                 # Extraction priority (first key present/not-None wins):
                 # 1. TraceRoot SDK: user-provided explicit values — highest authority.
--- a/backend/worker/tests/test_otel_transform_trace_naming.py
+++ b/backend/worker/tests/test_otel_transform_trace_naming.py
@@ -3,7 +3,7 @@
 Covers the corner cases introduced during live-trace-streaming work:
   - Eager trace name should use traceroot.span.path[0], not the span's own name
   - When a batch has spans at multiple depths, the shallowest span's path[0] wins
-  - traceroot.span.path / traceroot.span.ids_path must be preserved in span metadata
+  - traceroot.span.path / traceroot.span.ids_path must be extracted to dedicated columns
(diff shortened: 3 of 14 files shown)


backend/worker/otel_transform.py
```diff
@@ -52,13 +52,47 @@ def _scope_skips_text_token_estimation(scope_name: str | None) -> bool:
...
+    """Return bounded, index-aligned ancestor IDs and names for live tree repair.
+
+    SDKs historically emit ``traceroot.span.path`` as root→current names while
+    ``traceroot.span.ids_path`` is root→parent IDs. The stored/public columns are
+    a stricter ancestor-chain contract: both arrays exclude this span, align
+    index-for-index, and are capped by dropping the root-most prefix so the
+    immediate parent side of the chain remains repairable.
+    """
+
+    ids_path = list(ids_path_value) if isinstance(ids_path_value, (list, tuple)) else []
```
dark-sorceror, line 75: Validates the array container but never the element type. `extract_attribute_value` unwraps an OTLP `arrayValue` element-wise, so `intValue/bootValue/kvlistValue/nested` arrays drive as non-strings and flow into an `Array(String)` ClickHouse column. `insert_spans_batch` then raises (`'int' object has no attribute 'encode'`). The retrigger in `ingest_tasks` when Celery retries fails identically. So one malformed span permanently drops every span in the batch.

backend/db/clickhouse/migrations/006_add_span_path_columns.sql
```diff
@@ -0,0 +1,22 @@
...
+-- stored only in the metadata blob, they were lost when explicit metadata was set,
+-- breaking frontend enrichSpansWithPending.
+--
+-- Dedicated columns fix both the live-SSE delivery path AND the base-fetch skeleton
+-- read (which skips metadata to save payload size, issue #1040), following the
+-- existing pattern used for git_source_* fields.
+--
+-- These are internal bookkeeping keys, not user-facing metadata — distinct from the
+-- metadata column which stores leftover attributes and user annotations.
+ALTER TABLE spans
```
dark-sorceror, line 14: Missing the `spans_no_io_by_start_time` PROJECTION from migration 005. Either extend the projection or document in the migration comment that this read never uses the projection.

backend/worker/otel_transform.py
```diff
@@ -52,13 +52,47 @@ def _scope_skips_text_token_estimation(scope_name: str | None) -> bool:
...
+    path = list(path_value) if isinstance(path_value, (list, tuple)) else []
+
+    # Normalize legacy root→current path values to root→parent ancestor names.
+    ancestor_path = path[:-1] if len(path) == len(ids_path) + 1 else path
```
dark-sorceror, line 79: Can we reject mismatches instead of truncating into misalignment

backend/worker/tests/test_otel_transform_trace_naming.py
```diff
@@ -260,3 +272,147 @@ def test_sdk_name_version_preserved_in_metadata():
...
+    # Path attrs must still reach dedicated fields
+    assert span_record["ids_path"] == ids_path, (
+        f"ids_path must be extracted to dedicated field, got {span_record.get('ids_path')}"
+    )
+    assert span_record["path"] == path[:-1], (
+        f"path must be extracted to dedicated field, got {span_record.get('path')}"
+    )
+
+    # Custom attrs should be in metadata (they're not known attributes)
+    if "metadata" in span_record:
```
dark-sorceror, line 378: Make this unconditional, since if the leftover-attribute sweep were ever skipped entirely, metadata would be absent, then this condition would be false and the test would stay green while custom attributes silently vanished.

tests/rest/test_trace_reader.py
```diff
@@ -183,6 +183,8 @@ def side_effect(query, parameters=None):
                         "file.py",  # git_source_file
                         12,  # git_source_line
                         "fn",  # git_source_function
+                        ["root-id"],  # ids_path
+                        ["root"],  # path
```
dark-sorceror, line 187: Can we add tests for these

dark-sorceror (changes requested):
Great work! Looks good E2E - verified on ClickHouse both skeleton read and live SSE, migration 006 is idempotent. However some edge cases and minor issues need to be fixed.

YashManek1:
while working on it I noticed PR #1277 (merged into main on 2026-07-14, fixing #1208) already fixes this exact root cause.
Their otel_transform.py change does the same thing #1498 asks for: it now merges traceroot.span.ids_path/traceroot.span.path into the stored metadata blob even when explicit traceroot.span.metadata is set, instead of the old either/or that dropped them. They even added a regression test for the identical scenario (test_span_path_attributes_survive_alongside_explicit_metadata).
The difference is just mechanism: I went with two new dedicated ClickHouse columns (migration 006), they went with keeping it inside the existing metadata blob and extracting it back out via JSONExtract in trace_reader.py — no schema change needed. Since main already ships a working fix, merging this would just add a second, conflicting representation of the same data.
Recommend closing #1498 too, since #1277 already satisfies its acceptance criteria. Happy to open a small follow-up if there's any gap in their version, but from what I've checked it holds up.

dark-sorceror:
@YashManek1 You are right, this is mostly fixed by #1277 which was merged very recently. Seems like that PR grew from just containing the read-path/skeleton fix for #1208 to include the ingest-side that dropped `traceroot.span.ids_path/traceroot.span.path` on annotated spans bug was also covered before merge.

Acceptance criteria check against merged main:
- Path keys survive ingest alongside explicit metadata: otel_transform.py now merges span_path_attrs into the explicit metadata object instead of replacing it. Pinned by test_span_path_attributes_survive_alongside_explicit_metadata.
- Repair works on both delivery paths: live SSE spans carry the merged blob, and the base-fetch skeleton now server-extracts just the two path keys via JSONExtract in trace_reader.py (full blob stays off the wire, preserving the #1040 payload behavior). Pinned by test_spans_query_omits_io_blobs_and_extracts_only_span_path_attrs and TestDashboardKeepsSpanTreeMetadata.
- User-facing metadata display unchanged: the detail panel's existing traceroot.span.* filter handles the merged keys; public routes drop the subset entirely (drop_span_tree_metadata).
- Worker + frontend regression tests: both present in #1277.

Closing as completed.

PR #1560, opened 2026-07-14
fix: make gemini-3-flash pricing match both GA and preview model IDs

### LINKED ISSUE #1558: Cost is $0 for GA `google/gemini-3-flash` — matchPattern requires the `-preview` suffix, so the shipped model is unpriced
## Summary

A **successful** LLM call on `google/gemini-3-flash` (Google's GA flash model, routed through Vercel AI Gateway) records **no cost**, despite reporting real token usage (1.4K in → 95 out). Root cause: `standard-model-prices.json` only carries `gemini-3-flash-preview`, and that entry's `matchPattern` **requires the literal `-preview` suffix** — so the GA id (`gemini-3-flash`, with or without the `google/` gateway prefix) matches nothing and `get_model_price` returns `None`.

This is a missing / too-strict SKU problem — sibling to #1545, and **not** the gateway-prefix normalization problem in #1556. The `google/` prefix is already accepted by the gemini patterns; the blocker is the required `-preview`.

## Real-world repro (customer app: AI-TryOn / faishion-web-app)

- **The cost-loss case:** span `ai.streamText.doStream`, Span Kind `llm`, model badge **`google/gemini-3-flash`**, **success**, tokens **`1.4K → 95 (1.5K)`** — **no cost rendered**. Trace `89dc27e67be0303fb206de80d35d3a35`, routed via Vercel AI Gateway (`fetch POST https://ai-gateway.ve…`).
- Same model spelled **`gemini-3-flash`** (direct Google, no prefix) returns an API error — *"models/gemini-3-flash is not found for API version v1beta"* — tokens `25 → 0`, also no cost (errored / 0 output). Confirms the id in use is `gemini-3-flash` and that the app hits it both directly and via the gateway.

## Root cause

`frontend/packages/core/src/standard-model-prices.json` — the only Gen-3 flash entry:

```
gemini-3-flash-preview : (?i)^(google\/|models\/)?gemini-3-flash-preview(-[\d-]+)?$
```

Matching runs in `backend/worker/tokens/pricing.py` → `get_model_price(model)` (exact match, then regex over each `matchPattern`). Both `google/gemini-3-flash` and bare `gemini-3-flash` MISS → `None` → cost left unset (cf. #1545 for the same silent-`$0` symptom).

## Fix (data-only, one line): make `-preview` optional

```
(?i)^(google\/|models\/)?gemini-3-flash(-preview)?(-[\d-]+)?$
```

Verified against all four spellings:

| model string | current | proposed |
|---|---|---|
| `google/gemini-3-flash` (repro) | MISS | HIT |
| `gemini-3-flash` | MISS | HIT |
| `gemini-3-flash-preview` (existing) | HIT | HIT |
| `google/gemini-3-flash-preview` | HIT | HIT |

This reuses the existing preview rates ($0.50 / $3.00 per 1M input/output). **Confirm against Google's public pricing when landing** — GA rates may differ from preview. Alternative: add a separate `gemini-3-flash` entry with its own rates.

## Related
- #1545 — sibling missing-SKU (`gpt-5.1-codex-mini` / codex family)
- #1556 — sibling gateway-prefix normalization (different root cause + different fix)
- Match logic: `backend/worker/tokens/pricing.py`; price source: `frontend/packages/core/src/standard-model-prices.json`; tests: `tests/worker/tokens/test_pricing.py`

lesbass:
## Summary

The `gemini-3-flash-preview` matchPattern in `standard-model-prices.json` required the literal `-preview` suffix, causing GA model IDs like `google/gemini-3-flash` to return no pricing (cost reads $0 despite real token usage).

## What Changed

- **`frontend/packages/core/src/standard-model-prices.json`**: Made `-preview` optional in the `matchPattern` so both GA and preview spellings resolve to the same entry: `(?i)^(google\\/|models\\/)?gemini-3-flash(-preview)?(-[\d-]+)?$`
- **`tests/worker/tokens/test_pricing.py`**: Added test cases for `gemini-3-flash` and `google/gemini-3-flash`

## Verification

All four model string spellings now match:

| model string | before | after |
|---|---|---|
| `google/gemini-3-flash` | MISS | HIT |
| `gemini-3-flash` | MISS | HIT |
| `gemini-3-flash-preview` | HIT | HIT |
| `google/gemini-3-flash-preview` | HIT | HIT |

Closes #1558

---
## Summary by cubic
Fixes pricing lookup for Gemini 3 Flash by matching both GA and preview model IDs. GA IDs like `google/gemini-3-flash` now return correct costs instead of $0.

- **Bug Fixes**
  - Updated `frontend/packages/core/src/standard-model-prices.json` to make `-preview` optional in the pattern: `(?i)^(google\/|models\/)?gemini-3-flash(-preview)?(-[\d-]+)?$`.
  - Added tests in `tests/worker/tokens/test_pricing.py` to ensure `gemini-3-flash` and `google/gemini-3-flash` resolve to the `gemini-3-flash-preview` pricing entry.

<sup>Written for commit 22edfb2537187daa549e710e49b1c31fdb2eb0e1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1560?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/packages/core/src/standard-model-prices.json
+++ b/frontend/packages/core/src/standard-model-prices.json
@@ -797,7 +797,7 @@
   {
     "id": "tr-gemini-3-flash-preview",
     "modelName": "gemini-3-flash-preview",
-    "matchPattern": "(?i)^(google\\/|models\\/)?gemini-3-flash-preview(-[\\d-]+)?$",
+    "matchPattern": "(?i)^(google\\/|models\\/)?gemini-3-flash(-preview)?(-[\\d-]+)?$",
     "provider": "google",
     "prices": {
       "input": 0.0000005,
--- a/tests/worker/tokens/test_pricing.py
+++ b/tests/worker/tokens/test_pricing.py
@@ -149,6 +149,8 @@ GEMINI_MODEL_CASES = [
     ("gemini-3.1-flash-lite", "gemini-3.1-flash-lite"),
     ("gemini-3.1-flash-lite-preview", "gemini-3.1-flash-lite-preview"),
     ("gemini-3-flash-preview", "gemini-3-flash-preview"),
+    ("gemini-3-flash", "gemini-3-flash-preview"),
+    ("google/gemini-3-flash", "gemini-3-flash-preview"),
     ("gemini-2.5-pro", "gemini-2.5-pro"),
     ("gemini-2.5-flash", "gemini-2.5-flash"),
     ("gemini-2.5-flash-lite", "gemini-2.5-flash-lite"),


ka1kqi:
Note from #1593 (Anthropic fast-mode + dot-notation pricing, about to merge), which touches the same price table and test file: nothing there overlaps this fix — same bug class (pattern stricter than real-world ids → silent $0), different provider — so we didn't fold anything together, and this PR stands on its own. Three things that may help it along:

1. **Light rebase likely** after #1593 lands: it edits different JSON entries but appends to the same `tests/worker/tokens/test_pricing.py`, so expect a trivial conflict there.
2. **Your bug class is now log-visible**: #1593 added a deduped warning in `get_model_price` for any model id matching no entry — after it merges, GA `gemini-3-flash` ids pricing at $0 will show up in worker logs, which also gives you a clean way to confirm this fix in staging.
3. If useful, #1593 introduced an "exactly one pattern matches" parametrized guard over its test ids — applied here it would prove `gemini-3-flash` and `gemini-3-flash-preview` can never both match one id, which is the main regression risk when loosening a suffix requirement.

lesbass:
Closing this PR after 14 days with no human reviewer engagement (per our stale PR policy). The change is small and low-risk — 3 additions, 1 deletion, CI all green, CLA signed. A rebase may be needed after the merged #1593. Happy to reopen if a maintainer expresses interest.

PR: https://github.com/traceroot-ai/traceroot/pull/1560

PR #1566, opened 2026-07-15
fix(pricing): strip gateway/router prefixes before model price lookup

### LINKED ISSUE #1556: Pricing silently records $0 for gateway/router-prefixed model names (vertex_ai/, openrouter/, litellm/, inconsistent azure/) — prefix handling is hand-encoded per entry
## Summary

When a model name arrives with an **API-gateway / router prefix** (e.g. `vertex_ai/gemini-2.5-pro`, `azure/gpt-5.4`, `openrouter/anthropic/claude-opus-4-8`, `litellm/openai/gpt-5`), cost silently resolves to `$0`. This happens because pricing matching enumerates the accepted prefixes *inside each model's regex*, one entry at a time — so coverage is **inconsistent** across entries and **incomplete** for the many prefixes real gateways emit.

This is the systemic version of the one-off gaps patched in #877 (Bedrock/Vertex), #1545 (missing codex SKU), and PR #977 ("fix matchPattern gaps that silently record $0"). Rather than keep patching individual entries, we should normalize the prefix once.

## Root cause

Matching lives in `backend/worker/tokens/pricing.py` → `get_model_price(model)`:

1. exact match on `model_name`
2. regex fallback: `re.search(entry["match_pattern"], model, re.IGNORECASE)` over every entry

Every one of the 87 patterns in `frontend/packages/core/src/standard-model-prices.json` is anchored `^…$` and **hard-codes which gateway prefixes it will tolerate**, e.g.:

```
gpt-5.6-sol   (?i)^(openai\/|azure\/)?(gpt-5\.6-sol)(-[\d-]+)?$      # allows openai/ AND azure/
gpt-5.4       (?i)^(openai\/)?(gpt-5\.4)(-[\d-]+)?$                  # allows ONLY openai/  ← inconsistent
gemini-2.5-pro(?i)^(google\/|models\/)?gemini-2\.5-pro(-[\d-]+)?$   # allows google/ , models/ — NOT vertex_ai/
claude-opus-4-8 (?i)^(anthropic\/)?(claude-opus-4-8…|(eu\.|us\.|apac\.|global\.)?anthropic\.claude-opus-4-8…) # most complete
```

Because the prefix set is authored by hand per entry, it drifts: some SKUs accept `azure/`, siblings don't; no entry accepts `vertex_ai/`, `openrouter/`, `litellm/`, or double-prefixes; and every newly added prefix would require editing all ~87 patterns.

## Reproduction (against current `main`)

Running each string through `get_model_price` (exact-then-regex, same logic as the worker):

| model string | matches? | why |
|---|---|---|
| `vertex_ai/gemini-2.5-pro` | ❌ `$0` | gemini patterns allow only `google/` \| `models/` |
| `azure/gpt-5.4` | ❌ `$0` | `gpt-5.4` allows only `openai/`, though `gpt-5.6-*` allows `azure/` too |
| `openrouter/anthropic/claude-opus-4-8` | ❌ `$0` | double prefix not handled |
| `litellm/openai/gpt-5` | ❌ `$0` | double prefix not handled |
| `openai/gpt-5.6-sol` | ✅ | pattern already lists `openai/` \| `azure/` |
| `anthropic/claude-opus-4-8` | ✅ | pattern lists `anthropic/` |
| `google/gemini-2.5-pro` | ✅ | pattern lists `google/` |

Note on the two originally-reported examples:
- **`openai/gpt-5.6-sol`** already resolves correctly on `main` (a recent PR added the `openai/`\|`azure/` alternation to that entry).
- **`google/gemini-flash-3`** fails because there is **no `gemini-flash-3` SKU** in the JSON at all — a *missing-entry* problem (same class as #1545), independent of the prefix issue. The `google/` prefix itself works for existing gemini entries.

## Impact

- Any customer routing through an AI gateway/router (OpenRouter, LiteLLM, Portkey, Vertex-via-LiteLLM, Azure-via-router, …) can silently get `$0` cost on otherwise-known models.
- Failure is silent: `get_model_price` returns `None`, cost is left unset, and it looks like "cost just isn't rendering" (cf. #1545 screenshots) rather than a match failure.
- Maintenance burden grows with every new prefix × every entry.

## Proposed direction

Prefer normalization over more per-entry patching. Options, roughly in order of leverage:

1. **Normalize the model name once before matching** — strip a known set of gateway/router prefixes (`openai/`, `azure/`, `google/`, `googleai/`, `vertex_ai/`, `anthropic/`, `bedrock/`, `openrouter/`, `litellm/`, `models/`, and repeated/chained prefixes) in `get_model_price` before the exact/regex passes, then match the bare model id. This fixes all rows at once and lets the JSON patterns drop most of the baked-in `(provider\/)?` alternations. (Keep Bedrock's `us.anthropic.…` / Vertex `@date` forms, which are distinct id shapes, not simple slash prefixes.)
2. **Standardize the alternation in the generator** as a belt-and-suspenders measure — if we keep any prefixes in-pattern, emit the same canonical prefix set for every entry from the sync/generation script so siblings can't drift (e.g. `azure/` present on all OpenAI SKUs, not just some).
3. **Per-project custom model definitions** (larger, self-serve) — let a user register a model id + match pattern + prices scoped to their project that overrides the global defaults, so an unusual gateway id can be priced without a PR to the shared JSON. Some mature pricing systems make this the primary escape hatch; today our `standard_models` table is global-only with no override path.

(1) is the smallest high-leverage fix and directly resolves the reported gateway-prefix failures; (3) is the durable long-term answer for the long tail of ids we'll never enumerate.

Separately, genuinely-missing SKUs like `gemini-flash-3` should be added as normal price entries (track under the missing-model line, e.g. alongside #1545).

## Pointers
- Match logic: `backend/worker/tokens/pricing.py` (`get_model_price`)
- Cost call site: `backend/worker/otel_transform.py` (`cost_from_buckets(get_model_price(model_name), …)`)
- Price table source: `frontend/packages/core/src/standard-model-prices.json` (+ `sync-standard-prices.ts`)
- Tests: `tests/worker/tokens/test_pricing.py`
- Related: #877, #1545, PR #977

Satyam296:
## Problem

When a model name is sent with a gateway/router prefix — e.g. `vertex_ai/gemini-2.5-pro`, or `openrouter/anthropic/claude-opus-4-8` (OpenRouter-via-LiteLLM, where LiteLLM prepends its `openrouter/` provider tag onto OpenRouter's own `anthropic/<model>` id) — `get_model_price` returns `None` and the span is recorded with `$0` cost. This happens because the exact-match pass compares the full string (including prefix) against `model_name`, and the regex patterns in `standard-model-prices.json` only cover a subset of prefixes (e.g. `(openai\/)?`, `(azure\/)?`) — never `vertex_ai/`, `openrouter/`, `googleai/`, or chained double-prefixes.

Fixes #1556.

## Solution

Added `_strip_gateway_prefix(model)` in `backend/worker/tokens/pricing.py`. It iterates over the leading `/`-separated segments and removes each one while it is a known gateway name. This normalises the model string before it enters the exact-match / regex lookup:

```
vertex_ai/gemini-2.5-pro            → gemini-2.5-pro
openrouter/anthropic/claude-opus-4-8 → claude-opus-4-8  (two iterations)
azure/gpt-5.4                       → gpt-5.4
```

Note: `litellm` is deliberately **not** in the recognised prefix set. LiteLLM never emits a literal `litellm/` segment on the wire — when it routes through OpenRouter the id is `openrouter/<provider>/<model>`, not `litellm/openrouter/...` — so `litellm/gpt-5` correctly passes through unchanged (verified by a new unit test).

Bedrock dot-format (`us.anthropic.claude-…`) and Vertex `@date` variants (`claude-4-8-opus@20260601`) use `.` / `@` rather than `/`, so `model.find("/")` returns -1 immediately and they pass through unchanged — all existing Bedrock/Vertex test cases still pass.

The existing per-provider regex alternations (`(openai\/)?`, `(azure\/)?`, etc.) remain in `standard-model-prices.json` and continue to work; the normalisation step just makes every pattern work for every gateway prefix, not just the ones whose maintainers anticipated.

## Tests

- **`TestStripGatewayPrefix`** — 7 unit tests for the normaliser: no-prefix, single prefix, double prefix, unknown prefix (unchanged), `litellm/` specifically unchanged (not a recognised prefix), Bedrock/Vertex formats (unchanged).
- **`TestGatewayPrefixedModelIds`** — 12 parametrised integration tests against the real `standard-model-prices.json`, covering the previously-broken cases: `vertex_ai/`, `azure/gpt-5.4`, `openrouter/`, `openrouter/<provider>/`, `googleai/`.
- All existing tests continue to pass; **213 total** in `test_pricing.py`.

## Checklist

- [x] `backend/worker/tokens/pricing.py` — added `_GATEWAY_PREFIXES`, `_strip_gateway_prefix`, called from `get_model_price`
- [x] `tests/worker/tokens/test_pricing.py` — added `TestStripGatewayPrefix` + `TestGatewayPrefixedModelIds`
- [x] All 213 tests pass (`python -m pytest tests/worker/tokens/test_pricing.py`)
- [x] No changes to public API signatures

---
## Summary by cubic
Normalize model IDs by stripping gateway/router prefixes before price lookup so prefixed models resolve and costs are calculated correctly. Fixes #1556.

- **Bug Fixes**
  - Add `_strip_gateway_prefix` to remove known prefixes (e.g., `openai/`, `azure/`, `google/`, `googleai/`, `vertex_ai/`, `anthropic/`, `bedrock/`, `openrouter/`), including chained prefixes.
  - Call it in `get_model_price` before exact and regex matching.
  - Leave Bedrock dot-format and Vertex `@date` variants unchanged; do not strip `litellm/` (it isn’t emitted). Add a unit test to lock this in.
  - Add tests: unit tests for the normalizer, integration tests for prefixed IDs, and a guard that each stripped ID matches exactly one price entry.

<sup>Written for commit e2508b91464b1daf639e9704f0fccc9487ea76ce. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1566?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/worker/tokens/pricing.py
+++ b/backend/worker/tokens/pricing.py
@@ -77,17 +77,66 @@ def _load_cache() -> list[dict]:
 
 
 # ---------------------------------------------------------------------------
+# Gateway/router prefix normalisation
+# ---------------------------------------------------------------------------
+
+# Prefixes injected by LiteLLM, OpenRouter, and cloud provider SDKs before
+# the bare model name.  Strip them so `get_model_price` can match models like
+# `vertex_ai/gemini-2.5-pro` or `openrouter/anthropic/claude-opus-4-8` against
+# the same patterns used for plain `gemini-2.5-pro` / `claude-opus-4-8`.
+_GATEWAY_PREFIXES = frozenset(
+    {
+        "openai",
+        "azure",
+        "google",
+        "googleai",
+        "vertex_ai",
+        "anthropic",
+        "bedrock",
+        "openrouter",
+        "litellm",
+        "models",
+        "deepseek",
+        "xai",
+        "moonshot",
+        "zai",
+    }
+)
+
+
+def _strip_gateway_prefix(model: str) -> str:
+    """Strip one or more leading gateway/router prefixes separated by ``/``.
+
+    Iterates so chained prefixes like ``openrouter/anthropic/claude-opus-4-8``
+    or ``litellm/openai/gpt-5`` are fully reduced.  Stops as soon as the first
+    path segment is not a recognised gateway name, leaving Bedrock dot-format
+    (``us.anthropic.claude-...``) and Vertex ``@date`` variants unchanged.
+    """
+    while True:
+        slash = model.find("/")
+        if slash == -1:
+            break
+        prefix = model[:slash].lower()
+        if prefix not in _GATEWAY_PREFIXES:
+            break
+        model = model[slash + 1 :]
+    return model
+
+
+# ---------------------------------------------------------------------------
 # Public API (unchanged signatures)
 # ---------------------------------------------------------------------------
 
 
 def get_model_price(model: str) -> dict[str, float] | None:
-    """Lookup price for model. Tries exact match, then regex fallback.
+    """Lookup price for model. Normalises gateway prefixes, then tries exact
+    match and regex fallback.
 
     Returns dict with keys like ``input``, ``output``, ``cacheRead``, ``cacheWrite``
     (values in USD per token), or None if not found.
     """
     cache = _load_cache()
+    model = _strip_gateway_prefix(model)
 
     # Exact match on model_name
     for entry in cache:
--- a/backend/worker/tokens/pricing.py
+++ b/backend/worker/tokens/pricing.py
@@ -94,7 +94,6 @@ _GATEWAY_PREFIXES = frozenset(
         "anthropic",
         "bedrock",
         "openrouter",
-        "litellm",
         "models",
         "deepseek",
         "xai",
@@ -107,9 +106,9 @@ _GATEWAY_PREFIXES = frozenset(
 def _strip_gateway_prefix(model: str) -> str:
     """Strip one or more leading gateway/router prefixes separated by ``/``.
 
-    Iterates so chained prefixes like ``openrouter/anthropic/claude-opus-4-8``
-    or ``litellm/openai/gpt-5`` are fully reduced.  Stops as soon as the first
-    path segment is not a recognised gateway name, leaving Bedrock dot-format
+    Iterates so a chained prefix like ``openrouter/anthropic/claude-opus-4-8``
+    is fully reduced.  Stops as soon as the first path segment is not a
+    recognised gateway name, leaving Bedrock dot-format
     (``us.anthropic.claude-...``) and Vertex ``@date`` variants unchanged.
     """
     while True:
--- a/tests/worker/tokens/test_pricing.py
+++ b/tests/worker/tokens/test_pricing.py
@@ -6,7 +6,7 @@ from unittest.mock import patch
 
 import pytest
 
-from worker.tokens.pricing import calculate_cost, get_model_price
+from worker.tokens.pricing import _strip_gateway_prefix, calculate_cost, get_model_price
 
 MATCHED_MODEL_NAME = "__matched_model_name"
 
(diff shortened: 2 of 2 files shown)


Satyam296:
@XinweiHe @dark-sorceror this one's ready for a look . Would you be able to take a look when you get a chance?

ka1kqi:
Heads-up from #1593 (fast-mode rate cards + dot-notation tolerance for Anthropic ids, fixes #1581), which is complementary to this PR: while runtime-verifying it we probed `openrouter/anthropic/claude-opus-4.8-fast` against the updated table and confirmed it still matches nothing — the router-prefix stripping here is exactly the missing half, and once both land, the full customer string normalizes and then resolves to the new fast rate card.

Two things from that work that may be useful here:

1. `get_model_price` now logs a deduped warning for any model id that matches no entry (once per id per process), so after this PR merges, any prefix shape it doesn't strip will show up in worker logs instead of silently pricing at $0.
2. If helpful, #1593 added an "exactly one pattern matches" parametrized guard over its test ids — the same idiom applied to this PR's stripped outputs would lock in that normalization never causes a string to become ambiguous between two entries.

Satyam296:
Good catch @ka1kqi  added that guard test, mirrors what #1593 did. Re-verified all four cases from #1556 resolve now too. Pushed as ac6c7aa9, thanks for the pointer!

dark-sorceror:
LGTM

backend/worker/tokens/pricing.py
```diff
@@ -76,18 +76,67 @@ def _load_cache() -> list[dict]:
...
+    {
+        "openai",
+        "azure",
+        "google",
+        "googleai",
+        "vertex_ai",
+        "anthropic",
+        "bedrock",
+        "openrouter",
+        "litellm",
```
dark-sorceror: I'm pretty sure LiteLLM never emits this. See https://docs.litellm.ai/docs/providers/openrouter.  So remove this from the gateway prefixes and the `litellm/...` fixtures in the test files.

dark-sorceror (changes requested):
Some comment after re-review. The PR description is also a bit confusing, these double-prefix examples are OpenRouter-via-LiteLLM, not OpenRouter.

backend/worker/tokens/pricing.py
Satyam296: Good catch, you're right. Removed litellm from the prefix set and the litellm/... test fixtures, added a test locking in that litellm/gpt-5 passes through unchanged. Pushed as e2508b91.

Satyam296:
@dark-sorceror thanks for catching that  you're right, LiteLLM never actually emits a `litellm/` segment. Removed it from the prefix set and the test fixtures, added a test locking in that `litellm/gpt-5` passes through unchanged. All the real cases from #1556 still resolve. Pushed as e2508b91, 17/17 CI green , let me know if you get a chance to take another look.

PR #1569, opened 2026-07-15
fix(clickhouse): write the environment column so detector trigger conditions can match

### LINKED ISSUE #1567: Environment trigger conditions never match: ingest drops the environment attribute before ClickHouse
## Summary

Environment trigger conditions never work in production — and `environment` is the **only** field the trigger editor offers (`trigger-editor.tsx:29-36`), so this is the product's entire trigger-filtering feature:

- `environment = X` **never fires**, even on matching traces.
- `environment != X` **fires on everything**, including the excluded environment.

The SDK's `traceroot.environment` attribute is extracted by the transform but dropped at the insert layer, so the ClickHouse `environment` column is always NULL and evaluation falls into the missing-field path. No error is raised anywhere. This is the mirror image of #1506: that fixed "fires when it shouldn't"; this is "never fires when it should."

## How to replicate

1. Create a detector with condition `environment = production` (UI), sampling 100%.
2. Send SDK traces with environment set to `production`.
3. The detector's Runs tab stays empty — it never fires. Flip the condition to `environment != staging` and send staging traces: it now fires on them.

Confirmed e2e by mirroring `process_s3_traces` (`ingest_tasks.py:105-157`): the transform output carries `environment = "production"`, the ClickHouse row read back after `insert_spans_batch` is **NULL**, and an empty-conditions control detector fires on the same traces (pipeline otherwise healthy).

## Impact

- The only trigger-filtering knob does nothing (`=`) or inverts (`!=`), silently — no exception, no log, no failed run.
- `!=` conditions burn LLM evals on exactly the traffic the customer excluded.

## Acceptance criteria

- `insert_spans_batch` and `insert_traces_batch` write `environment` (dedupe with #1408's fix).
- An e2e test asserts a set environment round-trips transform → insert → `_get_trace_summaries` → trigger evaluation: `= X` fires only on matching traces; `!= X` does not fire on the excluded environment.
- A regression test pins `_eval_condition`'s missing-field behavior so the insert fix doesn't change intended absent-field semantics.

dark-sorceror:
## Summary

Fixes #1567 

Environment trigger conditions never worked in production — and `environment` is the only field the trigger editor offers, so trigger-based filtering as a feature was silently broken in both directions:

- `environment = production` **never fired**, even on production traces.
- `environment != staging` **fired on everything**, including the staging traces it was configured to exclude.

The SDK's `traceroot.environment` attribute was correctly extracted by the OTLP transform, but `insert_spans_batch` / `insert_traces_batch` omitted `environment` from their column lists, so the value was dropped and the ClickHouse column was always NULL. The detector worker's trace summary then read that NULL, and `_eval_condition`'s missing-field semantics (`actual is None` → `op == "!="`) inverted every condition. No error surfaced anywhere.

This PR writes the column through and hardens the extraction so a mis-typed attribute can't fail a whole insert batch.

## How it works

```
SDK exports span with traceroot.environment
  → otel_transform extracts it (string values only — new guard)
    → insert_spans_batch / insert_traces_batch write the environment column (new)
      → _get_trace_summaries reads a real value instead of NULL
        → _eval_condition compares against the customer's condition
          → "= production" fires only on production; "!= staging" skips staging
```

Key invariants:
- A mis-typed `traceroot.environment` (int/bool/list from a buggy client) degrades to "not set" (NULL) instead of reaching the `Nullable(String)` insert and failing the entire batch.
- `_eval_condition`'s missing-field semantics are unchanged and now pinned by regression tests.

### Backend (Python)
- `backend/db/clickhouse/client.py` — `environment` added to the row builder and `column_names` of both `insert_spans_batch` and `insert_traces_batch` (the columns have existed since ClickHouse migration 003; they were simply never written).
- `backend/worker/otel_transform.py` — the environment guards now accept only string values (`isinstance(environment, str)` instead of `is not None`), on both the span record and the root-span trace record.

### Tests
- `tests/db/test_client.py` — column/row-index assertions for `environment` on both insert paths, plus absent-key → NULL defaults.
- `tests/worker/test_environment_pipeline.py` (new) — round-trip through the real transform and real insert functions for spans and traces, threading the stored value into trigger evaluation; plus mis-typed-attribute cases asserting the insert row carries NULL.
- `tests/worker/test_otel_transform.py` — type-guard cases (string kept; int/bool/list dropped on span and trace records).
- `tests/worker/test_detector_tasks.py` — regression tests pinning `_eval_condition`'s missing-field semantics and populated-value behavior in both directions.

## Test plan
- [x] Full backend suite: 799 passed, 1 pre-existing unrelated skip
- [x] E2E against a live stack through the real pipeline (transform → insert → summary → trigger → enqueue): environment round-trips to ClickHouse; `environment = production` fires on a production trace and stays silent on staging; `environment != staging` no longer fires on staging traces; empty-conditions control detector unaffected
- [x] TDD throughout: every new test observed failing against the unfixed code first
- [x] ruff check + format clean on all changed files

---
## Summary by cubic
Fixes environment-based detector triggers by persisting the `environment` field to ClickHouse on span and trace inserts and type-guarding the OTLP transform to accept only string values. Conditions like `environment = production` and `environment != staging` now evaluate correctly without breaking inserts.

- **Bug Fixes**
  - ClickHouse: write `environment` in `insert_traces_batch` and `insert_spans_batch` so it’s no longer always NULL.
  - OTLP transform: keep `traceroot.environment` only if it’s a string; non-strings are dropped to NULL to avoid batch failures.
  - Trigger behavior: equality/inequality checks work as expected; missing-field semantics remain unchanged and are covered by regression tests.

<sup>Written for commit a6c42d463b1a0fdce1149a99072ab676a4bb7e95. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1569?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/db/clickhouse/client.py
+++ b/backend/db/clickhouse/client.py
@@ -59,6 +59,7 @@ class ClickHouseClient:
                     t.get("metadata"),
                     now,  # ch_create_time
                     now,  # ch_update_time
+                    t.get("environment"),
                 ]
             )
 
@@ -79,6 +80,7 @@ class ClickHouseClient:
                 "metadata",
                 "ch_create_time",
                 "ch_update_time",
+                "environment",
             ],
         )
 
@@ -116,6 +118,7 @@ class ClickHouseClient:
                     s.get("git_source_function"),
                     now,  # ch_create_time
                     now,  # ch_update_time
+                    s.get("environment"),
                 ]
             )
 
@@ -147,6 +150,7 @@ class ClickHouseClient:
                 "git_source_function",
                 "ch_create_time",
                 "ch_update_time",
+                "environment",
             ],
         )
 
--- a/backend/worker/otel_transform.py
+++ b/backend/worker/otel_transform.py
@@ -385,7 +385,7 @@ def transform_otel_to_clickhouse(
                     "span_kind": span_kind,
                     "status": SpanStatus.OK,
                 }
-                if environment is not None:
+                if isinstance(environment, str):
                     span_record["environment"] = environment
 
                 # Extract git source fields for span
@@ -771,7 +771,7 @@ def transform_otel_to_clickhouse(
                         }
                     )
 
-                    if environment is not None:
+                    if isinstance(environment, str):
                         traces[trace_id]["environment"] = environment
                     if trace_git_attrs[trace_id]["git_ref"] is not None:
                         traces[trace_id]["git_ref"] = trace_git_attrs[trace_id]["git_ref"]
--- a/tests/db/test_client.py
+++ b/tests/db/test_client.py
@@ -27,6 +27,7 @@ class TestInsertTracesBatch:
                 "git_repo": "owner/repo",
                 "input": "hello",
                 "output": "world",
+                "environment": "production",
             }
         ]
         client.insert_traces_batch(traces)
@@ -53,7 +54,9 @@ class TestInsertTracesBatch:
         # ch_create_time and ch_update_time are auto-set
         assert isinstance(row[11], datetime)
         assert isinstance(row[12], datetime)
-        assert len(columns) == 13
+        assert len(columns) == 14
+        assert "environment" in columns
+        assert row[columns.index("environment")] == "production"
 
     def test_empty_batch_no_insert(self):
         """Empty list -> no _client.insert() call."""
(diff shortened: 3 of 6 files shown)


XinweiHe (approved):
LGTM

Issue #1581, opened 2026-07-16
Cost is $0 for OpenRouter's anthropic/claude-opus-4.8-fast — no fast rate card, and dot-notation versions don't match


XinweiHe:
## Problem

A customer routing through OpenRouter sends `anthropic/claude-opus-4.8-fast`. It matches no entry in `standard-model-prices.json`, so `get_model_price` returns `None` and the span records no cost.

Their non-fast traffic is unpriced too: OpenRouter spells versions with **dots** (`anthropic/claude-opus-4.8`), and our Anthropic patterns only accept **dashes** (`claude-opus-4-8`). All 7 current-generation Anthropic slugs are affected, so the whole account reads $0.

## Repro

```py
get_model_price("anthropic/claude-opus-4.8-fast")  # -> None
get_model_price("anthropic/claude-opus-4.8")       # -> None
get_model_price("anthropic/claude-sonnet-4.6")     # -> None
get_model_price("anthropic/claude-opus-4-8")       # -> matches (dash form works)
```

## Root cause

1. **No rate card for fast mode.** Fast is a separately-priced tier, not a cosmetic suffix.
2. **Patterns require dash version spelling**; gateways use dots.

## Fast mode rates (verified against Anthropic's pricing page)

| Model | Input | Output | Note |
|---|---|---|---|
| Opus 4.8 | $10 / MTok | $50 / MTok | 2x standard |
| Opus 4.7 | $30 / MTok | $150 / MTok | 6x standard; removed 2026-07-24 |
| Opus 4.6 | — | — | Not fast-capable; bills standard since 2026-06-29 |

Caching multipliers apply on top of the **fast** base, so Opus 4.8 fast is $1 cache read / $12.50 5m write / $20 1h write. (OpenRouter's provider table independently confirms $1.00 cache read.)

> [!WARNING]
> The fast premium **dropped from 6x to 2x** at Opus 4.8. Do not copy 4.7's rates forward — that would overbill 3x. Same failure mode as the Opus 4.5 rate incident.

## Implementation note

Fast must be a **distinct entry**, not a relaxed pattern on the standard one. If a pattern swallows `-fast` into the standard entry, 2x traffic bills at 1x — a silent undercharge, which is worse than the current $0 (a zero cost is visible; a 50%-off invoice is not).

[...]

PR #1587, opened 2026-07-17
fix(pricing): match Anthropic dot-version names and add fast rate cards

hassaanch23:
## What

Two pricing gaps for Anthropic models sent through gateways, both in
`standard-model-prices.json`:

1. **Dot-notation versions** — gateways (e.g. OpenRouter) spell the version with
   a dot (`anthropic/claude-opus-4.8`), but the patterns only accept the dash
   form (`claude-opus-4-8`). So `get_model_price` returned `None` and the span
   recorded **$0** — affecting every current-generation Anthropic slug.
2. **No fast rate card** — fast is a separately-priced tier, not a cosmetic
   suffix, so `…-fast` traffic was also unpriced.

## Fix

- **Dot versions:** broaden the version separator to `[-.]` on the 9 affected
  Anthropic patterns (`claude-opus-4[-.]8`, …). This is surgical, per-entry — a
  blanket dot→dash normalization would corrupt `gpt-3.5-turbo`'s legitimate dot
  and Bedrock's separator dots (`us.anthropic.…`), so those are left untouched.
- **Fast tier:** add distinct rate cards `claude-opus-4-8-fast` (**2×** standard,
  $10/$50 per MTok) and `claude-opus-4-7-fast` (**6×**), each requiring the
  `-fast` suffix. Fast is a *dedicated* entry, never a relaxed pattern on the
  standard one: if a standard pattern swallowed `-fast`, 2× usage would bill at
  1× — a silent undercharge, worse than the visible $0. (Opus 4.6 is not
  fast-capable, so it gets no fast card and continues to bill standard.)

The JSON is the shared source for the Python worker (`worker/tokens/pricing`) and
the TS lookup (`packages/core/model-pricing`), which apply identical regex — so
one data fix covers both, and `sync-standard-prices.ts` upserts it on deploy.

## How validated

`tests/worker/tokens/test_pricing.py` (reads the real JSON via the `real_cache`
fixture — no DB, no LLM):

- **Dot versions** added to the Bedrock/Vertex matrix: `anthropic/claude-opus-4.8`
  → `claude-opus-4-8`, etc.; dash/Bedrock/Vertex forms still match.
- **New `TestAnthropicFastTier`:** each fast id resolves to its *own* card;
  standard traffic never hits a fast card; Opus 4.8 fast rates are exactly 2×
  (pinned to $10/$50) and Opus 4.7 fast is 6× (not copied from 4.8).

Results: `pytest tests/worker/tokens/test_pricing.py` → 200 passed; full backend
suite → 869 passed; TS `model-pricing.test.ts` → 11 passed; `ruff` clean.

## Scope

Covers the model-string case (OpenRouter puts the variant in the slug). The
Vercel AI Gateway / direct-API `speed: "fast"` paths can't be distinguished by
model string and have no affected customers today — out of scope, as the issue
notes.

Refs #1581

---
## Summary by cubic
Fixes zero-cost billing for Anthropic gateway slugs by matching dot-version model names, and adds dedicated fast-tier pricing for Opus 4.8 and 4.7 across both backends. Refs #1581.

- **Bug Fixes**
  - Broadened version separators to `[-.]` for nine Anthropic patterns (e.g., `claude-opus-4[-.]8`) so `anthropic/claude-opus-4.8` and similar slugs match and bill correctly; change is per-entry to avoid affecting other providers.

- **New Features**
  - Added fast-tier rate cards: `claude-opus-4-8-fast` (2× standard; $10/$50 per MTok) and `claude-opus-4-7-fast` (6×). `-fast` slugs resolve only to these cards; Opus 4.6 has no fast tier.

<sup>Written for commit 975e44921bba2a9778b3f252210c8a9ad304a776. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1587?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/packages/core/src/standard-model-prices.json
+++ b/frontend/packages/core/src/standard-model-prices.json
@@ -13,9 +13,22 @@
     }
   },
   {
+    "id": "tr-claude-opus-4-8-fast",
+    "modelName": "claude-opus-4-8-fast",
+    "matchPattern": "(?i)^(anthropic\\/)?claude-opus-4[-.]8-fast$",
+    "provider": "anthropic",
+    "prices": {
+      "input": 0.00001,
+      "output": 0.00005,
+      "cacheRead": 0.000001,
+      "cacheWrite": 0.0000125,
+      "cacheWrite1h": 0.00002
+    }
+  },
+  {
     "id": "tr-claude-opus-4-8",
     "modelName": "claude-opus-4-8",
-    "matchPattern": "(?i)^(anthropic\\/)?(claude-opus-4-8(\\[1m\\])?(-[\\d-]+)?|(eu\\.|us\\.|apac\\.|global\\.)?anthropic\\.claude-opus-4-8(-[\\d-]+)?(-v\\d+:\\d+)?|claude-4-8-opus(@[\\d-]+)?)$",
+    "matchPattern": "(?i)^(anthropic\\/)?(claude-opus-4[-.]8(\\[1m\\])?(-[\\d-]+)?|(eu\\.|us\\.|apac\\.|global\\.)?anthropic\\.claude-opus-4[-.]8(-[\\d-]+)?(-v\\d+:\\d+)?|claude-4[-.]8-opus(@[\\d-]+)?)$",
     "provider": "anthropic",
     "prices": {
       "input": 0.000005,
@@ -26,9 +39,22 @@
     }
   },
   {
+    "id": "tr-claude-opus-4-7-fast",
+    "modelName": "claude-opus-4-7-fast",
+    "matchPattern": "(?i)^(anthropic\\/)?claude-opus-4[-.]7-fast$",
+    "provider": "anthropic",
+    "prices": {
+      "input": 0.00003,
+      "output": 0.00015,
+      "cacheRead": 0.000003,
+      "cacheWrite": 0.0000375,
+      "cacheWrite1h": 0.00006
+    }
+  },
+  {
     "id": "tr-claude-opus-4-7",
     "modelName": "claude-opus-4-7",
-    "matchPattern": "(?i)^(anthropic\\/)?(claude-opus-4-7(\\[1m\\])?(-[\\d-]+)?|(eu\\.|us\\.|apac\\.|global\\.)?anthropic\\.claude-opus-4-7(-[\\d-]+)?(-v\\d+:\\d+)?|claude-4-7-opus(@[\\d-]+)?)$",
+    "matchPattern": "(?i)^(anthropic\\/)?(claude-opus-4[-.]7(\\[1m\\])?(-[\\d-]+)?|(eu\\.|us\\.|apac\\.|global\\.)?anthropic\\.claude-opus-4[-.]7(-[\\d-]+)?(-v\\d+:\\d+)?|claude-4[-.]7-opus(@[\\d-]+)?)$",
     "provider": "anthropic",
     "prices": {
       "input": 0.000005,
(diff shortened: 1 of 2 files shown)


ka1kqi:
Hi @hassaanch23 — thanks for this, and apologies for the collision: #1593 was opened against the same issue (#1581) a few hours after yours, from work that had been in progress internally. Rather than have two conflicting PRs racing on the same hunks, we've folded the two genuine improvements your PR had over ours **into #1593 with co-author credit to you on the commit** (`7b4e4f92`):

- the **`claude-opus-4-7-fast` rate card** ($30/$150 = 6x, cache derived from the fast base) — your call to include it was right; it prices historical 4.7-fast spans from before the July 24 removal, and we added your 6x-not-2x rate lock as a test
- **legacy 3.x dot tolerance** (`claude-3.5-sonnet` / `claude-3.5-haiku`) — a real gap in our version
- your `test_opus_4_6_has_no_fast_card` guard

Two review notes on this PR's remaining delta, in case useful:

1. The PR description says the TS lookup "applies identical regex — so one data fix covers both." We runtime-tested this: it doesn't hold — `new RegExp("(?i)...")` throws in JS and the catch silently skips the entry, so the regex path in `packages/core/model-pricing/lookup.ts` has never matched anything (only exact `modelName` hits work there). It's harmless today because nothing outside `packages/core` consumes that lookup, but it's worth knowing the data fix is effectively Python-worker-only.
2. The dot edits here also loosen the Bedrock (`anthropic.claude-opus-4[-.]8`) and Vertex (`claude-4[-.]8-opus`) alternatives — no billing risk, but those platforms genuinely emit dashes, so #1593 kept them dash-only.

With those two pieces incorporated, #1593 covers everything here — we'd suggest closing this one once it merges, and thank you again: the 4.7 card and 3.x slugs would have been follow-up bugs without your PR.

hassaanch23:
Thanks @ka1kqi — no worries at all on the collision, and I appreciate you folding the 4.7 card, the 3.x slugs, and the 4.6 guard into #1593 with co-author credit. Glad those were useful.

Both notes are fair:

1. Good catch on the TS lookup — I over-stated "covers both." I confirmed `new RegExp("(?i)…")` throws `Invalid group` in V8, so `lookup.ts`'s regex fallback has silently never matched; only exact `modelName` hits work there, so this is worker-only as you said. That's a real (if currently harmless) latent bug — happy to send a tiny follow-up that strips the leading `(?i)` before `new RegExp(…, "i")` so the TS regex path actually works and the two implementations reach true parity, if useful.
2. Agreed on keeping Bedrock/Vertex dash-only — those emit dashes, so #1593's tighter version is the right call.

Happy to close this once #1593 merges. Thanks again — good one to dig into.

hassaanch23:
Closing this out — #1593 landed the pricing-catalog half of this on Jul 22 (thanks again for folding in the 4.7 card, the 3.x slugs, and the 4.6 guard with co-author credit, @ka1kqi), so there's nothing left here that isn't already on `main`.

The one piece that wasn't covered is the `lookup.ts` regex issue we discussed at the end of this thread — I've sent that as its own focused PR: #1622. It confirms `new RegExp("(?i)…")` throws `Invalid group` in V8 and is swallowed by the `catch` below it, so the TS regex fallback has silently never matched; the fix strips the leading `(?i)` and keeps the `"i"` flag. Two regression tests, verified failing before and passing after.

Appreciate the review on this one.

PR #1593, opened 2026-07-17
fix(pricing): price fast-mode and dot-notation Anthropic model ids

### LINKED ISSUE #1581: Cost is $0 for OpenRouter's anthropic/claude-opus-4.8-fast — no fast rate card, and dot-notation versions don't match

ka1kqi:
Fixes #1581

## Problem

A customer's `anthropic/claude-opus-4.8-fast` spans recorded no cost. Two independent causes:

1. **No fast rate card.** Fast mode is a real, separately priced tier — $10/$50 per MTok on Opus 4.8, exactly 2x standard — that some gateways expose as a distinct `-fast` model slug. No entry matched it.
2. **Dot notation.** Gateways spell versions with dots (`claude-opus-4.8`); our first-party patterns only accepted dashes. This zeroed costs for every current-gen Anthropic model for that customer, not just fast traffic.

## Change

- New `claude-opus-4-8-fast` entry: input/output/cacheRead/cacheWrite/cacheWrite1h all exactly 2x the standard Opus 4.8 entry, matching Anthropic's published fast-mode rates (cache multipliers apply on top of the fast base, per their pricing docs).
- Dot-tolerant first-party patterns (`4[.-]8` etc.) on the 7 current-gen Anthropic entries. Bedrock/Vertex pattern branches keep dashes (those platforms genuinely use dashes).

Also adopted from #1587 by @hassaanch23 (co-author credit on the commit), who independently built the same fix and caught two things this PR originally lacked:
- An **Opus 4.7 fast** entry ($30/$150 = 6x standard, cache derived from the fast base) — prices historical fast spans from before Anthropic's 2026-07-24 removal of 4.7 fast. A dedicated test locks it at 6x so neither tier's rate can be copy-forwarded to the other.
- **Legacy 3.x dot tolerance** (`claude-3.5-sonnet` / `claude-3.5-haiku` are real old gateway slugs), first-party alternative only.

Deliberately excluded: Opus 4.6 fast (bills standard since 2026-06-29; adding it would over-charge — now enforced by a `test_opus_4_6_has_no_fast_card` guard, also adopted from #1587).

Complementary to #1566 (gateway prefix stripping, #1556) and independent of it: that PR normalizes `openrouter/anthropic/...` prefixes, this one adds the fast rate card and dot tolerance those normalized strings still need.

Note on rollout: prices reach the backend via the startup sync (`syncStandardPrices`) into Postgres, so the new entry takes effect when the worker/agent services next start — no manual migration needed.

## Tests

- `TestClaudeFastAndDotNotationIds`: 12 model-id cases — including all four repro strings from #1581 (`anthropic/claude-opus-4.8-fast`, `anthropic/claude-opus-4.8`, `anthropic/claude-sonnet-4.6`, and the dash form) — each asserting resolution to the exact expected entry **and** that exactly one pattern matches (guards the undercharge trap where a loosened pattern could swallow `-fast` into the standard entry), plus a lock that every fast price stays exactly 2x standard.
- Full suite: `uv run pytest tests/worker/tokens/ -q` — 241 passed.

## Runtime verification

Exercised the real deploy pipeline against a throwaway database: prisma migrations → actual `syncStandardPrices()` run (88 models synced, fast entry landed with all 5 price rows) → Python worker's DB-backed lookup. All customer strings priced correctly, `calculate_cost` fast/standard ratio exactly 2.000, and adversarial probes held (`-fastest` no-match, gateway-prefixed strings left to the prefix-stripping PR, Bedrock/Vertex ids unchanged).

---
## Summary by cubic
Fixes zero-cost Anthropic spans by pricing fast mode and accepting dot-notation model ids. Also logs a warning once per unmatched model id so unpriced spans are visible.

- **Bug Fixes**
  - Added `claude-opus-4-8-fast` at exactly 2x `claude-opus-4-8` for input/output/cache.
  - Added `claude-opus-4-7-fast` at 6x `claude-opus-4-7` for historical fast spans.
  - First‑party Anthropic patterns accept dot or dash in versions (e.g., `4[.-]8`, `3[.-]5`); Bedrock/Vertex branches stay dashed.
  - Warn once per unmatched model id (bounded) when no pricing entry matches.

- **Migration**
  - No manual steps. Prices sync on startup via `syncStandardPrices`; restart worker/agent services.

<sup>Written for commit 7b4e4f92630f7e746ba271b3ace5a3d48b69aeba. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1593?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/packages/core/src/standard-model-prices.json
+++ b/frontend/packages/core/src/standard-model-prices.json
@@ -39,6 +39,19 @@
     }
   },
   {
+    "id": "tr-claude-opus-4-7-fast",
+    "modelName": "claude-opus-4-7-fast",
+    "matchPattern": "(?i)^(anthropic\\/)?claude-opus-4[.-]7-fast$",
+    "provider": "anthropic",
+    "prices": {
+      "input": 0.00003,
+      "output": 0.00015,
+      "cacheRead": 0.000003,
+      "cacheWrite": 0.0000375,
+      "cacheWrite1h": 0.00006
+    }
+  },
+  {
     "id": "tr-claude-opus-4-7",
     "modelName": "claude-opus-4-7",
     "matchPattern": "(?i)^(anthropic\\/)?(claude-opus-4[.-]7(\\[1m\\])?(-[\\d-]+)?|(eu\\.|us\\.|apac\\.|global\\.)?anthropic\\.claude-opus-4-7(-[\\d-]+)?(-v\\d+:\\d+)?|claude-4-7-opus(@[\\d-]+)?)$",
@@ -672,7 +685,7 @@
   {
     "id": "tr-claude-3-5-sonnet",
     "modelName": "claude-3-5-sonnet",
-    "matchPattern": "(?i)^(anthropic\\/)?(claude-3-5-sonnet(-[\\d-]+)?|(eu\\.|us\\.|apac\\.|global\\.)?anthropic\\.claude-3-5-sonnet(-[\\d-]+)?(-v\\d+:\\d+)?|claude-3-5-sonnet(@[\\d-]+)?)$",
+    "matchPattern": "(?i)^(anthropic\\/)?(claude-3[.-]5-sonnet(-[\\d-]+)?|(eu\\.|us\\.|apac\\.|global\\.)?anthropic\\.claude-3-5-sonnet(-[\\d-]+)?(-v\\d+:\\d+)?|claude-3-5-sonnet(@[\\d-]+)?)$",
     "provider": "anthropic",
     "prices": {
       "input": 0.000003,
@@ -685,7 +698,7 @@
   {
     "id": "tr-claude-3-5-haiku",
     "modelName": "claude-3-5-haiku",
-    "matchPattern": "(?i)^(anthropic\\/)?(claude-3-5-haiku(-[\\d-]+)?|(eu\\.|us\\.|apac\\.|global\\.)?anthropic\\.claude-3-5-haiku(-[\\d-]+)?(-v\\d+:\\d+)?|claude-3-5-haiku(@[\\d-]+)?)$",
+    "matchPattern": "(?i)^(anthropic\\/)?(claude-3[.-]5-haiku(-[\\d-]+)?|(eu\\.|us\\.|apac\\.|global\\.)?anthropic\\.claude-3-5-haiku(-[\\d-]+)?(-v\\d+:\\d+)?|claude-3-5-haiku(@[\\d-]+)?)$",
     "provider": "anthropic",
     "prices": {
       "input": 0.0000008,
--- a/frontend/packages/core/src/standard-model-prices.json
+++ b/frontend/packages/core/src/standard-model-prices.json
@@ -13,9 +13,22 @@
     }
   },
   {
+    "id": "tr-claude-opus-4-8-fast",
+    "modelName": "claude-opus-4-8-fast",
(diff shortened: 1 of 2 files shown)


backend/worker/tokens/pricing.py
```diff
@@ -27,6 +27,12 @@
 
 _cache: list[dict] | None = None
 
+# Model ids already warned about as unmatched, so each id logs once per
```
XinweiHe: let's remove this logic

XinweiHe (approved):

PR #1594, opened 2026-07-18
feat(ui): add DOMAIN_ICONS registry, consolidate scattered domain-concept icons

### LINKED ISSUE #1517: Add a single source of truth for domain-concept icons (tokens, cost, model, user, session, …)
### Problem

Domain concepts (tokens, cost, model, user, session, trace, …) are each drawn with a lucide icon, but there is no single source of truth mapping concept → icon. Icons are imported ad-hoc per file, with three localized maps that don't know about each other:

- `FIELD_ICONS` in `frontend/ui/src/features/filters/filter-builder.tsx` — filter fields only; its own comment says it "mirror[s] the trace detail / list UI for consistency", i.e. consistency is maintained by hand.
- `getSpanKindIcon()` + `SPAN_KIND_COLORS` in `frontend/ui/src/features/traces/components/SpanKindIcon.tsx` — span kinds only.
- Nav-tab `icon:` arrays duplicated inline across the traces/sessions/users pages, `settings-layout.tsx`, and the detector detail page.

Because every new surface (stat card, filter field, nav item, chip) forces the author to rediscover the convention, the UI has already drifted:

| Concept | Conflicting representations |
|---|---|
| Model / LLM | `Box` (filter field) · `Bot` ("Model Providers" settings tab) · `Sparkle` (`llm` span kind) · no icon at all (model pill in `SpanInfoPanel.tsx`) |
| User | `Users` everywhere except the sidebar, which uses `UserRoundSearch` |
| Trace | `Workflow` everywhere except the `trace_id` filter field, which uses `Hash` |
| Project | `FolderKanban` (projects page) vs `LayoutGrid` (sidebar, onboarding) |
| Bot-family glyphs | `Bot` means both "Model Providers" and the `agent` span kind, while `BotMessageSquare` means the AI assistant and `Sparkle` doubles as the `llm` span kind and generic decoration |

Consistent-but-fragile today: tokens = `CircleStop`, cost = `CircleDollarSign` (+ `$` input prefix), session = `Layers`, latency = `Clock`, detector = `Eye`. Nothing stops the next surface from breaking these.

### Proposed Solution

Add a semantic icon registry — one map from domain concept to lucide icon — and make every surface consume it:

```ts
// frontend/ui/src/components/icons/domain-icons.ts
import type { LucideIcon } from "lucide-react";

export const DOMAIN_ICONS = {
  tokens: CircleStop,        // decision: keep, or switch to Coins (see below)
  cost: CircleDollarSign,
  model: Box,
  agent: Bot,
  llm: Sparkle,
  assistant: BotMessageSquare,
  user: Users,
  session: Layers,
  trace: Workflow,
  id: Hash,
  latency: Clock,
  error: AlertCircle,
  project: FolderKanban,
  detector: Eye,
  environment: Globe,
} satisfies Record<string, LucideIcon>;
```

Then:

1. Rewire `FIELD_ICONS`, `getSpanKindIcon()`, the duplicated nav-tab arrays, `TokenChip`/`CostChip`, and the sidebar to import from `DOMAIN_ICONS` instead of importing lucide icons directly.
2. Fix the drifted surfaces to the canonical choice (proposed to minimize churn — each picks the icon already used on the majority of surfaces):
   - **model** = `Box` → change the "Model Providers" settings tab from `Bot`; `Bot` then exclusively means the `agent` span kind.
   - **user** = `Users` → change the sidebar from `UserRoundSearch`.
   - **trace** = `Workflow`; the `trace_id` filter keeps `Hash` but as the explicit `id` concept (IDs are a different concept from the trace entity).
   - **project** = `FolderKanban` → change the sidebar/onboarding from `LayoutGrid`.
3. Keep `PROVIDER_ICONS` (`components/icons/provider-icons.tsx`) as-is — it maps provider *brands*, not concepts — but colocate the new registry next to it.

Open decision: **tokens**. `CircleStop` is used consistently but reads as a stop sign, not a token count. Switching to `Coins` (available in our lucide version) would be more intuitive but is a user-visible change across trace list, span tree, timeline, and filters. The registry makes either choice a one-line change.

### Alternatives Considered

- **Do nothing / fix drift case-by-case** — the five inconsistencies above show manual mirroring doesn't hold; each fix without a registry just resets the clock.
- **ESLint rule restricting direct `lucide-react` imports for domain concepts** — enforceable but heavy; a registry plus review convention should be enough at current codebase size. Could be layered on later if drift recurs.
- **Folding span-kind colors into the same registry** — `SPAN_KIND_COLORS` (surface/text Tailwind classes) could live alongside the icons, but it's span-tree-specific today; keep it in `SpanKindIcon.tsx` sourcing its glyphs from `DOMAIN_ICONS`, and only promote the colors if another surface needs them.

### Additional Context

Suggested scope of the migration (all under `frontend/ui/src`): `features/filters/filter-builder.tsx`, `features/traces/components/{SpanKindIcon,TokenChip,CostChip,SpanTreeView,SpanTimelineView,SpanInfoPanel,SessionDetailPanel,TraceViewerPanel}.tsx`, `components/layout/sidebar.tsx`, `features/settings/settings-layout.tsx`, the nav-tab arrays in `app/projects/[projectId]/{traces,sessions,users}/page.tsx`, and `app/workspaces/[workspaceId]/projects/page.tsx`. Pure refactor except the four canonical-icon fixes (and optionally the token glyph), so no behavior/API changes.

Satyam296:
## Summary

Fixes #1517.

Domain concepts (tokens, cost, model, user, session, trace, …) were each drawn with a lucide icon picked ad-hoc per file. Adds a single source of truth — `frontend/ui/src/components/icons/domain-icons.ts` — and migrates every surface with an actual duplicated/drifted icon import to source from it:

- `filter-controls.tsx`'s `FIELD_ICONS`
- the dashboard widget builder's `field-icons.ts` — a **fourth** ad-hoc icon map that appeared after the dashboards stack merged, since this issue was filed
- `SpanKindIcon`'s `getSpanKindIcon`
- the token/cost chips and their direct-import duplicates in `SpanTreeView`/`SpanTimelineView`
- the `BotMessageSquare` "AI assistant" icon, duplicated across `app-layout`, `SessionDetailPanel`, `TraceViewerPanel`
- the sidebar's trace/detector/dashboard/workspace nav icons
- the settings tabs
- the triplicated traces/sessions/users nav-tab arrays

**Canonical fixes for the real collisions:**
- Model Providers settings tab: `Bot` → `Box` (model), so `Bot` exclusively means the `agent` span kind afterward.
- `SpanInfoPanel`'s `model_name` pill had no icon at all — now shows the model glyph like every other model-bearing surface.
- Onboarding's "Project" step drew `Layers` (the registry's `session` glyph) and the projects empty-state drew a raw `FolderKanban` — both now source from `DOMAIN_ICONS.project`.

**Also fixed along the way** (same concepts, direct imports found while migrating each file): the latency/user/session/error icons duplicated in `SpanInfoPanel`, `SessionDetailPanel`, and the traces/sessions/users empty states.

**Corrections to the issue's originally proposed fixes**, verified against the current codebase:
- "Project" (`FolderKanban`) vs "Workspace" (`LayoutGrid`, sidebar/onboarding) are not the same drifted concept — they're genuinely distinct entities in this app. Both are now separate registry entries; the sidebar's Workspace icon is unchanged.
- Sidebar's `UserRoundSearch` (admin impersonation banner) is a distinct, single-use meaning, not the general "user" concept — left as a local icon, not folded into the registry.
- `tokens` keeps `CircleStop` (the issue's own "open decision") — this PR is a pure consolidation, not a visual redesign.

Pure refactor: no behavior change except the three canonical fixes above.

## Test plan

- [x] `tsc --noEmit` clean on all touched files (2 pre-existing, unrelated errors elsewhere untouched by this PR)
- [x] `eslint` clean on all touched files (1 pre-existing, unrelated warning)
- [x] Full frontend suite: 891/893 pass — the 2 failures are pre-existing, environment-specific flakes (a locale-dependent formatting test, a Slack OAuth timeout), unrelated to anything this PR touches
- [x] Manually traced every migrated call site's diff to confirm the rendered icon is pixel-identical except the three stated canonical fixes

---
## Summary by cubic
Add `DOMAIN_ICONS` as the single source of truth for domain icons and replace ad‑hoc `lucide-react` imports across the UI for consistent visuals. Adds a neutral `fallback` icon (decoupled from `model`), fixes icon drift in onboarding and workspace/project empty states, and enforces the registry with an ESLint rule; fixes #1517.

- **Refactors**
  - Added `frontend/ui/src/components/icons/domain-icons.ts` exporting `DOMAIN_ICONS` (tokens, cost, model, user, session, trace, detector, etc.) plus a neutral `fallback`.
  - Replaced direct `lucide-react` imports in filters, dashboards, span views (tree/timeline/info), chips, sidebar, settings, app layout, onboarding, detectors (panel/page), and traces/sessions/users tabs and empty states.
  - Canonical fixes: Model Providers tab uses `DOMAIN_ICONS.model` (`Box`); `model_name` pill shows the model icon; onboarding uses `DOMAIN_ICONS.workspace`; project/workspace empty states use `DOMAIN_ICONS.project`.
  - Added ESLint guard in `frontend/eslint.config.mjs` banning direct `lucide-react` glyphs the registry owns in `ui/src` (registry/tests exempt); pattern covers `Lucide*` and `*Icon` aliases, and the relative-import `.ts` extension rule is hoisted for Turbopack.

- **Tests**
  - Added and tightened coverage: `getSpanKindIcon`, SpanTreeView/SpanTimeline token/cost badges (assert glyphs render), settings tabs, nav tabs, and empty states (users/sessions/detectors), `model_name` pill icon, and traces page `user_id` badge.
  - Updated mocks and switched to current `lucide-react` names; diff-cover 100% on this PR.

<sup>Written for commit 2e781d6e571af1d48527f2a32b591a9dada60c59. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1594?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/ui/src/app/onboarding/page.tsx
+++ b/frontend/ui/src/app/onboarding/page.tsx
@@ -12,7 +12,8 @@ import { Input } from "@/components/ui/input";
 import { Card, CardContent } from "@/components/ui/card";
 import { useLayout } from "@/components/layout/app-layout";
 import { createWorkspace, createProject } from "@/lib/api";
-import { LayoutGrid, Layers, Check } from "lucide-react";
+import { LayoutGrid, Check } from "lucide-react";
+import { DOMAIN_ICONS } from "@/components/icons/domain-icons";
 import { AddDetectorsStep } from "@/features/detectors/components/add-detectors-step";
 
 const onboardingSchema = z.object({
@@ -223,7 +224,7 @@ function OnboardingContent() {
                   <div className="flex items-start gap-3">
                     <div className="flex flex-col items-center">
                       <div className="flex h-8 w-8 shrink-0 items-center justify-center rounded bg-muted">
-                        <Layers className="h-4 w-4 text-muted-foreground" />
+                        <DOMAIN_ICONS.project className="h-4 w-4 text-muted-foreground" />
                       </div>
                     </div>
                     <div className="flex-1 pb-2 pt-0.5">
--- a/frontend/ui/src/app/onboarding/page.tsx
+++ b/frontend/ui/src/app/onboarding/page.tsx
@@ -12,7 +12,7 @@ import { Input } from "@/components/ui/input";
 import { Card, CardContent } from "@/components/ui/card";
 import { useLayout } from "@/components/layout/app-layout";
 import { createWorkspace, createProject } from "@/lib/api";
-import { LayoutGrid, Check } from "lucide-react";
+import { Check } from "lucide-react";
 import { DOMAIN_ICONS } from "@/components/icons/domain-icons";
 import { AddDetectorsStep } from "@/features/detectors/components/add-detectors-step";
 
@@ -189,7 +189,7 @@ function OnboardingContent() {
                         {isProjectOnlyMode ? (
                           <Check className="h-4 w-4 text-primary" />
                         ) : (
-                          <LayoutGrid className="h-4 w-4 text-primary" />
+                          <DOMAIN_ICONS.workspace className="h-4 w-4 text-primary" />
                         )}
                       </div>
                       {/* Vertical connector line */}
--- a/frontend/ui/src/components/icons/domain-icons.ts
+++ b/frontend/ui/src/components/icons/domain-icons.ts
@@ -27,7 +27,7 @@ import {
  * Single source of truth for domain-concept icons. Every surface that draws
  * one of these concepts (filter fields, span kinds, nav tabs, chips, …)
  * should import from here instead of picking a lucide icon directly — that's
- * what let "model" drift to four different glyphs across the app (#1517).
+ * what let "model" drift to four different glyphs across the app.
  *
  * `project` and `workspace` are deliberately separate entries: they read as
  * the same concept in the original issue text, but they're distinct concepts
@@ -65,6 +65,6 @@ export const DOMAIN_ICONS = {
   // Neutral "unknown field" fallback for filter/widget dropdowns. Kept
   // decoupled from `model` even though both currently render as Box — if the
   // model glyph ever changes, unmapped fields shouldn't silently change with
-  // it (ka1kqi, #1594).
+  // it.
   fallback: Box,
 } satisfies Record<string, LucideIcon>;
--- a/frontend/ui/src/components/icons/domain-icons.ts
+++ b/frontend/ui/src/components/icons/domain-icons.ts
@@ -1,0 +1,65 @@
+import type { LucideIcon } from "lucide-react";
+import {
+  AlertCircle,
+  ArrowRight,
+  Bot,
+  BotMessageSquare,
+  Box,
+  CircleCheck,
+  CircleDollarSign,
+  CircleStop,
+  Clock,
+  Eye,
+  FolderKanban,
+  Globe,
+  Hash,
+  LayoutDashboard,
+  LayoutGrid,
+  Layers,
+  Shapes,
+  Sparkle,
+  Users,
+  Workflow,
+  Wrench,
+} from "lucide-react";
+
+/**
+ * Single source of truth for domain-concept icons. Every surface that draws
+ * one of these concepts (filter fields, span kinds, nav tabs, chips, …)
+ * should import from here instead of picking a lucide icon directly — that's
+ * what let "model" drift to four different glyphs across the app (#1517).
+ *
+ * `project` and `workspace` are deliberately separate entries: they read as
+ * the same concept in the original issue text, but they're distinct concepts
+ * in this app (a workspace contains projects), each already icon-consistent
+ * on its own.
+ */
+export const DOMAIN_ICONS = {
+  tokens: CircleStop,
+  cost: CircleDollarSign,
+  model: Box,
+  agent: Bot,
+  llm: Sparkle,
+  tool: Wrench,
+  // Default/generic span kind (anything that isn't trace/llm/agent/tool).
+  span: ArrowRight,
+  assistant: BotMessageSquare,
+  user: Users,
+  session: Layers,
+  trace: Workflow,
+  id: Hash,
+  // Distinct concept from `id` (a count of things vs. an identifier) that
+  // happens to share the same glyph today.
+  count: Hash,
+  // "Category of kinds" — e.g. the span_kind filter/widget field, which
+  // covers all kinds at once rather than meaning one specific kind.
+  kind: Shapes,
+  status: CircleCheck,
+  latency: Clock,
+  error: AlertCircle,
+  project: FolderKanban,
+  workspace: LayoutGrid,
+  detector: Eye,
+  environment: Globe,
+  dashboard: LayoutDashboard,
+} satisfies Record<string, LucideIcon>;
--- a/frontend/ui/src/components/icons/domain-icons.ts
+++ b/frontend/ui/src/components/icons/domain-icons.ts
@@ -62,4 +62,9 @@ export const DOMAIN_ICONS = {
   detector: Eye,
   environment: Globe,
   dashboard: LayoutDashboard,
+  // Neutral "unknown field" fallback for filter/widget dropdowns. Kept
+  // decoupled from `model` even though both currently render as Box — if the
+  // model glyph ever changes, unmapped fields shouldn't silently change with
+  // it (ka1kqi, #1594).
+  fallback: Box,
 } satisfies Record<string, LucideIcon>;
--- a/frontend/ui/src/components/icons/domain-icons.ts
+++ b/frontend/ui/src/components/icons/domain-icons.ts
@@ -27,12 +27,12 @@ import {
  * Single source of truth for domain-concept icons. Every surface that draws
  * one of these concepts (filter fields, span kinds, nav tabs, chips, …)
  * should import from here instead of picking a lucide icon directly — that's
- * what let "model" drift to four different glyphs across the app.
+ * what left "model" as Box in the filters and Bot in settings, with no glyph at
+ * all on the span detail pill.
  *
  * `project` and `workspace` are deliberately separate entries: they read as
- * the same concept in the original issue text, but they're distinct concepts
- * in this app (a workspace contains projects), each already icon-consistent
- * on its own.
+ * the same concept at first glance, but they're distinct concepts in this app
+ * (a workspace contains projects), each already icon-consistent on its own.
  */
 export const DOMAIN_ICONS = {
   tokens: CircleStop,
@@ -67,4 +67,4 @@ export const DOMAIN_ICONS = {
   // model glyph ever changes, unmapped fields shouldn't silently change with
   // it.
   fallback: Box,
-} satisfies Record<string, LucideIcon>;
+} as const satisfies Record<string, LucideIcon>;
--- a/frontend/ui/src/components/icons/domain-icons.ts
+++ b/frontend/ui/src/components/icons/domain-icons.ts
@@ -1,10 +1,10 @@
 import type { LucideIcon } from "lucide-react";
 import {
-  AlertCircle,
   ArrowRight,
   Bot,
   BotMessageSquare,
   Box,
+  CircleAlert,
   CircleCheck,
   CircleDollarSign,
   CircleStop,
@@ -56,7 +56,7 @@ export const DOMAIN_ICONS = {
   kind: Shapes,
   status: CircleCheck,
   latency: Clock,
-  error: AlertCircle,
+  error: CircleAlert,
   project: FolderKanban,
   workspace: LayoutGrid,
   detector: Eye,
--- a/frontend/ui/src/features/dashboards/components/field-icons.ts
+++ b/frontend/ui/src/features/dashboards/components/field-icons.ts
@@ -6,7 +6,8 @@ import { DOMAIN_ICONS } from "@/components/icons/domain-icons";
 // names the trace list spells differently (errors -> error_count), the token
 // variants, a distinct count symbol (the shared map's Box doubles as the
 // model icon, so count falling back to it read as "model"), and the user /
-// session symbols the trace-page tabs use. Box stays the generic fallback.
+// session symbols the trace-page tabs use. DOMAIN_ICONS.fallback (Box) stays
+// the generic fallback for unmapped fields.
 const WIDGET_FIELD_ICONS: Record<string, LucideIcon> = {
   ...FIELD_ICONS,
   error_count: DOMAIN_ICONS.error,
@@ -24,4 +25,4 @@ const WIDGET_FIELD_ICONS: Record<string, LucideIcon> = {
 };
 
 export const fieldIcon = (field: string): LucideIcon =>
-  WIDGET_FIELD_ICONS[field] ?? DOMAIN_ICONS.model;
+  WIDGET_FIELD_ICONS[field] ?? DOMAIN_ICONS.fallback;
--- a/frontend/ui/src/features/dashboards/components/field-icons.ts
+++ b/frontend/ui/src/features/dashboards/components/field-icons.ts
@@ -1,16 +1,6 @@
-import {
-  AlertCircle,
-  Box,
-  CircleCheck,
-  CircleStop,
-  Hash,
-  Layers,
-  Shapes,
-  Users,
-  Workflow,
-  type LucideIcon,
-} from "lucide-react";
+import { type LucideIcon } from "lucide-react";
 import { FIELD_ICONS } from "@/features/filters/filter-controls";
+import { DOMAIN_ICONS } from "@/components/icons/domain-icons";
 
 // The trace-list filter icons, extended with the widget registry's field
 // names the trace list spells differently (errors -> error_count), the token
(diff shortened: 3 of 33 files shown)


frontend/ui/src/features/filters/filter-controls.tsx
```diff
@@ -239,7 +230,7 @@ export function FieldDropdown({
   onPick: (key: string) => void;
 }) {
   const selected = valueKey ? (options.find((o) => o.key === valueKey) ?? null) : null;
-  const Icon = selected ? (selected.icon ?? Box) : null;
+  const Icon = selected ? (selected.icon ?? DOMAIN_ICONS.model) : null;
```
ka1kqi: Minor: this repoints the neutral "unknown field" fallback (previously `Box`) onto the `model` concept. Pixel-identical today since `model === Box`, but if the model glyph ever changes, every unmapped field silently changes with it. A dedicated `DOMAIN_ICONS.fallback`/`generic` = `Box` would keep the two intentions decoupled. (Same applies to `field-icons.ts`'s `?? DOMAIN_ICONS.model`.)

frontend/ui/src/features/dashboards/components/field-icons.ts
```diff
@@ -19,18 +9,19 @@ import { FIELD_ICONS } from "@/features/filters/filter-controls";
 // session symbols the trace-page tabs use. Box stays the generic fallback.
```
ka1kqi: Stale after this change — the fallback now reads `?? DOMAIN_ICONS.model` and `Box` is no longer imported in this file, so "Box stays the generic fallback" no longer matches the code. Worth updating the wording.

frontend/ui/src/features/traces/components/SpanInfoPanel.test.tsx
```diff
@@ -189,4 +189,18 @@ describe("SpanInfoPanel - Error box source location badges", () => {
     expect(headerRefLabel).toBeTruthy();
     expect(headerRefLabel.className).toContain("shrink-0");
   });
+
+  it("renders User and Session link badges when the trace has both ids", () => {
```
ka1kqi, line 203: One of the two intentional visual changes — the model glyph added to the `model_name` pill — isn't asserted anywhere, unlike the `Bot`→`Box` change which `settings-layout.test.ts` locks. A quick assertion that the pill renders the model name alongside its icon would close the gap.

ka1kqi (commented):
Nice consolidation — the migration is thorough, no orphaned `lucide-react` imports left behind in any touched file, and the registry mapping has good test coverage. One thing I'd suggest addressing, plus a few minor notes inline.

**`project` registry entry is dead, and real project surfaces still drift.** `DOMAIN_ICONS.project` (`FolderKanban`) has no usages, while the onboarding "Project" step (`app/onboarding/page.tsx`) draws `Layers` — which the registry canonically assigns to `session` — and the projects empty-state (`app/workspaces/[workspaceId]/projects/page.tsx`) draws `FolderKanban` ad-hoc. That's exactly the one-glyph-two-concepts drift this PR is meant to eliminate, so it undercuts the description's claim that project is "already icon-consistent on its own." Worth either finishing the migration on those two surfaces (point them at `DOMAIN_ICONS.project`, move onboarding off `Layers`) or dropping the unused key so it doesn't ship speculative. (These files aren't in the diff, hence noting here rather than inline.)

A couple of Low notes inline. Really appreciate the care on the semantic modeling — keeping `id` vs `count` and `project` vs `workspace` as distinct entries is the right instinct.

frontend/ui/src/features/filters/filter-controls.tsx
Satyam296: Good point — added a separate `DOMAIN_ICONS.fallback` (Box) so it's decoupled from `model`. Same fix applied to `field-icons.ts`.

frontend/ui/src/features/dashboards/components/field-icons.ts
Satyam296: Fixed, updated to say `DOMAIN_ICONS.fallback` (Box) instead of the old model wording.

frontend/ui/src/features/traces/components/SpanInfoPanel.test.tsx
Satyam296: Added a test asserting the model_name pill renders both the name and its icon.

Satyam296:
@ka1kqi  Finished the migration instead of dropping the key onboarding's Project step and the projects empty-state now both use `DOMAIN_ICONS.project`.

All pushed in b066ced, CI's 17/17 green now.

frontend/ui/src/features/traces/components/SpanInfoPanel.test.tsx
```diff
@@ -142,6 +142,18 @@ describe("SpanInfoPanel - Error box source location badges", () => {
...
+  it("renders the model_name pill with its icon", () => {
+    render(<SpanInfoPanel projectId="proj-123" trace={mockTrace} selection={mockSelection} />);
+
+    const modelPill = screen.getByText("gpt-4o");
+    expect(modelPill).toBeTruthy();
+
+    const modelParentDiv = modelPill.parentElement;
+    expect(modelParentDiv).toBeTruthy();
+    const modelIcon = modelParentDiv?.querySelector("svg");
+    expect(modelIcon).toBeTruthy();
```
ka1kqi: This assertion is vacuous: `getByText("gpt-4o")` returns the pill `<div>` itself (the model name is its direct text node), so `modelPill.parentElement` is the whole badge row — which also contains the latency `Clock` svg — and `querySelector("svg")` finds an icon even if the model icon is deleted from the pill. Query inside the pill instead:

```suggestion
    const modelIcon = modelPill.querySelector("svg");
    expect(modelIcon).toBeTruthy();
```

frontend/ui/src/components/icons/domain-icons.ts
```diff
@@ -0,0 +1,70 @@
...
+  Users,
+  Workflow,
+  Wrench,
+} from "lucide-react";
+
+/**
+ * Single source of truth for domain-concept icons. Every surface that draws
+ * one of these concepts (filter fields, span kinds, nav tabs, chips, …)
+ * should import from here instead of picking a lucide icon directly — that's
+ * what let "model" drift to four different glyphs across the app (#1517).
```
ka1kqi: Repo convention: no issue/PR numbers in code comments — the prose already carries the full rationale, so just drop the `(#1517)`.

frontend/ui/src/components/icons/domain-icons.ts
```diff
@@ -0,0 +1,70 @@
...
+  error: AlertCircle,
+  project: FolderKanban,
+  workspace: LayoutGrid,
+  detector: Eye,
+  environment: Globe,
+  dashboard: LayoutDashboard,
+  // Neutral "unknown field" fallback for filter/widget dropdowns. Kept
+  // decoupled from `model` even though both currently render as Box — if the
+  // model glyph ever changes, unmapped fields shouldn't silently change with
+  // it (ka1kqi, #1594).
```
ka1kqi: Same convention here, and my username doesn't belong in source either — drop the `(ka1kqi, #1594)`; the sentence stands on its own.

frontend/ui/src/features/settings/settings-layout.tsx
```diff
@@ -24,16 +22,23 @@ interface SettingsTab {
 
 export const WORKSPACE_SETTINGS_TABS: SettingsTab[] = [
   { id: "general", label: "General", icon: SlidersHorizontal, href: "general" },
-  { id: "members", label: "Members", icon: Users, href: "members" },
-  { id: "model-providers", label: "Model Providers", icon: Bot, href: "model-providers" },
+  { id: "members", label: "Members", icon: DOMAIN_ICONS.user, href: "members" },
+  {
+    id: "model-providers",
+    label: "Model Providers",
+    // Canonical fix (#1517): was Bot, which also means the agent span kind.
```
ka1kqi: Drop the `(#1517)` — "Canonical fix: was Bot, …" reads fine on its own.

frontend/ui/src/features/settings/settings-layout.test.ts
```diff
@@ -0,0 +1,22 @@
+import { describe, it, expect } from "vitest";
+import { WORKSPACE_SETTINGS_TABS, PROJECT_SETTINGS_TABS } from "./settings-layout";
+import { DOMAIN_ICONS } from "@/components/icons/domain-icons";
+
+describe("WORKSPACE_SETTINGS_TABS", () => {
+  it("gates Model Providers on the model icon, not Bot (#1517 canonical fix)", () => {
```
ka1kqi: Same no-issue-numbers convention applies to test names — `"gates Model Providers on the model icon, not Bot (canonical fix)"` or just `"…not Bot"`.

frontend/ui/src/app/onboarding/page.tsx
```diff
@@ -12,7 +12,8 @@ import { Input } from "@/components/ui/input";
 import { Card, CardContent } from "@/components/ui/card";
 import { useLayout } from "@/components/layout/app-layout";
 import { createWorkspace, createProject } from "@/lib/api";
-import { LayoutGrid, Layers, Check } from "lucide-react";
+import { LayoutGrid, Check } from "lucide-react";
```
ka1kqi: The workspace-step icon further down (line 192) still draws this direct `LayoutGrid` import, while the registry has `workspace: LayoutGrid` and this file already imports `DOMAIN_ICONS` — the exact drift pattern this PR exists to eliminate (the sidebar's workspace icon *was* migrated). Pixel-identical one-liner; worth folding in.

ka1kqi (commented):
Thanks — the fallback decoupling, the `field-icons.ts` comment fix, and the `project` migration all check out, and the registry now has no dead entries. Two things still need fixing before this lands (inline): the new model-pill test doesn't actually test the icon, and a few comments/test names carry issue-PR numbers, which this repo keeps out of source (rationale in prose only — the prose is already there, so it's just deleting the parentheticals).

Two description-level notes:

- **PR body is stale.** It still says "pixel-identical except the two stated canonical fixes", but the onboarding Project step's `Layers`→`FolderKanban` swap is now a third visible change. Worth promoting it into the canonical-fix list.
- **Commit messages**: "Closes #1517" in the first commit and "flagged in #1594" in the latest also fall under the no-issue-refs convention (links belong in the PR description only). Moot if this squash-merges with a clean message.

Optional polish, no need to block on these: the two new render tests duplicate the same ~40-line trace fixture and virtualizer mock (a shared helper would halve it); the user-filter test name promises a "clearable badge" but never exercises the clear button; and the sessions/users tab tests assert labels only, so an icon regression wouldn't be caught — the identity-assertion pattern in `settings-layout.test.ts` is the stronger model if you want to extend it.

frontend/ui/src/features/traces/components/SpanInfoPanel.test.tsx
Satyam296: Good catch, that scoping bug would've let a deleted icon slip through silently. Fixed exactly as suggested — querying inside `modelPill` now.

frontend/ui/src/components/icons/domain-icons.ts
Satyam296: Dropped, thanks for the pointer on the convention.

frontend/ui/src/components/icons/domain-icons.ts
Satyam296: Dropped both the issue number and the username, the sentence stands fine alone.

frontend/ui/src/features/settings/settings-layout.tsx
Satyam296: Dropped.

frontend/ui/src/features/settings/settings-layout.test.ts
Satyam296: Renamed to drop the issue ref.

frontend/ui/src/app/onboarding/page.tsx
Satyam296: Good catch, missed that one. Now sources from DOMAIN_ICONS.workspace.

Satyam296:
Requested changes are incorporated @ka1kqi !

ka1kqi:
Rather than leave you waiting on a round trip, I pushed a few commits to this branch
directly — all fast-forward, so nothing of yours was rewritten and a `git pull` will
pick them up cleanly. Please don't force-push over it.

**What landed:**

1. A merge of latest `main`. The branch was 71 commits behind and had picked up a
   conflict.
2. `test(ui)`: repair for the three new page tests that the merge displaced. The
   sessions and users pages now import `useRetention` + `PricingDialog`, so they need
   those mocked (unmocked, the real hook runs a React Query hook and the test fails on
   a missing QueryClient, which points at the wrong culprit). And the traces page's
   onboarding gate no longer makes a second `useTraces({ limit: 1 })` call — it uses
   `useTracesExist`, so that mock needs `exists: true` or the filter bar under test is
   suppressed. The `users/page.test.tsx` you added collided with one that landed on
   `main` in the meantime; I kept `main`'s and folded your empty-state assertion into
   it, since `main`'s only covered the loading state.
3. `chore(ui)`: a `no-restricted-imports` rule so the registry can't be bypassed, plus
   some test tightening — the settings-tab test now asserts against the lucide glyphs
   rather than the registry entries it's built from (the old form would have passed
   even with an entry repointed at the wrong glyph), the token/cost badge tests assert
   a glyph actually renders, and the last two detector surfaces that still drew `Eye`
   directly now source it from the registry.

**On that third commit:** the lint rule is a repo-wide policy decision, and it's ours,
not yours — you shouldn't be in the position of defending a rule you didn't write. It's
scoped to glyphs with a single meaning in this app, so there are no pre-existing
violations and nothing else had to change. If you'd rather it not ride along in your PR,
say so and I'll pull it out into its own change.

Verified locally before pushing: full frontend suite green, `eslint .` and
`prettier --check` both clean, diff-cover at 98%.

The registry itself is untouched. Nice work on it — the doc comment explaining *why* it
exists is the part that'll keep it alive, and pushing back on the issue's framing instead
of implementing it literally was the right call.

ka1kqi (approved):
Approving. I audited every migrated call site glyph-by-glyph and the substitutions are 1:1
in both glyph and className, except the three visual changes you called out (Model
Providers `Bot`→`Box`, the `model_name` pill gaining a glyph, the onboarding/empty-state
project icon). So the "pure consolidation otherwise" claim holds. The registry has no dead
entries, and the free-list carve-outs are each justified in the code — `Clock` for
"Alert window", `Sparkle` for the Using-AI tab, `Eye` for password toggles, `Hash` for a
Slack channel — all correctly left unmigrated.

CI is green (17 pass, 1 neutral from the review bot skipping) and diff-cover is at 98%.

Non-blocking follow-ups, none of which need to hold this up:

- `page.user-filter.test.tsx` duplicates ~60 lines of mock scaffolding from its sibling
  `page.prefetch.test.tsx`, which already uses mutable module-level state for exactly this
  "vary one input" case. Worth collapsing into that file later.
- Three doc-comment wordings in `domain-icons.ts` and `settings-layout.tsx` reference the
  originating discussion or describe the diff rather than the durable invariant, which will
  read oddly once merged. I'll fold those into a small follow-up.
- The lint ban pattern covers lucide's `*Icon` suffix and its deprecated aliases but not
  the `Lucide*`-prefixed export names. Nothing uses that spelling today; same follow-up.
- `fallback` is still `Box`, the same glyph as `model`, so an unmapped widget field
  continues to read as "model". Pre-existing rather than introduced here, and the
  decoupling you added is the right groundwork for fixing it properly.

XinweiHe (approved):
LGTM

PR #1602, opened 2026-07-23
fix(detectors): stop the screening-model picker from silently pinning an auto-picked model

ka1kqi:
## Summary

The detector create form and edit panel reuse the AI-assistant `ModelSelector`, whose auto-pick effect wrote the first-priority model (`claude-opus-4-8`) into form state whenever the incoming selection was empty. Submitting the create form, or saving **any** field in the edit panel of an unpinned detector, then persisted that as a pinned `detectionModel` — silently moving detection screening (one LLM eval per sampled trace at ingestion, the highest-volume LLM workload) off the cheap default (`claude-haiku-4-5`) onto the most expensive model, which is also marked up on paid plans.

This adds an opt-out `defaultModelId` prop to `ModelSelector`: when set, an empty selection is a valid persistent state — the picker presents that model as if it were the selection instead of writing one into form state, and only an explicit pick of a *different* model pins anything. Both detector screening surfaces pass it.

It also closes the other half of the same inconsistency in the worker: a detector whose `detectionSource` was never set (`null`) was excluded from the screening default and fell through to the provider-priority model, even though every UI surface labelled it as tracking the default.

## Changes

- `ModelSelector` gains an optional `defaultModelId` prop. When set, the reconcile effect no longer auto-picks for an empty selection, and that model is presented as the selection — same trigger label, same checkmark — so an unpinned parent is indistinguishable from one that stored it. Clicking that row is suppressed, since it would look like a no-op while pinning the model and opting the parent out of later changes to the default. The adapter/legacy backfill branches and every consumer that omits the prop (assistant chat, the Agent Model pickers) are unchanged.
- The create form and edit panel pass `DETECTOR_SYSTEM_DEFAULT_MODEL_ID`, so an unpinned detector shows the bare model id.
- The detectors list model column drops its `(default)` marker and shows the model name alone, matching the picker. That was the only consumer of the helper's `isDefault` flag, so it returns a plain label string now.
- An untouched create form now persists `detectionModel: null`; the edit panel no longer converts an unpinned detector to a pin on an unrelated save.
- The worker resolves a null `detectionSource` as `"system"`, so legacy and API-created rows screen on the same default as everything else. That makes the list page's separate `"Auto-selected"` branch unreachable, so it is removed.

## Behavior

- Created or edited without touching the picker → detector tracks the system default, shown as `claude-haiku-4-5` in both the picker and the list.
- Pick a model explicitly → that model is pinned and shown.
- An unpinned detector and one explicitly pinned to the same model now read identically everywhere. The distinction survives in the data — `detectionModel: null` follows the default if it ever changes, a pin does not — but is deliberately not surfaced.
- Legacy `detectionSource: null` rows now screen on the system default instead of the provider-priority model, which is what the UI already claimed they did.
- No data migration: rows already silently pinned are indistinguishable from deliberate picks, so this only stops new pins. Existing pinned rows keep their model.

## Deployment invariant this establishes

Unpinned detectors now resolve to a **fixed** default model rather than an availability-aware provider pick. The previous auto-pick was availability-aware by construction, because the system-model catalog endpoint filters `SYSTEM_MODELS` by env-var presence — so a deployment with only `OPENAI_API_KEY` used to pin a working OpenAI model.

After this change, a deployment that holds no key for the default model's provider (neither in env nor as a workspace BYOK row) will fail those evals with `No API key configured` rather than quietly screening on whatever else is configured. This is accepted deliberately: it is the same tradeoff every explicitly-`"system"` detector already made, and it keeps the highest-volume LLM workload on a known cheap model instead of silently drifting to whatever a deployment happens to have. Deployments are expected to configure a key for the provider of `DETECTOR_SYSTEM_DEFAULT_MODEL_ID`.

## Tests

- `ModelSelector`: the default suppresses the auto-pick, is ticked, is not written when picked, is still selectable once another model is pinned, is withheld from a BYOK selection, and is not ticked twice when a BYOK provider shares the id; the legacy model-only backfill still works across sources. Existing auto-pick tests unchanged, proving the other surfaces are unaffected.
- Create form and edit panel: assert the picker receives the system-default model id.
- `packages/core`: `isSystemModelId` accepts every catalog id and the screening default, and rejects retired ids.
- Worker: null-source detectors resolve to the screening default, and attribute the run as `system` — billing is unaffected, since the only consumer asks whether the source is `byok`.
- List page: null-source rows render the default label rather than the removed `"Auto-selected"` string, and no row renders a `(default)` marker.

## Displayed model vs. model that runs

Three ways those could differ are closed here:

- A pinned model id that has left the catalog resolved to an unrelated fallback rather than failing, so retiring a model would silently move every detector pinned to it onto that fallback while the UI kept showing the retired id. The worker now rejects the id. The detector stops producing findings instead of producing them off the wrong model; the run row has no error column, so the reason reaches the worker log only.
- A BYOK detector with no pinned model was shown the system default, which is not what it runs.
- An id-only backfill could bind a system selection to a BYOK provider exposing the same id — same model name on screen, different credentials at run time.

## Known gaps, deliberately not addressed here

- Un-pinning is not reachable from the UI: the popover lists only concrete models, and the patch builder omits an empty model, so a user who pins a model cannot return the detector to default-tracking. Pre-existing, but this change makes "empty" a first-class state and so raises the question.
- A `detectionSource: "byok"` row with no model now reads "Select model" rather than naming the wrong one, but still does not name the model that runs (the adapter's first curated one). That needs the deferred write-time validation.
- The project RCA/agent model has the same displayed-vs-run gap on a different surface: an unknown stored `rca_model` falls back to a hardcoded model while the settings picker keeps showing the stored id. Out of scope here; tracked separately.
- Detector write failures are still invisible to the user; tracked separately.

## Screenshots

The earlier screenshots showed the interim `System default (claude-haiku-4-5)` wording and have been removed — the picker now renders the bare model id.

### DIFF
--- a/frontend/packages/core/src/__tests__/model-resolver.test.ts
+++ b/frontend/packages/core/src/__tests__/model-resolver.test.ts
@@ -16,11 +16,13 @@ vi.mock("../lib/encryption", () => ({
 
 import {
   resolvePiModel,
+  isSystemModelId,
   fetchProviderConfig,
   findByokKeyForPiProvider,
   invalidateProviderConfigCache,
   type ProviderModelConfig,
 } from "../model-resolver.ts";
+import { SYSTEM_MODELS, DETECTOR_SYSTEM_DEFAULT_MODEL_ID } from "../llm-providers.ts";
 import { prisma } from "../lib/prisma.ts";
 
 describe("resolvePiModel", () => {
@@ -204,5 +206,26 @@ describe("findByokKeyForPiProvider", () => {
     );
     const key = await findByokKeyForPiProvider("ws-5", "anthropic");
     expect(key).toBeNull();
+  });
+});
+
+describe("isSystemModelId", () => {
+  it("accepts every id the catalog offers", () => {
+    const ids = SYSTEM_MODELS.flatMap((s) => s.models.map((m) => m.id));
+    expect(ids.length).toBeGreaterThan(0);
+    for (const id of ids) {
+      expect(isSystemModelId(id)).toBe(true);
+    }
+  });
+
+  // Callers reject ids this returns false for, so a default that fell out of
+  // the catalog would fail every unpinned detector rather than degrade.
+  it("accepts the detector screening default", () => {
+    expect(isSystemModelId(DETECTOR_SYSTEM_DEFAULT_MODEL_ID)).toBe(true);
+  });
+
+  it("rejects a retired or invented id", () => {
+    expect(isSystemModelId("claude-retired-1")).toBe(false);
+    expect(isSystemModelId("")).toBe(false);
   });
 });
--- a/frontend/packages/core/src/model-resolver.ts
+++ b/frontend/packages/core/src/model-resolver.ts
@@ -48,6 +48,15 @@ for (const sys of SYSTEM_MODELS) {
 }
 
 /**
+ * Whether `modelId` is one of the compiled-in system models. Lets callers
+ * holding a *stored* id reject one that has left the catalog, which
+ * `resolvePiModel` would otherwise answer with a substitute model.
+ */
+export function isSystemModelId(modelId: string): boolean {
+  return systemModelLookup.has(modelId);
+}
+
+/**
  * Build a pi-ai Model object with the correct `api` protocol while preserving
  * pi-ai registry data (pricing, context window) when available.
  * Falls back to a manual object for models not in the registry.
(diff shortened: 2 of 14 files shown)


XinweiHe:
maybe remove the system default ..., just say haiku-4.5 as usual should be fine.

XinweiHe (approved):
Let's do these before the merge
1. need new screenshots
2. fix the merge conflict

ka1kqi:
> Let's do these before the merge
> 
> 1. need new screenshots
> 2. fix the merge conflict

Screenshot 2026-07-29 at 1 10 29 PM
Screenshot 2026-07-29 at 1 10 33 PM

PR #1611, opened 2026-07-24
feat(dashboards): expose cache read/write token measures in the widget builder

### LINKED ISSUE #1608: feat(dashboards): expose cache read/write token measures in the widget builder
## Problem

Dashboard widgets can chart **Input tokens**, **Output tokens**, and **Total tokens**, but not cache tokens — even though every span already stores them in ClickHouse, in the `usage_details` map (keys `cache_read_tokens`, `cache_write_tokens`; see `backend/db/clickhouse/migrations/004_add_token_breakdown_columns.sql`).

So a user monitoring prompt-cache effectiveness can't build widgets like "Cache read tokens by model" or "Cache write tokens over time", even though the span detail panel already shows these numbers per span.

## Proposed change

Add two new measures alongside the existing token measures, in both the **spans** and **traces** views of the widget field registry:

| Measure id | Label |
|---|---|
| `cache_read_tokens` | Cache read tokens |
| `cache_write_tokens` | Cache write tokens |

All the work is in `backend/rest/services/widget_registry.py`:

1. **Spans view** — in `_SPANS_BASE`, select `usage_details` in the inner deduped subquery, then expose the map values alongside the existing token columns, e.g.
   `if(total_tokens IS NOT NULL, usage_details['cache_read_tokens'], NULL) AS cache_read_tokens` (same for write).
   The `if(...)` keeps the semantics consistent with the existing `Nullable` token columns: non-LLM spans yield `NULL` (excluded from `avg`/`min`), while LLM spans without cache activity yield `0`. A bare `usage_details['cache_read_tokens']` would instead return `0` for every span and skew `avg`.
2. **Traces view** — in `_TRACES_BASE`, add the same map accesses to the inner spans subquery, `sum(...)` them in the per-trace aggregation, and wrap with the same `if(sa.trace_id = '', NULL, ...)` treatment as the existing token sums.
3. Register both fields as numeric measures (same shape as `input_tokens`) in `REGISTRY["spans"]` and `REGISTRY["traces"]`.

No frontend change should be needed: the builder's measure dropdown is driven by `GET /projects/{projectId}/widgets/schema`, so the new measures appear automatically — please verify the dropdown, generated widget titles, and rendering as part of testing.

## Notes

- `cache_write_tokens` already includes the 1h-TTL sub-partition (`cache_write_1h_tokens` in the map is a sub-count of it, stamped only when non-zero) — expose only the two top-level cache measures, don't add the 1h key as its own measure.
- Keep label wording consistent with the span detail panel's token breakdown ("cache read" / "cache write").
- Tests: extend the existing widget registry/query tests to cover the new measures (schema output + compiled SQL + a query over seeded spans whose sums match the span-level `usage_details` values). CI enforces ≥ 80% diff coverage on changed lines.

## Out of scope / future work

A richer treatment — splitting the tokens measure itself into stacked uncached-input / cache-read / cache-write / output segments per bar — is deliberately deferred; this issue only adds the standalone measures.

Gaurav-meena95:
…t builder\n\nFixes #1608

Fixes <issue-id>

## Type of Change

- [ ] feat - A new feature
- [ ] fix - A bug fix

## Checklist

- [ ] I have read the [CONTRIBUTING.md](../CONTRIBUTING.md)
- [ ] I have added/updated tests where applicable

---
## Summary by cubic
Exposes cache read/write token measures in the dashboards widget builder so you can chart cache usage on spans and traces. Orders token measures so Total tokens appears last. Fixes #1608.

- **New Features**
  - Adds `cache_read_tokens` and `cache_write_tokens` to `spans` and `traces`.
  - Reads from `usage_details[...]`, aggregates at trace level, and preserves NULLs when not reported.
  - Orders token measures as input, output, cache read, cache write, then total; updates schema/icons and adds tests.

<sup>Written for commit 74049fbfde76fda82886bd4ecdb2a3486d117807. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1611?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/rest/services/widget_registry.py
+++ b/backend/rest/services/widget_registry.py
@@ -38,20 +38,29 @@ class ViewDef:
     fields: dict[str, FieldDef] = field(default_factory=dict)
 
 
-_SPANS_BASE = """
+# Preserve NULL when cache token data was never reported.
+_CACHE_READ_TOKENS_EXPR = (
+    "if(mapContains(usage_details, 'cache_read_tokens'), usage_details['cache_read_tokens'], NULL)"
+)
+_CACHE_WRITE_TOKENS_EXPR = "if(mapContains(usage_details, 'cache_write_tokens'), usage_details['cache_write_tokens'], NULL)"
+
+
+_SPANS_BASE = f"""
     SELECT
         name, span_kind, status, model_name, environment,
         span_start_time AS event_time,
         dateDiff('millisecond', span_start_time, span_end_time) AS duration_ms,
-        cost, input_tokens, output_tokens, total_tokens
+        cost, input_tokens, output_tokens, total_tokens,
+        {_CACHE_READ_TOKENS_EXPR} AS cache_read_tokens,
+        {_CACHE_WRITE_TOKENS_EXPR} AS cache_write_tokens
     FROM (
         SELECT
             span_id, trace_id, name, span_kind, status, model_name, environment,
-            span_start_time, span_end_time, cost, input_tokens, output_tokens, total_tokens
+            span_start_time, span_end_time, cost, input_tokens, output_tokens, total_tokens, usage_details
         FROM spans
-        WHERE project_id = {project_id:String}
-          AND span_start_time >= {start_time:DateTime64(3)}
-          AND span_start_time < {end_time:DateTime64(3)}
+        WHERE project_id = {{project_id:String}}
+          AND span_start_time >= {{start_time:DateTime64(3)}}
+          AND span_start_time < {{end_time:DateTime64(3)}}
         ORDER BY ch_update_time DESC
         LIMIT 1 BY project_id, trace_id, span_id
     )
@@ -64,7 +73,7 @@ _SPANS_BASE = """
 # cost, and token counts here may differ from the per-trace detail page (which
 # joins all spans for a trace). The tradeoff is a bounded, fast scan for
 # dashboards vs. exact per-trace metrics in the trace list.
-_TRACES_BASE = """
+_TRACES_BASE = f"""
     SELECT
         t.name AS name, t.user_id AS user_id,
         t.session_id AS session_id, t.environment AS environment,
@@ -77,14 +86,16 @@ _TRACES_BASE = """
         if(sa.trace_id = '', NULL, sa.total_cost) AS cost,
         if(sa.trace_id = '', NULL, sa.input_tokens) AS input_tokens,
         if(sa.trace_id = '', NULL, sa.output_tokens) AS output_tokens,
-        if(sa.trace_id = '', NULL, sa.total_tokens) AS total_tokens
+        if(sa.trace_id = '', NULL, sa.total_tokens) AS total_tokens,
+        if(sa.trace_id = '', NULL, sa.cache_read_tokens) AS cache_read_tokens,
+        if(sa.trace_id = '', NULL, sa.cache_write_tokens) AS cache_write_tokens
     FROM (
         SELECT
             trace_id, name, user_id, session_id, environment, trace_start_time
         FROM traces
-        WHERE project_id = {project_id:String}
-          AND trace_start_time >= {start_time:DateTime64(3)}
-          AND trace_start_time < {end_time:DateTime64(3)}
+        WHERE project_id = {{project_id:String}}
+          AND trace_start_time >= {{start_time:DateTime64(3)}}
+          AND trace_start_time < {{end_time:DateTime64(3)}}
         ORDER BY ch_update_time DESC
         LIMIT 1 BY project_id, trace_id
     ) AS t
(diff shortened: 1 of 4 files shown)


ka1kqi:
Thanks for picking this up — the approach is right (registry-only change, so the builder picks the measures up for free), the key names match what the ingest writer persists, and the traces-side NULL handling lines up with how `total_tokens` already behaves.

I also checked whether pulling the `usage_details` map into the base scan would cost widgets that don't ask for it: it doesn't. ClickHouse prunes `base_sql` columns per requested measure — a traces `count` widget reads zero payload columns from `spans` and its `read_bytes` is unchanged by this PR, while `sum(cache_read_tokens)` reads `total_tokens` + `usage_details` and nothing else. No perf concern.

One thing to change, then a few nits.

## 1. `total_tokens IS NOT NULL` isn't the right guard

Cache tokens are never estimated — the text-estimation fallback in `otel_transform.py` sets `total_tokens`/`cost` and never writes `usage_details`, and nothing else synthesizes a cache figure. So the map has exactly two meaningful states: key present = the provider reported it (and a `0` there is a real measured zero), key absent = we never had the data. ClickHouse Map access returns `0` for an absent key rather than NULL (migration `004_add_token_breakdown_columns.sql` notes the map defaults to `{}`), so the current guard collapses those two states into one.

`total_tokens IS NOT NULL` can't distinguish them, because `total_tokens` gets set by estimation too. Two populations land on the wrong side of it:

- text-estimated LLM spans (no provider usage at all), and
- every span ingested before `usage_details` existed — 004 added the column with no backfill, so all that history carries `{}`.

Both report cache read/write = 0. That's harmless for `sum`, but `_number_measure` also exposes `avg`/`min`/`p50`/`p95`, and those get dragged toward zero by spans we simply have no cache data for.

`mapContains` says what the data actually means, and subsumes the non-LLM case (empty map → absent → NULL) so the dependency on an unrelated column disappears:

```sql
if(mapContains(usage_details, 'cache_read_tokens'), usage_details['cache_read_tokens'], NULL) AS cache_read_tokens
```

It's also safe in the other direction: the writer always emits both cache keys together, so a genuine provider-reported `0` stays `0` and never becomes NULL.

(`trace_reader.py` and the trace-detail frontend do default a missing key to `0`, but both are summing for a cost/total — that doesn't settle it for an averaged measure.) Either way, please add the one-line rationale comment; every other NULL decision in this file has one.

## 2. The `usage_details[...]` assertions in the new tests don't test anything

Those expressions are baked into `base_sql` unconditionally, so a `cost`-metric widget's SQL also contains `usage_details['cache_read_tokens']` — the assertion passes regardless of the measure. Only `sum(cache_read_tokens)` is measure-specific. Either drop the second assertion in each test, or swap it for something that pins the semantics (the NULL guard, or an `avg` compile to cover the non-additive path).

Similarly, `test_registry_schema_exposes_cache_tokens` mostly restates the registry dict — `test_every_field_expr_is_reachable_in_base_sql` already covers new fields generically. The behavior that's actually untested is the traces path: per-span map read → `sum` → the `sa.trace_id = ''` NULL-out.

## 3. Small frontend gap

`features/dashboards/components/field-icons.ts` maps `input_tokens`/`output_tokens`/`total_tokens` to `CircleStop`; the two new measures fall through to `Box`, which that file's own comment flags as reading like "model". Two lines in `WIDGET_FIELD_ICONS`. (No `FIELD_UNIT` entry needed — tokens are unitless.)

## 4. Nits

- The guard expression is spelled out four times across the two SQL strings; a shared module-level fragment would keep them from drifting.
- The commit subject came through as `feat(dashboards): … widget builder\n\nFixes #1608` — the `\n\n` wasn't interpreted, so it's one 93-char subject (which is also why the PR title is truncated). We also keep issue refs out of commit messages entirely; `Fixes #1608` in the PR description is enough. Could you reword the commit?
- The PR description still has the template placeholders (`Fixes <issue-id>`, empty Summary, unchecked boxes) — mind filling those in?

Only (1) needs a code change; the rest is polish. Nice, tightly scoped work.

ka1kqi:
Also please include a screenshot of the ui with all ui changes

ka1kqi (approved):
LGTM — implementation matches the issue plan, keys line up with ingestion's usage_details map, and the f-string placeholder escaping is correct in both base templates.

One deliberate deviation, accepted: gating on `mapContains(usage_details, ...)` instead of `total_tokens IS NOT NULL` means spans whose token counts came from text estimation (or predate the usage_details migration) report NULL rather than 0 for cache tokens. That's the more truthful semantics — those spans have no cache data — and real LLM spans without cache activity still correctly report 0, so avg/min behave as intended for the reported-usage population.

XinweiHe:
Screenshot 2026-07-29 at 12 45 30 PM

fix the formating?

XinweiHe:
could we add some screenshots?

ka1kqi:
Screenshot 2026-07-29 at 3 12 36 PM
image

ka1kqi:
@Gaurav-meena95 made some slight amendments to the ui, also for future PRs that incorporate some ui change please include a screenshot of the results

ka1kqi (approved):

XinweiHe (approved):

PR #1612, opened 2026-07-24
fix(worker): stop double-counting usage from semconv agent wrapper spans

### LINKED ISSUE #1610: fix(worker): Vercel AI SDK v7 semconv telemetry double-counts tokens and cost on operation root spans
## Summary

## Repro / observed data

|---|---|---|---|---|---|

## Proposed fix

Notes:

## Caveat

ka1kqi:
## Summary

Traces from the Vercel AI SDK's semconv OTel emitter (`@ai-sdk/otel`, tracer scope exactly `"gen_ai"`) were stored with token usage **and cost on both the operation root span and its LLM child**, so every trace-level aggregate (trace list, trace-tree root chip, dashboards, digests) showed double the real usage — worse than 2x for multi-step agent runs, since the root restates the **sum** of all its children. Verified on a live staging trace: a single 10→16-token call displayed as 20→32 (52) at double the cost.

**Root cause:** the semconv emitter stamps aggregate `gen_ai.usage.*` totals on the operation root (`onGenerateEnd` in `@ai-sdk/otel`), restating its `chat <model>` children. The ingest transform trusts `gen_ai.usage.*` on any span kind with a model attribute (deliberate — some instrumentors report usage only on AGENT/CHAIN spans). The existing wrapper guard from #1187 only gates the **raw** `ai.usage.*` spellings; the semconv emitter moved the duplicates into the normalized namespace under a new tracer scope, bypassing it. The duplication is invisible in the UI because token chips render only on LLM-kind spans.

## Changes

- **`backend/worker/otel_transform.py`**
  - Gate API-provided token adoption: on tracer scope exactly `"gen_ai"`, usage is trusted only on LLM-kind spans. Skipping the branch also skips `usage_details` buckets and cost (all derived inside it). Model-name extraction is kept for display. No behavior change for any other scope.
  - `get_span_kind`: map `gen_ai.operation.name: "invoke_agent"` → AGENT, decided before the has-model→LLM fallback. Semconv operation roots carry `gen_ai.request.model`, so without this a raw `@ai-sdk/otel` export (no openinference kind attrs) would classify the root as LLM and dodge the gate. Also the semantically correct kind for these spans.
- **`backend/worker/tokens/buckets.py`** — register `"gen_ai"` as a known gross scope (exact match, mirroring `"ai"`); it previously fired the unknown-scope warning.

## Tests

- Semconv trace shape taken verbatim from the staging trace (AGENT root with aggregate usage + `step N` + `chat <model>` leaf): root stores no tokens/cost, leaf stores 10/16/26 with cost, trace-wide sums equal the single real call.
- Same shape **without** `openinference.span.kind` attrs — covers raw `@ai-sdk/otel` exports; passes via the `invoke_agent` → AGENT mapping.
- Control: an identical usage-bearing AGENT span under a different scope is still priced — the gate is fenced to `"gen_ai"`.
- `get_span_kind` units (`invoke_agent` → AGENT, including with a model attr present) and a buckets known-scope test.
- Full backend suite: 892 passed, 1 skipped (pre-existing).

## Known limitations (see issue)

- Ingest-time fix only — already-stored rows remain doubled unless re-ingested; no backfill.
- Embeddings/rerank operations have the same root+leaf duplicate pattern but both spans classify as LLM, so they keep today's (doubled) behavior — separate follow-up.

Closes #1610

---
## Summary by cubic
Fixes double-counted token usage and cost for Vercel AI SDK semconv traces. Totals in trace lists, trees, dashboards, and digests now match the real LLM calls.

- **Bug Fixes**
  - Detect aggregate agent wrappers by `gen_ai.operation.name` (`invoke_agent`, `agent_step`) and ignore their `gen_ai.usage.*`; also skip text-based estimation for them. Model names are still extracted.
  - Fallback when no operation name: treat non-LLM spans on scope `gen_ai` as aggregates so wrapper usage is still ignored.
  - Classify `gen_ai.operation.name: invoke_agent` as AGENT in `get_span_kind`, so agent roots don’t appear as LLM spans.
  - Register `gen_ai` as a known gross scope in token buckets to apply standard cache subtraction and avoid warnings.

<sup>Written for commit 95c16b749634d154e30b9b5b4623104022cd1a60. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1612?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/worker/otel_transform.py
+++ b/backend/worker/otel_transform.py
@@ -268,6 +268,10 @@ def get_span_kind(attrs: dict[str, Any], otel_kind: int | str | None) -> str:
         return SpanKind.LLM
     if operation_name == "execute_tool":
         return SpanKind.TOOL
+    # Agent-invocation roots carry gen_ai.request.model too; decide AGENT here
+    # so the model fallback below cannot flip them to LLM.
+    if operation_name == "invoke_agent":
+        return SpanKind.AGENT
 
     # Infer from LLM-related attributes
     if (
@@ -538,7 +542,23 @@ def transform_otel_to_clickhouse(
                         ],
                     )
 
-                    if api_input_tokens is not None or api_output_tokens is not None:
+                    # The Vercel AI SDK's semconv emitter (tracer scope exactly
+                    # "gen_ai") stamps aggregate gen_ai.usage.* totals on the
+                    # operation root span, restating the SUM of its LLM children
+                    # — the same wrapper pattern as the raw ai.usage.* keys
+                    # gated above, moved into the normalized namespace. Trust
+                    # usage on that scope only on the LLM span itself; other
+                    # scopes are untouched (some instrumentors legitimately
+                    # report usage only on AGENT/CHAIN spans).
+                    vercel_semconv_wrapper = (
+                        isinstance(scope_name, str)
+                        and scope_name.lower() == "gen_ai"
+                        and span_kind != SpanKind.LLM
+                    )
+
+                    if not vercel_semconv_wrapper and (
+                        api_input_tokens is not None or api_output_tokens is not None
+                    ):
                         # Use API-provided counts (accurate).
                         input_tokens = int_or_zero(api_input_tokens)
                         output_tokens = int_or_zero(api_output_tokens)
--- a/backend/worker/otel_transform.py
+++ b/backend/worker/otel_transform.py
@@ -53,6 +53,42 @@ def _scope_skips_text_token_estimation(scope_name: str | None) -> bool:
     return scope_name in _SKIP_TEXT_TOKEN_ESTIMATION_SCOPES
 
 
+# GenAI semconv operation names whose spans AGGREGATE the usage of their model-call
+# descendants: an agent invocation, and one step within it. Emitters restate the
+# children's totals on these spans, so adopting them prices the same tokens twice.
+# Keyed on the operation name rather than the tracer scope: the scope is chosen by
+# the emitting application (a private tracer provider can name it anything), so a
+# scope-keyed check silently misses those traces. "invoke_agent" is a GenAI semconv
+# operation name; "agent_step" is an emitter extension, not in the spec enum.
+_AGGREGATE_USAGE_OPERATIONS = frozenset({"invoke_agent", "agent_step"})
+
+
+def _span_aggregates_child_usage(
+    scope_name: str | None, span_kind: str, attrs: dict[str, Any]
+) -> bool:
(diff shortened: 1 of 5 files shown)


ka1kqi:
## Update — the scope-fenced gate did not cover the motivating case

Pushed a follow-up commit that re-keys the gate. Summary of why, since it changes the rule this PR is built on.

**The instrumentation scope is chosen by the application.** An app that builds a private `NodeTracerProvider` — the recommended shape when you don't want every HTTP span exported — names its own tracer, so its spans arrive under an arbitrary scope rather than `gen_ai`. Measured on a replica of exactly that shape, real usage 10 input tokens:

| code | stored trace input |
|---|---|
| `main` | 20 — double counted |
| this PR, as originally submitted | **20 — still double counted** |
| this PR, with the follow-up commit | 10 — correct |

So the fix as first written would have closed the issue while leaving that class of user broken, and any workaround they had reverted on the strength of it would have silently restored 2x.

**What changed**

- The predicate now keys on `gen_ai.operation.name` ∈ {`invoke_agent`, `agent_step`}, which is emitter- and scope-independent and holds even when a processor labels the span LLM. The scope check stays as a narrow fallback for a wrapper arriving without an operation name, and now has a test pinning it.
- `agent_step` is newly covered. Those spans aggregate the model calls inside them *and* classify LLM, so the span-kind half of the original check structurally could not catch them.
- Refusing an aggregate's reported counts now also excludes it from the text-estimation fallback. Without that, an LLM-kind aggregate carrying the conversation text fell through and had tokens **fabricated** from it — worse than the double count, since the numbers are invented rather than merely restated. Regression test added.
- Test fixture naming genericised; the stale prose describing the old scope+kind rule updated.

**Correction to the description above:** the "No behavior change for any other scope" line applies to the token gate only. The `get_span_kind` change (`invoke_agent` → AGENT) applies to every emitter and does change stored span kinds, which affects kind filters and the trace tree.

**Verified end-to-end**, not only by unit tests: real model calls through the SDK → ingest → ClickHouse, on a matched pair with identical input. `main` stored 10,562 input tokens for a trace whose real usage was 5,281; this branch stores 5,281, with the operation roots null and the per-call children exact.

XinweiHe (approved):

PR #1615, opened 2026-07-25
fix(worker): Avoid Vercel v7 operation usage double count

### LINKED ISSUE #1610: fix(worker): Vercel AI SDK v7 semconv telemetry double-counts tokens and cost on operation root spans

Gujiassh:
Fixes #1610

## Summary

- Ignore aggregate `gen_ai.usage.*` values on Vercel AI SDK v7 `invoke_agent` roots and non-LLM wrappers.
- Preserve the root's model, input, and output while keeping exact `chat` child usage and other instrumentation scopes unchanged.
- Register the exact `gen_ai` tracer scope so valid inference spans no longer emit an unknown-scope warning.

## Details

The v7 emitter uses the exact `gen_ai` scope. Its `invoke_agent` operation root repeats the sum of the `chat` inference children, so ingest previously stored and priced both copies. The emitter-specific gate now suppresses both API-provided and text-estimated usage on the aggregate wrapper.

This does not change span classification, token bucket math, storage schemas, or non-Vercel emitters. It affects newly ingested rows only; existing rows are not backfilled.

## Validation

- Added regression coverage for raw exports and roots marked as AGENT by OpenInference.
- Added a control proving another emitter's authoritative AGENT usage remains priced.
- `uv run coverage run -m pytest tests/`: 890 passed, 1 skipped.
- `uv run ruff check .`
- `uv run ruff format --check .`

---
## Summary by cubic
Prevent double-counted usage and cost from Vercel AI SDK v7 by ignoring aggregate `gen_ai.usage.*` on `invoke_agent` roots and using child `chat` LLM spans instead. Root still records model, input, and output; only new ingests change.

- **Bug Fixes**
  - Ignore usage on `gen_ai` operation wrappers and non-LLM spans; rely on child `chat` spans for tokens and cost.
  - Recognize the `gen_ai` tracer scope to remove unknown-scope warnings.
  - No changes to pricing math, schemas, or non-Vercel emitters.
  - Added tests for raw exports and AGENT roots; other emitters’ AGENT usage remains authoritative.

<sup>Written for commit 73a968e6ca1d6ca4bc9fe225a88b0a852e41a8d7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1615?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/worker/otel_transform.py
+++ b/backend/worker/otel_transform.py
@@ -53,6 +53,14 @@ def _scope_skips_text_token_estimation(scope_name: str | None) -> bool:
     return scope_name in _SKIP_TEXT_TOKEN_ESTIMATION_SCOPES
 
 
+def _scope_skips_token_usage(scope_name: str | None, span_kind: str, attrs: dict[str, Any]) -> bool:
+    """Reject aggregate usage copied onto Vercel v7 operation wrappers."""
+    if scope_name != "gen_ai":
+        return False
+    operation_name = attrs.get("gen_ai.operation.name")
+    return operation_name == "invoke_agent" or span_kind != SpanKind.LLM
+
+
 # Attributes that are already extracted into dedicated fields
 _KNOWN_ATTRIBUTE_PREFIXES = {
     "traceroot.span.input",
@@ -474,8 +482,19 @@ def transform_otel_to_clickhouse(
                         # real usage still lands on an LLM .doGenerate child.
                         input_token_keys += ["ai.usage.inputTokens", "ai.usage.promptTokens"]
                         output_token_keys += ["ai.usage.outputTokens", "ai.usage.completionTokens"]
-                    api_input_tokens = first_present_number(span_attrs, input_token_keys)
-                    api_output_tokens = first_present_number(span_attrs, output_token_keys)
+                    accept_token_usage = not _scope_skips_token_usage(
+                        scope_name, span_kind, span_attrs
+                    )
+                    api_input_tokens = (
+                        first_present_number(span_attrs, input_token_keys)
+                        if accept_token_usage
+                        else None
+                    )
+                    api_output_tokens = (
+                        first_present_number(span_attrs, output_token_keys)
+                        if accept_token_usage
+                        else None
+                    )
                     # Cache buckets. The OpenInference keys (prompt_details.*) are the
                     # verified path for Anthropic/OpenAI and MUST be listed first —
                     # they are the same family as llm.token_count.prompt (read above).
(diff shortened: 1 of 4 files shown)


ka1kqi:
Thanks for this — closing it, but I want to be precise about why, because it isn't simply "duplicate".

The fix that merged in #1612 keys the aggregate check on `gen_ai.operation.name` ∈ {`invoke_agent`, `agent_step`} rather than on the tracer scope. That is a **different rule from this PR, not a superset of it** — and this PR's control test, `test_other_scope_non_llm_usage_remains_authoritative`, asserts the opposite: that a span with `gen_ai.operation.name: invoke_agent` under a non-`gen_ai` scope stays priced at 11/7. The merged version deliberately skips exactly that span. So closing this is a design decision, and you should know what it was.

The reason is that the instrumentation scope is **chosen by the emitting application**. An app that builds a private `NodeTracerProvider` — the sensible shape when you don't want every HTTP span exported — names its own tracer, so its spans never arrive under `gen_ai`. Measured against a real app of that shape, on a trace whose true usage is 10 input tokens:

| | stored |
|---|---|
| `main` before the fix | 20 — double counted |
| scope-fenced approach | **20 — still double counted** |
| operation-name approach | 10 — correct |

`invoke_agent` and `agent_step` are semconv-defined names meaning "this span restates its descendants", so keying on them is both narrower in intent and broader in coverage than keying on scope. The scope check survives in the merged version only as a fallback for a wrapper arriving with no operation name at all, and it now has a test pinning it.

Two things from this PR that were better than what I first credited:

- **The `agent_step` case.** Your instinct to handle more than just the operation root was right; the merged fix covers it.
- **Threading the guard into the text-estimation branch.** I initially dismissed that as redundant, because under a scope-and-kind predicate the estimation branch was already unreachable. Once the predicate became kind-independent it stopped being redundant — an LLM-kind aggregate carrying the conversation text fell straight into estimation and had tokens *fabricated* from it, which is worse than the double count. You had guarded that; I had to add it back after a review caught it. Credit where it's due.

Your parametrized fixture shape — raw export versus OpenInference-labelled root — was also good coverage, and the merged tests cover both paths.

Closing as superseded by #1612.

PR #1616, opened 2026-07-25
fix(worker): skip gen_ai scope token extraction on non-LLM spans

### LINKED ISSUE #1610: fix(worker): Vercel AI SDK v7 semconv telemetry double-counts tokens and cost on operation root spans

AkarshBandi:
Closes #1610

**Problem**
The Vercel AI SDK v7 (`@ai-sdk/otel`, scope `"gen_ai"`) stamps aggregate
`gen_ai.usage.*` attributes on both the operation root span (AGENT kind)
and its inference children (LLM kind). The token extraction pipeline reads
these attributes on any span kind, so the root's aggregate is priced on
top of the children — producing 2x token counts and cost.

**Root cause**
The existing guard for the legacy `"ai"`-scope emitter (#1187) restricts
raw `ai.usage.*` keys to LLM spans, but no equivalent guard existed for
the semconv `"gen_ai"` scope, where `gen_ai.usage.*` is read on any span
kind.

**Fix**
When scope is `"gen_ai"` and span kind is not LLM, null out the API token
values. Neither the API-token branch nor the text-estimation branch fires.
Model name is still extracted. This mirrors the #1187 pattern exactly.

**Verification**
- Before: AGENT root with `gen_ai.usage.input_tokens=100, output_tokens=50`
  → 150 phantom tokens added to the trace total
- After: same span → no tokens extracted, model name preserved
- LLM children with same scope → tokens extracted normally
- Other scopes (openinference, ai, pydantic-ai) → unaffected
- 10 new regression tests covering AGENT/LLM/TOOL/CHAIN/embeddings
- **896 existing tests pass, 0 regressions**

---
## Summary by cubic
Fixes double-counted tokens and cost for Vercel AI SDK v7 (`@ai-sdk/otel`, scope `gen_ai`) by skipping token extraction on non-LLM spans. LLM spans still extract tokens; model names are preserved.

- **Bug Fixes**
  - Skip `gen_ai.usage.*` extraction when span kind is not LLM (AGENT/TOOL/CHAIN) to avoid charging aggregates on top of children.
  - Keep API token counts and cache details for LLM spans; model name still extracted for all spans.
  - Register `gen_ai` as a known scope to suppress unknown-scope warnings.
  - Add regression tests for AGENT/LLM/TOOL/CHAIN/embeddings.

<sup>Written for commit 5578f1946f47512f6c936f234e8c116a54a4c3e8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1616?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/worker/otel_transform.py
+++ b/backend/worker/otel_transform.py
@@ -537,7 +537,8 @@ def transform_otel_to_clickhouse(
                             "ai.usage.reasoningTokens",
                         ],
                     )
-
+                    if scope_name == "gen_ai" and span_kind != SpanKind.LLM:
+                        api_input_tokens = api_output_tokens = None
                     if api_input_tokens is not None or api_output_tokens is not None:
                         # Use API-provided counts (accurate).
                         input_tokens = int_or_zero(api_input_tokens)
--- a/backend/worker/tokens/buckets.py
+++ b/backend/worker/tokens/buckets.py
@@ -73,7 +73,8 @@ _KNOWN_SCOPE_PREFIXES: tuple[str, ...] = (
 # "ai" prefix would also swallow unrelated ai*-named scopes).
 _KNOWN_SCOPE_EXACT: frozenset[str] = frozenset(
     {
-        "ai",  # Vercel AI SDK tracer; GROSS input with cache detail under ai.usage.*
+        "ai", # Vercel AI SDK tracer; GROSS input with cache detail under ai.usage.*
+        "gen_ai",  
     }
 )
 
--- a/tests/worker/test_otel_transform_genai_semconv.py
+++ b/tests/worker/test_otel_transform_genai_semconv.py
@@ -1,0 +1,318 @@
+"""Regression tests for Vercel AI SDK v7 GenAI semconv telemetry.
+
+The v7 ``@ai-sdk/otel`` emitter (instrumentation scope ``"gen_ai"``) stamps
+aggregate ``gen_ai.usage.*`` attributes on BOTH the operation root span (AGENT)
+and its inference children (LLM). The root span's counts restate the sum of its
+LLM children. Extracting and pricing both produces 2x token totals and cost.
+
+The fix: when scope is ``"gen_ai"`` and the span kind is not LLM, skip token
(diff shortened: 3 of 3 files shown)


ka1kqi:
Thanks for this — closing it as superseded by #1612, which merged. A couple of specifics worth passing on, because you found something real here.

**What the merged fix does differently.** It keys the aggregate check on `gen_ai.operation.name` ∈ {`invoke_agent`, `agent_step`} rather than on the tracer scope, and it also maps `invoke_agent` to the AGENT span kind. Both matter:

- Scope is **chosen by the emitting application** — anyone using a private tracer provider names their own tracer, so their spans never arrive under `gen_ai` and a scope-keyed check silently misses them. Measured on a real app of that shape: `main` stored 20 for a true 10, the scope-keyed approach also stored 20, the operation-name approach stored 10.
- Without the classification change, the gate depends on the OpenInference processor having labelled the root as AGENT. A raw `@ai-sdk/otel` export has no `openinference.span.kind`, so the root infers **LLM** from its model attribute and passes straight through a kind-based check.

**Something your implementation got right that the title doesn't convey.** The description says "skip gen_ai scope token extraction", but the code nulls the *extracted values* after extraction, which is the correct shape. That distinction is load-bearing: on this path spans carry **both** namespaces — the emitter's `gen_ai.usage.*` and the OpenInference processor's `llm.token_count.*` — and the key list reads `llm.token_count.*` first, so the value actually being double-priced was the OpenInference-mapped one. A literal implementation of "skip the `gen_ai` namespace" would have been a no-op against real data. Your code avoided that; the title would have led someone else into it.

Your regression test file was thorough, and the merged tests cover the same shapes.

PR #1640, opened 2026-07-29
feat(db): add metadata-only source column to spans and traces

ka1kqi:
Adds `source LowCardinality(String) DEFAULT 'user'` to `spans` and `traces`, distinguishing detector self-traces from customer traffic.

Metadata-only: `source` sits outside every ORDER BY sort key, PARTITION key and PRIMARY KEY, so the `ADD COLUMN` needs no table rewrite and the `DEFAULT` backfills existing rows for free. A test asserts that property over the migration files, so a future key reorder that would turn this into a rewrite trips a test rather than a production surprise.

**Duplicated deliberately.** The same migration file also exists on the self-tracing stack. It is byte-identical, `ADD COLUMN IF NOT EXISTS` makes the double-apply a no-op, and goose keys applied migrations by version number, so whichever lands second is skipped. This copy exists because the dashboard read paths live on `main` and cannot reach the stack's copy.

If the self-tracing stack merges to `main` before this does, this PR's diff becomes empty — **close it rather than merge it.**

**Note for anyone adding a migration:** 006 is claimed by two branch families and 007 by the self-tracing stack, so the next migration from any branch must be **008**. goose panics on a duplicate version rather than warning, and `goose validate` does not catch it, so a collision surfaces as a failed pre-upgrade hook blocking the release.

---
## Summary by cubic
Add a metadata-only `source` column to ClickHouse `spans` and `traces` to label detector self-traces vs customer traffic. Defaults to `'user'` and backfills existing rows without a table rewrite.

- **Migration**
  - Adds `source LowCardinality(String) DEFAULT 'user'` via `ADD COLUMN IF NOT EXISTS` on both tables.
  - Stays outside ORDER BY, PARTITION BY, and PRIMARY KEY; a test enforces this and checks Up/Down.
  - Migration file is duplicated on the self-tracing stack; `goose` will skip whichever lands second. If that merges first, close this PR. Next migration version must be `008`.

<sup>Written for commit 133521cfdd75321480a5129df3d56d1a96e10f0a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1640?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/db/clickhouse/migrations/006_add_source_column.sql
+++ b/backend/db/clickhouse/migrations/006_add_source_column.sql
@@ -1,0 +1,12 @@
+-- +goose Up
+-- Distinguishes detector self-traces from customer traffic. DEFAULT 'user'
+-- backfills existing rows with no table rewrite: source is outside both the
+-- ORDER BY sort key and the PARTITION key on spans and traces (and outside
+-- the spans projection's ORDER BY), so this ADD COLUMN is metadata-only.
+-- LowCardinality keeps the low-distinct-value column cheap.
+ALTER TABLE spans  ADD COLUMN IF NOT EXISTS source LowCardinality(String) DEFAULT 'user';
+ALTER TABLE traces ADD COLUMN IF NOT EXISTS source LowCardinality(String) DEFAULT 'user';
+
+-- +goose Down
+ALTER TABLE spans  DROP COLUMN IF EXISTS source;
+ALTER TABLE traces DROP COLUMN IF EXISTS source;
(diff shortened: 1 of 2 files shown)


ka1kqi:
Unnecessary — duplicate of migration 006, which already ships with the self-tracing stack. This copy only existed so the dashboard fix could land before the stack merged. #1641 now targets main and merges once 006 is there.

Issue #1649, opened 2026-07-30
Preserve `EVALUATION`/`TASK`/`SCORER` span kinds and classify evaluation traces on ingest


dark-sorceror:
## Summary

Make evaluation traces first-class in the ingest pipeline (`backend/worker`, ClickHouse). Each result is backed by a real trace shaped `evaluation-item → task → scorer`; ingestion must keep those span kinds, tag the trace at its root so the surface can find it, and order sibling spans deterministically so a trace renders identically every time. The root tag is also what lets a later issue exclude these traces from the default list and from detector runs.

## Scope

- Carry the traceroot span kind through `otel_transform` into the `span_kind` column, and set `cost` on model-call leaf spans from provider usage.
- Persist `environment` and mark the evaluation trace at its root span so it is identifiable downstream and can be filtered as a group.
- Deterministic sibling ordering (start → end → span id), including for reconstructed evaluation traces.
- Preserve the `evaluation-item → task → scorer` nesting so a result's trace reconstructs with the scorer subtree distinguishable from the candidate task.

## Acceptance criteria

- `EVALUATION` / `TASK` / `SCORER` span kinds survive ingestion and are queryable by kind.
- An evaluation trace is never indistinguishable from an ordinary trace: its root always carries the classification.
- The evaluation trace is identifiable at its root, and `environment` is persisted alongside it.
- The `evaluation-item → task → scorer` hierarchy is intact so the scorer subtree can be told apart from the candidate task.
- A given trace's spans render in a stable order across reloads, siblings ordered by (start, end, span id).
- Model-call leaf spans carry a `cost` where provider usage is present, and none is written where usage is absent.

## Out of scope

- Excluding evaluation traces from the default list / detectors (a sibling issue).
- Deriving per-result cost from the trace (its own issue).

dark-sorceror:
Delivered with the offline-evaluation epic (#1921, merged to main).

PR #1710, opened 2026-07-30
feat(traces): preserve eval span kinds and classify eval traces

dark-sorceror:
## Summary

Fixes: #1649

Makes evaluation traces first-class in the ingest pipeline. Each result is backed by a real trace shaped `evaluation-item → task → scorer`; ingestion now preserves those span kinds, classifies the trace at its root so downstream surfaces can find and group it, and renders sibling spans in a deterministic order so a trace reads identically every time. The root classification is also what lets a sibling change exclude these traces from the default list and from detector runs.

## How it works

```
EVALUATION (root)   ← classifies the trace: environment = "evaluation"
  └─ TASK           ← the candidate task
       └─ SCORER    ← one per scorer that ran
```

The SDK emits the `EVALUATION`/`TASK`/`SCORER` span kinds but does not stamp `traceroot.environment`; ingestion derives `environment = "evaluation"` from the eval span kinds and carries it from the root onto the trace record, so an evaluation trace is never indistinguishable from an ordinary one.

### Backend
- `backend/shared/enums.py` — adds `EVALUATION`, `TASK`, and `SCORER` to the `SpanKind` enum.
- `backend/worker/otel_transform.py` — `get_span_kind` preserves the SDK's `EVALUATION`/`TASK`/`SCORER` kinds through into the `span_kind` column instead of coercing them to `SPAN`; `transform_otel_to_clickhouse` classifies a span with an eval kind as `environment = "evaluation"` when the SDK left `traceroot.environment` unset, and the root eval span carries that classification onto the trace record.

### Frontend
- `frontend/packages/core/src/constants.ts` — mirrors the `EVALUATION`/`TASK`/`SCORER` additions on the shared `SpanKind` constant.
- `frontend/ui/src/features/traces/utils/index.ts` — `compareSpansForStableDisplay`, the deterministic sibling comparator: orders spans sharing a parent by (start, end, span id) so ms-precision ties fall back to span id rather than arrival order.
- `frontend/ui/src/features/traces/utils/timeline.ts` — sorts each parent's children and the top-level roots with `compareSpansForStableDisplay` when building the render tree, including for reconstructed evaluation traces.

### Tests
- `tests/worker/test_otel_transform.py` — `EVALUATION`/`TASK`/`SCORER` survive `get_span_kind`; an `EVALUATION` root classifies the trace and span as `environment = "evaluation"` without a stamped `traceroot.environment`.
- `tests/fixtures/otel_payloads.py` — evaluation-trace fixtures mirroring the real attribute keys the SDK stamps on an eval run's spans (`eval_root_attributes` and the item → task → scorer nesting).
- `frontend/ui/src/features/traces/utils/index.test.ts` — the stable sibling ordering, including the ms-tie fallback to span id.

## Test plan

- [ ] Ingest an evaluation trace and confirm `EVALUATION`/`TASK`/`SCORER` are stored in `span_kind` and queryable by kind.
- [ ] Confirm the root span carries the classification and that `environment = "evaluation"` is persisted on the trace even when the SDK did not stamp it.
- [ ] Confirm the `evaluation-item → task → scorer` nesting is intact so the scorer subtree can be told apart from the candidate task.
- [ ] Render a trace with sub-ms-parallel siblings twice and confirm identical ordering, siblings ordered by (start, end, span id).
- [ ] Confirm a model-call leaf span carries a `cost` where provider usage is present and none where usage is absent.


ka1kqi:
**Recommendation: classify eval traces from span kind, not `environment`.**

The current mechanism holds only under specific conditions. The eval engine writes `traceroot.environment` (default `'evaluation'`) on the root span after the processor's `onStart`, and the trace row takes environment from the root only — so defaults classify correctly. But three things sit under it:

1. It rides a **write-ordering guarantee** (engine-after-processor, last-write-wins on the root) that no test pins in either repo — across a version boundary where the processor starts writing the same attribute from global config.
2. `evaluate()` exposes `environment=` as a first-class option. Any caller passing a real value — `'staging'` — silently un-classifies the run: traces enter the default list, detectors fire on them. Normal use of the API defeats the feature.
3. A customer legitimately using `environment="evaluation"` for real traffic gets those traces hidden **and detectors silently skipped** — the missing-alerts half is undiagnosable from outside.

Both SDKs already tag the eval root `traceroot.span.type = 'evaluation'` and assert it in their own tests (the Python test comments it "authoritative eval classifier"). Classification should read that marker; `environment` should mean only what the user set.

## Changes, all in this repo

- **Migration**: `ALTER TABLE traces ADD COLUMN root_span_kind LowCardinality(String) DEFAULT ''` — metadata-only, no rewrite. Numbering: two unmerged stacks currently both claim 006 (`006_add_source_column` vs `006_create_public_sql_views`), and goose **panics** on duplicate versions — confirm the free slot at merge time rather than assuming 008.
- **`otel_transform.py`**: delete the `environment is None` derivation; promote `span_kind` in the existing root-upgrade block (the one already promoting root name/input/output onto the trace record).
- **`client.py`**: add the column to the traces insert as `t.get("root_span_kind") or ""`. The `or ""` is load-bearing: the shallow child-first placeholder record has no such key, and the driver rejects `None` for a non-Nullable column — aborting the entire batch, not one row.
- **Transform tests**: none exist for the classification today. Pin the exact stored literal (`'EVALUATION'`, post-uppercase) on the promotion side — the filter is a string equality across two codebases, so the value contract needs a test on each side.

Verified solid as-built, no change needed: the child-before-root placeholder cannot clobber classification later (the traces writer checks ClickHouse and drops shallow records for traces that already exist), and the reader's `argMax` dedup handles placeholder ordering correctly.

No SDK changes required — the marker this reads is already emitted and tested. Optional cleanup later: drop the SDK-side `environment` default of `'evaluation'`, so eval runs stop injecting a value into the customer's environment dimension.

## Why a new column — and why the ways to avoid one don't work as well

`root_span_kind` adds no new information to the schema. `span_kind` already exists on every span; this is a denormalized copy of the root's value into the table the list actually scans. `traces` is already exactly that kind of projection — `name`, `input`, `output`, `user_id`, `session_id` are all promoted from the root span because the list query cannot afford to join `spans`. This is one more field in an existing pattern, and it generalizes: any future trace kind that wants the same default-hidden treatment is a new *value*, not a new column.

The alternatives, and where each loses:

- **Keep overloading `environment`** — that is the current design, and the three failure modes above are its scorecard. Classification keyed off a user-owned dimension inherits every way users legitimately use that dimension.
- **Add an `'eval'` value to `source`** — `source` is the provenance boundary: route-stamped, unforgeable, and consumed as a fail-closed `source = 'user'` equality at every customer-facing read (lists, sessions, users, dropdowns, widgets). Eval traces *are* customer traffic and must stay visible in nearly all of those surfaces, so an `'eval'` value would hide them everywhere at once and then have to be walked back site by site — converting a single-equality guarantee into a per-site allowlist. It would also mix an unforgeable value and a forgeable one in the same column, so "`source` is authoritative" would stop being uniformly true.
- **No column at all — semi-join at read time** (`trace_id NOT IN (SELECT trace_id FROM spans WHERE span_kind = 'EVALUATION' AND parent_span_id IS NULL …)`) — workable, but it makes every default list load pay a spans scan forever to exclude a rare row class. The column pays once, at write time, in metadata. Same trade the existing root promotions already made, in the same direction.
- **Container separation** (each eval run gets its own project/container) — the strongest isolation, and the right investment if evals grow full experiment semantics later. But it costs an experiment object model, SDK routing at init time, and cross-container UX for "open the underlying trace", to satisfy requirements that are currently just: hide from the default list, skip detectors, opt-in flag.

Net schema shape after this: one enforced provenance axis (`source`, fail-closed — unknown markers hidden), one derived semantics axis (`root_span_kind`, fail-open — unknown kinds visible, which is the correct bias for a default list that should show everything not provably eval), and `environment` returned to being ordinary user data. The two axes wanting **opposite** failure polarities is the clearest sign they belong in separate columns.

ka1kqi:
**Decision: the environment-based classification as implemented here is fine to keep.** Eval instrumentation is always agent-driven, so the instructions layer owns the main failure mode — never pass `environment=` to `evaluate()` — and the column decoupling proposed above is not needed.

Two notes that go with that decision:

1. **Recommended hardening:** an SDK-side test pinning that the eval root's `traceroot.environment` survives a configured global environment. As of 0.2.0 the span processor stamps that key on every span from global config, so classification depends on the eval engine's root write landing after `onStart` (last-write-wins). That ordering is deterministic today but tested in neither repo — one test turns it into a contract.
2. **Accepted risk:** a project legitimately using `environment="evaluation"` for real traffic gets those traces hidden from the default list *and* skipped by detectors, with nothing on-screen explaining either. Accepted; worth a warning in the eval skill/docs.

The smaller items above stand independent of this choice: `run_on_evaluation` reads a field that doesn't exist on detector rows yet, and `traceroot.eval.*` isn't in the transform's known-attribute prefixes, so the identity attributes land raw in span metadata.

Issue #1650, opened 2026-07-30
Exclude evaluation traces from the default trace list and from detector runs


dark-sorceror:
## Summary

Evaluation runs should not pollute production trace views or alerting. Exclude evaluation traces from the default trace list and skip them in detector runs by default, applying the exclusion after dedup so it operates on resolved rows rather than raw duplicates. The filter keys off the root-level evaluation classification set on ingest, so it never has to re-derive eval membership span by span.

## Scope

- Filter evaluation traces out of the default traces-list query; opt-in inclusion is handled elsewhere and is not part of this issue.
- Make detector runs skip offline-evaluation traces by default, so alerting never fires on synthetic candidate output.
- Apply the exclusion after the dedup pass, so it never drops a legitimate non-eval trace.
- Key the exclusion off the root classification marker, not a per-span check, so partial or late ingests still resolve correctly.
- Keep the filter in the query layer so pagination totals reflect the visible non-eval set.

## Acceptance criteria

- The default trace list does not show evaluation traces.
- Detectors do not fire on evaluation traces by default.
- The exclusion runs post-dedup and leaves non-eval traces untouched.
- A project with only evaluation traces shows an empty (not broken) default list.
- The opt-in path that surfaces evaluation traces elsewhere still returns them unfiltered.
- Trace-list pagination totals count only visible non-eval traces.
- Detector scheduling excludes evaluation traces, so queue depth is not inflated by eval volume.

## Out of scope

- Root classification / span-kind preservation on ingest (its own issue) — this consumes those markers.
- Any UI toggle for showing evaluation traces in the trace list.

dark-sorceror:
Delivered with the offline-evaluation epic (#1921, merged to main).

PR #1711, opened 2026-07-30
feat(traces): exclude evaluation traces from lists and detectors

dark-sorceror:
## Summary

Fixes: #1650

Evaluation runs should never pollute production trace views or alerting. This excludes offline-evaluation traces from the default trace list and skips them in detector runs, keyed off the root-level `environment = "evaluation"` classification that ingest already sets — so nothing has to re-derive eval membership span by span. The list filter runs after the ReplacingMergeTree dedup, and its pagination count uses the same predicate, so totals reflect only the visible non-eval set.

## How it works

```
list query
  → dedup ReplacingMergeTree rows to the latest version per trace
    → drop rows where environment = 'evaluation' (NULL-safe: untagged prod kept)
      → page + count over the surviving non-eval rows
```

The exclusion is deliberately applied post-dedup, never to the raw rows: an eval trace can have a shallow early row (before the root's `environment` lands) alongside the full root row, so filtering raw rows would leak the shallow copy. Filtering the deduped latest row avoids that and keeps every legitimate non-eval trace.

### Backend (Python)
- `backend/rest/services/trace_reader.py` — `include_evaluations` flag on the traces-list query; excludes `environment = 'evaluation'` traces after the dedup pass (NULL-safe, keeping untagged production traces), and applies the identical predicate to the total-count query so pagination totals match the visible set.
- `backend/rest/routers/traces.py` — exposes an `include_evaluations` query param (default `false`) on the internal traces-list route.
- `backend/rest/routers/public/traces_read.py` — exposes the same `include_evaluations` param (default `false`) on the public traces-read route.
- `backend/worker/detector_tasks.py` — `_detector_runs_on_trace` skips traces classified `environment == "evaluation"` by default (opt-in reserved via a detector's `run_on_evaluation` flag), so detector scheduling never fires on synthetic candidate output and eval volume never inflates queue depth.

### Tests
- `backend/worker/tests/test_detector_tasks.py` — evaluation traces skipped by default; opt-in path runs.
- `tests/rest/test_trace_reader.py` — default exclusion, post-dedup shard guard, and opt-in inclusion.

## Test plan
- [ ] Confirm the default trace list does not show evaluation traces.
- [ ] Confirm detectors do not fire on evaluation traces by default.
- [ ] Verify the exclusion runs post-dedup and leaves non-eval traces untouched.
- [ ] A project with only evaluation traces shows an empty (not broken) default list.
- [ ] The opt-in path (`include_evaluations=true`) returns evaluation traces unfiltered.
- [ ] Trace-list pagination totals count only visible non-eval traces.
- [ ] Detector scheduling excludes evaluation traces, so queue depth is not inflated by eval volume.


ka1kqi:
**Consumer-side changes for span-kind classification** (pairs with the comment on #1710):

- **`trace_reader.py`**: predicate becomes `root_span_kind != 'EVALUATION'` on the deduped row. The post-dedup placement, the `argMax` count rewrite and `include_evaluations` defaulting off are correct as written and stay. Fail-open polarity is deliberate here: the default list should show everything not provably eval, so an unknown future kind surfaces rather than vanishing — the opposite bias from `source = 'user'`, which is a security boundary and stays fail-closed.
- **`detector_tasks.py`**: `_get_trace_summaries` reads `FROM spans`, not `traces`, so the new traces column is not the input here — the edit is `anyIf(span_kind, parent_span_id IS NULL) AS root_span_kind`, spans-side, no migration needed for this half. **Keep `environment` in that SELECT**: it is a user-facing detector trigger field consumed by `_eval_condition`, and removing it would silently kill every `environment = <x>` trigger. Nothing currently tests that — the unit tests monkeypatch `_get_trace_summaries` entirely — so add one test asserting the query's SELECT list, the same pattern used elsewhere to pin a column that assertions on the result dict cannot see.
- **`test_detector_tasks.py`** (already in this PR): both `_detector_runs_on_trace` tests hard-code `{"environment": "evaluation"}` and flip with the mechanism; move them to the new field.
- **Behavioural test gap** worth closing in the same pass: nothing in `tests/` exercises `include_evaluations` or the default exclusion end-to-end — line coverage passes via existing list tests, but the behaviour itself is unasserted.

The trace-list exclusion (page + count, which must agree) is deliberately the **only** consumer of `root_span_kind`: sessions, users, dropdowns and widgets keep showing eval traces because they are the customer's own data, and hiding them there would make the dashboard disagree with the bill. By-id reads don't filter on it, so opening an eval trace from a run detail needs no opt-in.

ka1kqi:
Per the decision on #1710: the environment-keyed skip here stays as written. See the note there for the recommended SDK ordering test and the accepted collision risk.

PR #1762, opened 2026-07-31
feat(rest): publish reporting OpenAPI with a parity guard

### LINKED ISSUE #1648: Publish the reporting OpenAPI artifact + a Zod↔Pydantic parity and drift guard
## Summary

Publish the public eval reporting API as a machine-readable OpenAPI artifact so a client can be generated from it, and keep the two contract layers — Zod in the control plane and Pydantic in the gateway — provably in lockstep. Without the parity + drift guards, the gateway could accept requests the control plane rejects (or vice-versa) and no test would catch it. The published artifact covers the full reporting surface — register, upsert, complete, and dataset sync — so a generated client always matches the server.

## Scope

- Explicit typed FastAPI routes for the reporting surface, with Pydantic models mirroring the Zod schemas field-for-field, including the scorer definition and its message shape.
- Match strictness across layers: the mirror is strict where Zod is strict and ignores extras where Zod strips, so accept/reject verdicts line up.
- Build and publish `backend/rest/openapi/public.json`, with a check that fails if it drifts from the models.
- A cross-layer parity test that feeds identical payloads to both the Zod and Pydantic layers and asserts the same accept/reject verdicts.
- Regenerate the schema from the models in the check so a hand-edited or stale `public.json` fails rather than silently diverging.

## Acceptance criteria

- `public.json` is generated from the models and committed; the drift check fails when it is stale.
- The Zod and Pydantic layers agree on every representative valid/invalid payload, including strict-vs-ignore behavior on unknown keys.
- Parity fixtures include a bare `{ name, version }` scorer and a full LLM-judge / code definition, and both layers accept each identically.
- Editing one contract layer without the other turns a test red.
- The published schema is usable as a codegen input for a client and stays in step with the served routes.

## Out of scope

- The gateway itself (a separate issue); this issue documents and guards its contract.
- Generating or shipping an actual client SDK from the artifact.

dark-sorceror:
## Summary

Fixes: #1648

Publishes the public eval reporting API as a machine-readable OpenAPI artifact so a client can be generated from it, and keeps the two contract layers — Zod in the control plane and Pydantic in the gateway — provably in lockstep. Without the parity and drift guards, the gateway could accept a request the control plane rejects (or vice-versa) and no test would catch it; with them, editing one layer without the other turns a test red.

### Contract
- `backend/rest/schemas/eval.py` — the Pydantic mirror of the Zod contract, field-for-field: `RegisterRunRequest`, `UpsertResultRequest`, `CompleteRunRequest` and their responses, `ScorerRef` (including the definition fields and `ScorerMessage`), `ScoreInput`, and the shared `ErrorResponse` envelope. Field constraints mirror the Zod validators, but the request models are deliberately **not** `extra="forbid"` even though the Zod schemas are `.strict()`: these run in the *gateway*, which forwards bodies verbatim to the Next.js handler that persists them. A `forbid` here would 422 a field the gateway does not yet model so it never reaches persistence, and would break rolling deploys by rejecting a newer SDK's optional field on every older gateway instance — and the field constraints (`min_length`/`max_length`, `ge=0`, nullability, defaults) mirror the Zod validators, so accept/reject verdicts line up across layers.

### Tests
- `tests/rest/test_eval_contract_parity.py` — the Pydantic side of the cross-layer check: feeds the representative payloads to the Pydantic models and asserts the expected accept/reject verdicts.
- `frontend/packages/core/src/__tests__/eval-contract-parity.drift.test.ts` — the Zod side: feeds the same payloads to the Zod schemas and asserts identical verdicts.
- `frontend/packages/core/src/__tests__/eval-contract-parity-fixtures.json` — the shared fixture set both layers consume, including a bare `{ name, version }` scorer and a full LLM-judge / code definition, plus strict-vs-ignore unknown-key cases.
- `tests/rest/test_public_openapi.py` — the drift guard: regenerates the schema from the models and fails if `public.json` is stale or hand-edited.
- `tests/rest/test_public_eval_gateway.py` — Tests for the public offline-eval API gateway.
- `tests/rest/test_public_eval_gateway_security.py` — Security tests for the public offline-eval gateway (a reverse proxy).

### Frontend
- `frontend/packages/core/src/__tests__/contract-shape.ts` — Structural shape extraction for the cross-layer contract guard (Zod side).
- `frontend/packages/core/src/__tests__/eval-contract-shape.json` — part of this change.
- `frontend/packages/core/src/eval-contract.ts` — Offline-evaluation contract — the shared, typed shapes for both the user-session CRUD routes and the API-key SDK reporting routes.

## Test plan

- [ ] Regenerate `public.json` from the models and confirm it matches the committed file; hand-edit it and confirm the drift check fails.
- [ ] Run both parity suites and confirm the Zod and Pydantic layers agree on every representative valid/invalid payload, including unknown-key strict-vs-ignore behavior.
- [ ] Confirm the fixtures include a bare `{ name, version }` scorer and a full LLM-judge / code definition, and that both layers accept each identically.
- [ ] Change one contract layer without the other and confirm a parity test turns red.
- [ ] Feed `public.json` to a client generator and confirm it produces a client that matches the served routes.

### DIFF
--- a/backend/rest/openapi/public.json
+++ b/backend/rest/openapi/public.json
@@ -984,6 +984,16 @@
             ],
             "title": "Metadata"
           },
+          "provenance": {
+            "anyOf": [
+              {
+                "$ref": "#/components/schemas/RunProvenance"
+              },
+              {
+                "type": "null"
+              }
+            ]
+          },
           "scorers": {
             "items": {
               "$ref": "#/components/schemas/ScorerRef"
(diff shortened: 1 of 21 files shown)


dark-sorceror:
@cubic-dev-ai review

XinweiHe (approved):

PR #1763, opened 2026-07-31
feat(traces): preserve eval span kinds and classify eval traces

### LINKED ISSUE #1649: Preserve `EVALUATION`/`TASK`/`SCORER` span kinds and classify evaluation traces on ingest

dark-sorceror:
## Summary

Fixes: #1649

Makes evaluation traces first-class in the ingest pipeline. Each result is backed by a real trace shaped `evaluation-item → task → scorer`; ingestion now preserves those span kinds, classifies the trace at its root so downstream surfaces can find and group it, and renders sibling spans in a deterministic order so a trace reads identically every time. The root classification is also what lets a sibling change exclude these traces from the default list and from detector runs.

## How it works

```
EVALUATION (root)   ← classifies the trace: is_evaluation = 1
  └─ TASK           ← the candidate task
       └─ SCORER    ← one per scorer that ran
```

The SDK emits the `EVALUATION`/`TASK`/`SCORER` span kinds but does not stamp `traceroot.environment`; ingestion sets a dedicated `is_evaluation` flag from the eval span kinds and carries it onto the trace record, so an evaluation trace is never indistinguishable from an ordinary one. The flag is deliberately separate from `environment`, which is the customer's own free-text deployment tag — overloading it would make a team that names a stack "evaluation" lose its entire trace list.

### Backend
- `backend/shared/enums.py` — adds `EVALUATION`, `TASK`, and `SCORER` to the `SpanKind` enum.
- `backend/worker/otel_transform.py` — `get_span_kind` preserves the SDK's `EVALUATION`/`TASK`/`SCORER` kinds through into the `span_kind` column instead of coercing them to `SPAN`; `transform_otel_to_clickhouse` flags a span with an eval kind as `is_evaluation`, and any eval-kind span in the batch flags the trace — the flag only ever goes 0 → 1, so a later batch of non-eval leaves cannot un-flag it.
- `backend/db/clickhouse/client.py` — ClickHouse client using clickhouse-connect.
- `backend/worker/ingest_tasks.py` — Celery task definitions. This module defines the async tasks that process trace data from S3 to ClickHouse.

### Tests
- `tests/db/test_client.py` — Unit tests for ClickHouseClient row-building logic. Tests row construction without a real ClickHouse connection.
- `tests/worker/test_eval_result_cost.py` — Per-result cost derived from an evaluation trace must measure the CANDIDATE only.
- `tests/worker/test_eval_trace_classification.py` — Offline-evaluation traces must be classified on ingest by a dedicated signal.
- `tests/worker/test_ingest_tasks.py` — Unit tests for Celery task logic with mocked S3 + ClickHouse.

## Test plan

- [ ] Ingest an evaluation trace and confirm `EVALUATION`/`TASK`/`SCORER` are stored in `span_kind` and queryable by kind.
- [ ] Confirm the root span carries the classification and that `environment = "evaluation"` is persisted on the trace even when the SDK did not stamp it.
- [ ] Confirm the `evaluation-item → task → scorer` nesting is intact so the scorer subtree can be told apart from the candidate task.
- [ ] Render a trace with sub-ms-parallel siblings twice and confirm identical ordering, siblings ordered by (start, end, span id).
- [ ] Confirm a model-call leaf span carries a `cost` where provider usage is present and none where usage is absent.

### DIFF
--- a/backend/db/clickhouse/client.py
+++ b/backend/db/clickhouse/client.py
@@ -68,6 +68,7 @@ class ClickHouseClient:
                     now,  # ch_create_time
                     now,  # ch_update_time
                     t.get("environment"),
+                    1 if t.get("is_evaluation") else 0,
                 ]
             )
 
@@ -90,6 +91,7 @@ class ClickHouseClient:
                 "ch_create_time",
                 "ch_update_time",
                 "environment",
+                "is_evaluation",
             ],
         )
 
@@ -135,6 +137,7 @@ class ClickHouseClient:
                     now,  # ch_create_time
                     now,  # ch_update_time
                     s.get("environment"),
+                    1 if s.get("is_evaluation") else 0,
                 ]
             )
 
@@ -168,6 +171,7 @@ class ClickHouseClient:
                 "ch_create_time",
                 "ch_update_time",
                 "environment",
+                "is_evaluation",
             ],
         )
 
--- a/backend/shared/enums.py
+++ b/backend/shared/enums.py
@@ -13,6 +13,25 @@ class SpanKind(StrEnum):
     SCORER = "SCORER"
 
 
+# The span kinds that mark a span as part of an offline-evaluation run, and the source
+# of truth for the trace-level `is_evaluation` flag written at ingest.
+#
+# This is the only trustworthy signal: it comes from the SDK's eval engine
+# (traceroot.span.type), never from user-supplied configuration. In particular it must
+# not be conflated with `environment`, which is the user's own deployment tag
+# (TRACEROOT_ENVIRONMENT: production, staging, ...). Deriving classification from
+# `environment` both misses every eval run in a project that sets the tag and
+# misclassifies a customer who happens to name their environment "evaluation".
+#
+# Note this membership answers "is this span part of an eval run", which is a DIFFERENT
+# question from "does this span count toward the candidate's cost" — the latter needs a
+# subtree walk that drops SCORER spans and everything under them (see
+# worker.ingest_tasks._task_cost_by_trace). Do not collapse the two.
+EVALUATION_SPAN_KINDS = frozenset(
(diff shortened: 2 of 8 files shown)


dark-sorceror:
@cubic-dev-ai review

XinweiHe (approved):

PR #1764, opened 2026-07-31
feat(traces): exclude evaluation traces from lists and detectors

### LINKED ISSUE #1650: Exclude evaluation traces from the default trace list and from detector runs

dark-sorceror:
## Summary

Fixes: #1650

Evaluation runs should never pollute production trace views or alerting. This excludes offline-evaluation traces from the default trace list and skips them in detector runs, keyed off the `is_evaluation` flag that ingest already sets — so nothing has to re-derive eval membership span by span. The list filter runs after the ReplacingMergeTree dedup, and its pagination count uses the same predicate, so totals reflect only the visible non-eval set.

## How it works

```
list query
  → dedup ReplacingMergeTree rows to the latest version per trace
    → drop traces with any is_evaluation = 1 row (UInt8 DEFAULT 0, never NULL)
      → page + count over the surviving non-eval rows
```

The exclusion is a `trace_id` set membership, not a predicate on the deduped row: ingest can only make the flag monotonic *within* a batch, so a later batch carrying only non-eval leaves rewrites the trace row with `is_evaluation = 0` and a newer `ch_update_time`. Reading the deduped latest row would un-hide the trace the moment such a batch landed. Any row anywhere flagged 1 hides the trace permanently, whatever order the writes arrived in — which also makes it safe in the shared where-clause, so the page and count queries agree by construction.

### Backend (Python)
- `backend/rest/services/trace_reader.py` — `include_evaluations` flag on the traces-list query; excludes `environment = 'evaluation'` traces after the dedup pass (NULL-safe, keeping untagged production traces), and applies the identical predicate to the total-count query so pagination totals match the visible set.
- `backend/rest/routers/traces.py` — exposes an `include_evaluations` query param (default `false`) on the internal traces-list route.
- `backend/rest/routers/public/traces_read.py` — exposes the same `include_evaluations` param (default `false`) on the public traces-read route.
- `backend/worker/detector_tasks.py` — `_detector_runs_on_trace` skips traces classified `environment == "evaluation"` by default (opt-in reserved via a detector's `run_on_evaluation` flag), so detector scheduling never fires on synthetic candidate output and eval volume never inflates queue depth.
- `backend/rest/openapi/public.json` — part of this change.
- `backend/rest/routers/sessions.py` — Session query endpoints (user-authenticated, not public API).
- `backend/rest/routers/users.py` — User query endpoints (user-authenticated, not public API).

### Tests
- `tests/rest/test_evaluation_exclusion_routers.py` — Every trace-derived endpoint must ask the reader to hide evaluation traces. The service-layer tests (``test_trace_reader_evaluations.py``) prove the SQL is right.
- `tests/rest/test_trace_reader_evaluations.py` — Offline-evaluation traces must be hidden from EVERY trace-derived read path. The SQL half of the feature.
- `tests/worker/test_detector_tasks.py` — Unit tests for the exactly-once detector enqueue path (Redis lock + BullMQ).

## Test plan
- [ ] Confirm the default trace list does not show evaluation traces.
- [ ] Confirm detectors do not fire on evaluation traces by default.
- [ ] Verify the exclusion runs post-dedup and leaves non-eval traces untouched.
- [ ] A project with only evaluation traces shows an empty (not broken) default list.
- [ ] The opt-in path (`include_evaluations=true`) returns evaluation traces unfiltered.
- [ ] Trace-list pagination totals count only visible non-eval traces.
- [ ] Detector scheduling excludes evaluation traces, so queue depth is not inflated by eval volume.

### DIFF
--- a/backend/rest/openapi/public.json
+++ b/backend/rest/openapi/public.json
@@ -2,12 +2,12 @@
   "components": {
     "schemas": {
       "CompleteRunRequest": {
-        "additionalProperties": false,
         "description": "Complete/fail a run, reporting final completeness counts.",
         "properties": {
           "case_count": {
             "anyOf": [
               {
+                "maximum": 9007199254740991.0,
                 "minimum": 0.0,
                 "type": "integer"
               },
@@ -31,6 +31,7 @@
           "scored_count": {
             "anyOf": [
               {
+                "maximum": 9007199254740991.0,
                 "minimum": 0.0,
                 "type": "integer"
               },
@@ -43,6 +44,7 @@
           "scorer_error_count": {
             "anyOf": [
               {
+                "maximum": 9007199254740991.0,
                 "minimum": 0.0,
                 "type": "integer"
               },
@@ -67,6 +69,7 @@
           "task_error_count": {
             "anyOf": [
               {
+                "maximum": 9007199254740991.0,
                 "minimum": 0.0,
                 "type": "integer"
               },
@@ -880,7 +883,6 @@
         "type": "object"
       },
       "RegisterRunRequest": {
-        "additionalProperties": false,
         "description": "Register/start a run. Idempotent on ``client_run_id`` within an evaluation.",
         "properties": {
           "baseline_run_id": {
@@ -905,6 +907,7 @@
           "case_count": {
             "anyOf": [
               {
+                "maximum": 9007199254740991.0,
                 "minimum": 0.0,
                 "type": "integer"
               },
@@ -988,6 +991,7 @@
             "items": {
               "$ref": "#/components/schemas/ScorerRef"
             },
+            "maxItems": 200,
             "title": "Scorers",
             "type": "array"
           }
@@ -1039,7 +1043,6 @@
         "type": "object"
       },
       "ScoreInput": {
-        "additionalProperties": false,
         "description": "One scorer's outcome on one result. ``error`` set = the scorer failed to judge.",
         "properties": {
           "bool_value": {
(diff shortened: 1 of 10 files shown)


dark-sorceror:
@cubic-dev-ai review

XinweiHe (approved):

PR #1769, opened 2026-07-31
feat(evaluations): pure candidate-vs-baseline comparison engine

### LINKED ISSUE #1653: Pure candidate-vs-baseline comparison engine (single source of truth + trustworthiness reasons)
## Summary

A single, pure, unit-testable engine (`lib/eval/comparison.ts`) is the source of truth for candidate-vs-baseline math and for whether a comparison is trustworthy. Both the read routes and the UI consume its output and neither recomputes it, so the whole product agrees on what "improved", "regressed", and "not comparable" mean. Outcomes and trustworthiness travel together in one payload so callers never re-derive either.

## Scope

- `compareRuns`: main-score delta, per-scorer cells, per-case outcomes (improved / regressed / unchanged / changed / unpaired), and duration.
- Classify each paired case by numeric main-score movement (improved / regressed / unchanged), fall back to changed for non-numeric scores, and mark unpaired when a case exists on only one side.
- Trustworthiness: `trustworthy` is true only when the comparison is available with no blocking reason — no baseline, different dataset version, different evaluation, non-terminal baseline, or a main scorer that can't be matched on any shared case.
- Emit the blocking reasons as a data list so callers render why a comparison is untrustworthy without re-deriving it.
- Per-scorer cells carry the candidate value, the baseline value, and their delta, so a scorer table renders without recomputation.
- Fold duration comparison (per-case, run-elapsed, and paired mean) into the engine output, averaging only cases present in both runs.

## Acceptance criteria

- Pure function with no I/O; unit-tested across the edge cases above.
- `trustworthy` is false with an explicit reason list whenever a blocking condition holds.
- Per-case outcomes and the aggregate counts are internally consistent.
- Given the same inputs it always returns the same result (deterministic).
- The paired-mean duration and paired counts reflect only cases matched on both sides.

## Out of scope

- Fetching runs/results (the read routes do that and pass data in).
- Any UI; this issue is the engine only.

dark-sorceror:
## Summary

Fixes: #1653

A single pure, unit-testable engine (`lib/eval/comparison.ts`) is the source of truth for candidate-vs-baseline math and for whether a comparison is trustworthy. Both the read routes and the UI consume its output and neither recomputes it, so the whole product agrees on what "improved", "regressed", and "not comparable" mean. Outcomes and trustworthiness travel together in one payload so callers never re-derive either.

### Tests
- `frontend/ui/src/lib/eval/comparison.test.ts` — no baseline, dataset-version mismatch, different evaluation, non-terminal baseline, missing/errored scores, main-scorer incompatibility, per-case outcome classification, and paired-mean duration over matched cases.
- `app/api/projects/[projectId]/evaluations/runs/[runId]/route.readback.test.ts` — covers `route.readback`.

## Test plan
- [ ] `compareRuns` is a pure function with no I/O.
- [ ] `trustworthy` is false with an explicit reason list whenever a blocking condition holds (no baseline, different dataset version, different evaluation, non-terminal baseline, unmatchable main scorer).
- [ ] Per-case outcomes and the aggregate case counts are internally consistent.
- [ ] Given the same inputs it always returns the same result.
- [ ] The paired-mean duration and paired counts reflect only cases matched on both sides.
- [ ] Per-scorer cells carry candidate value, baseline value, and delta so a scorer table renders without recomputation.

### DIFF
--- a/frontend/ui/src/app/api/projects/[projectId]/evaluations/runs/[runId]/route.readback.test.ts
+++ b/frontend/ui/src/app/api/projects/[projectId]/evaluations/runs/[runId]/route.readback.test.ts
@@ -191,10 +191,10 @@ describe("run/result read-back", () => {
     expect(run.baselineRunId).toBe("run0");
     expect(run.baselineComparable).toBe(true);
     // NOT the raw run.mainScore subtraction (0.9 - 0.75 = 0.15): case-2's candidate
-    // task errored (no routing-accuracy score), so it is excluded from the paired
+    // task errored (no routing-accuracy score), so it's excluded from the paired
     // aggregate. The only actually-comparable case is case-1 (candidate 1 vs
-    // baseline 0.5), so the trustworthy headline delta is 0.5 — derived the same way
-    // as every per-scorer aggregate, never a subtraction of the two runs' raw
+    // baseline 0.5), so the trustworthy headline delta is 0.5, derived the same way
+    // as every per-scorer aggregate — never a subtraction of the two runs' raw
     // SDK-reported aggregates, which can silently cover different case sets.
     expect(run.changeFromBaseline).toBeCloseTo(0.5);
   });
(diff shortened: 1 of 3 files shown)


dark-sorceror:
@cubic-dev-ai review

XinweiHe (approved):

PR #1771, opened 2026-07-31
feat(traces): open a result's eval-shaped trace with stepping

### LINKED ISSUE #1664: Open a result's real, eval-shaped trace led by the test case, with result stepping
## Summary

Opening a result shows its real execution trace, led by the test case's identity and status, reusing the trace viewer. Wire the evaluation↔trace linkage and the trace-to-dataset workflow, and let the user step between results with the trace panel's chevrons without re-animating the panel. Leading with the test case — rather than the raw root span — keeps the result's identity and pass/fail state at the top while the spans below explain how it got there.

## Scope

- Open the eval-shaped trace from a result; lead the panel with the test case (identity + status in the header, a copy button by the id) and strip span detail down to I/O.
- Step between results with up/down chevrons; reset the view on a run switch and don't re-animate on each step.
- Chevron stepping walks the run's results in table order, so the panel stays in sync with the list behind it.
- On a run switch the panel resets to the first result rather than holding the previous run's position.
- Fall back to the span's own I/O when the dedicated I/O fetch is empty.

## Acceptance criteria

- A result opens its real trace, led by the test case, with its status shown and its id copyable from the header.
- Result stepping works, walks results in table order, and does not re-animate the panel on each switch.
- The header exposes the test case id with a copy button beside it.
- The view resets when switching runs, clearing the previous run's stepping position.
- The eval-shaped trace reduces span rows to their I/O and leads with the test case.
- Span I/O falls back gracefully when the I/O fetch is empty.

## Out of scope

- Editing captured test cases from the trace panel — that happens on the dataset detail.
- Comparing two runs' traces side by side.
- Diff mode and I/O format switching (the trace-viewer issue).

dark-sorceror:
## Summary

Fixes: #1664

Opening a result from the run detail shows its real execution trace in the trace viewer, led by the test case's identity and status rather than the raw root span — so the result's identity and pass/fail state sit at the top while the spans below explain how it got there. The panel's up/down chevrons step between results in table order without re-animating the panel, and the view resets to the first result when the run switches. When a result reported no telemetry, the panel falls back to a clearly-labelled eval-shaped trace built from the result and its scores; span I/O falls back to the span's own values when the dedicated I/O fetch is empty.

## How it works

The run detail prefers the result's real ingested trace, sharing the trace viewer's cache key so opening the panel does not refetch. While that trace is still ingesting the panel shows a pending state with retry; once it lands it opens directly. For a fully-local result with no trace id, an eval-shaped trace is reconstructed — an evaluation-item root over a task span and one scorer span per scorer that ran — and its span I/O is seeded into the viewer's cache. The panel leads with a "Test case" header carrying the case id and a copy button and the case's status badge, and exposes per-result actions (review output, view source test case, and — for a real trace — save the selected span as a test case). Stepping walks the run's currently-visible results in table order; because the panel updates in place on trace-id change instead of remounting, the slide-in isn't replayed on each step.

### UI
- `features/traces/components/SpanTreeView.tsx` — provides `TreeRow`, `getVisibleSpanRows`, `buildTreeRows`, `SpanTreeViewHandle`.

### Trace viewer integration
- `features/traces/components/TraceViewerPanel.tsx` — the `headerIdentity` (label + copyable id) and `headerStatus` header slots, `traceOverride` for the reconstructed trace, `onNavigate` / `newTabPath` / `spanActions` / `spanExtraTags` / `onSelectionChange` slots, and updating in place on trace-id change so stepping doesn't re-animate.
- `features/traces/components/SpanInfoPanel.tsx` — reduces the eval-shaped span rows to their I/O and falls back to the span object's own input/output/metadata when the per-span I/O fetch returns nothing.

### Hooks / data
- `features/traces` — `useTrace` for the result's real ingested trace.

### Tests
- `features/evaluations/run-detail-compare.smoke.test.tsx` — covers `run-detail-compare.smoke`.
- `features/traces/components/SpanInfoPanel.review.test.tsx` — covers `SpanInfoPanel.review`.

## Test plan

- [ ] A result opens its real trace, led by the test case, with the case status shown and the case id copyable from the header.
- [ ] Result stepping with the chevrons walks the run's results in table order and does not re-animate the panel on each step.
- [ ] Switching runs resets the panel to the first result rather than holding the previous run's position.
- [ ] A result with no ingested trace opens a clearly-labelled reconstructed eval-shaped trace led by the test case, with span rows reduced to their I/O.
- [ ] Span I/O falls back to the span's own values when the dedicated I/O fetch is empty.
- [ ] A result whose telemetry is still ingesting shows the pending state with a working retry.

### DIFF
--- a/frontend/ui/src/features/evaluations/compare-derive.test.ts
+++ b/frontend/ui/src/features/evaluations/compare-derive.test.ts
@@ -8,6 +8,7 @@ import {
   type CaseFilterId,
 } from "./compare-derive";
 import type { CompareResultRow, RunComparison } from "./types";
+import { deriveComparisonState } from "@/lib/eval/comparison";
 
 function counts(p: Partial<RunComparison["caseCounts"]> = {}) {
   return {
@@ -21,7 +22,7 @@ function counts(p: Partial<RunComparison["caseCounts"]> = {}) {
   };
 }
 function comparison(p: Partial<RunComparison> = {}): RunComparison {
-  return {
+  const merged = {
     available: true,
     trustworthy: true,
     reasons: [],
@@ -32,6 +33,12 @@ function comparison(p: Partial<RunComparison> = {}): RunComparison {
     scorers: [],
     duration: { candidateMeanMs: null, baselineMeanMs: null, deltaMs: null, pairedCount: 0 },
     ...p,
+  } satisfies Omit<RunComparison, "state"> & Partial<Pick<RunComparison, "state">>;
+  // Keep `state` consistent with whatever available/trustworthy/reasons the test set,
+  // unless the test pinned it explicitly.
+  return {
+    ...merged,
+    state: p.state ?? deriveComparisonState(merged.available, merged.trustworthy, merged.reasons),
   };
 }
 
--- a/frontend/ui/src/features/evaluations/compare-runs.smoke.test.tsx
+++ b/frontend/ui/src/features/evaluations/compare-runs.smoke.test.tsx
@@ -194,6 +194,7 @@ const labData = {
   comparison: {
     available: true,
     trustworthy: true,
+    state: "trustworthy",
     reasons: [],
     baseline: { runId: "opus", runNumber: 41, candidateVersion: "opus" },
     mainScore: { candidate: 6 / 7, baseline: 1, delta: 6 / 7 - 1 },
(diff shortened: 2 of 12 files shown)


dark-sorceror:
@cubic-dev-ai review

XinweiHe (approved):

PR #1772, opened 2026-07-31
feat(compare): compare-with on run detail with diff-mode trace

### LINKED ISSUE #1665: Compare-with on run detail — per-cell diffs and a diff-mode trace against a picked baseline
## Summary

Comparison is an action, not stored state: from a run, pick another run of the same evaluation to compare against. Picking one drives the headline metrics, adds per-cell diffs to the results table, marks outputs that differ from the baseline, and opens result traces straight into diff mode. The comparison is shown plainly — no trustworthiness banner — because the picker only offers baselines of the same evaluation on the same dataset.

## Scope

- A content-sized compare-with picker whose options are the sibling runs of the same evaluation, wired to the backend-derived comparison for the selected pair.
- Headline metrics repointed to show candidate-vs-baseline deltas, with unchanged values reading as a plain zero delta rather than a blank.
- Per-cell diffs (main score, duration, cost) shown as a signed delta; a "≠ baseline" marker on outputs that differ from the matched baseline case; the change column shown only while comparing and hidden again on clear.
- Open the case trace in diff mode against the picked run's matching case, and repoint the run-drawer metrics at the selection.
- Diff cells sit inline in the existing single-line, ellipsis-truncated results rows without widening them.
- A signed-delta convention (improvement vs regression) shared across score, duration, and cost so headline and per-cell changes read the same way.

## Acceptance criteria

- Picking a baseline drives the metrics, per-cell diffs, and ≠-baseline marks.
- The case trace opens in diff mode against the picked run's matching case.
- Clearing the selection removes the change column and returns to the plain view.
- The picker lists only runs of the same evaluation, so a cross-evaluation baseline can't be chosen here.
- Diff cells stay inline and the results rows remain single-line while comparing.
- The comparison is presented plainly, with no trustworthiness banner.
- The UI consumes the comparison payload and recomputes no diff math itself.

## Out of scope

- The shareable comparison page (its own issue).
- The comparison engine and read model (backend issues).

dark-sorceror:
## Summary

Fixes: #1665

Comparison on the run detail is an action, not stored state: from a run, pick another run of the same evaluation to compare against. Picking one drives the headline counts, adds per-cell diffs to the results table, marks outputs that differ from the baseline, and opens result traces straight into diff mode. The comparison is shown plainly — the picker only offers baselines of the same evaluation, so there is no trustworthiness banner. The UI renders the backend-derived comparison payload and computes no diff math itself.

## How it works

The current run is the candidate; the picked run is the baseline. Selecting a baseline fetches the on-demand comparison for that pair from the same read model the shareable comparison page uses. The results table gains a "vs baseline" column and per-cell deltas, and opening a case enters the trace viewer already in diff mode against the baseline's matching case. Clearing the selection removes the column and returns to the plain view.

### UI
- `frontend/ui/src/features/evaluations/views/run-detail-view.tsx` — the content-sized "Compare with" `Select` whose options are the sibling runs of the same evaluation (newest first, current run excluded); the comparison banner with regressed/improved counts and a Clear action; the `vs baseline` change column rendered only while comparing; `CellDelta` signed deltas on main score, duration, and cost; the `≠ baseline` marker on outputs that differ from the matched case; `MainScoreCell` showing candidate-vs-baseline with the point delta; `ChangeCell` per-case verdict; and the wiring of `defaultDiffOn` + `diffBaseline` into `TraceViewerPanel` so a picked comparison opens each result directly into diff mode against its baseline counterpart.

### Data
- Consumes the shared `useCompareRuns` read model for the selected candidate/baseline pair and the `compareByCase` map keyed by test case, so the per-cell diffs, marks, and headline counts all come from one server-derived payload.

## Test plan

- [ ] Pick a baseline of the same evaluation and confirm the headline counts, per-cell diffs, and `≠ baseline` marks appear.
- [ ] Open a case while comparing and confirm its trace opens in diff mode against the picked run's matching case.
- [ ] Clear the selection and confirm the change column disappears and the plain view returns.
- [ ] Confirm the picker lists only runs of the same evaluation, so no cross-evaluation baseline can be chosen here.
- [ ] Confirm diff cells stay inline and the results rows remain single-line, ellipsis-truncated while comparing.
- [ ] Confirm no trustworthiness banner is shown for the comparison.
- [ ] Confirm an unchanged value reads as a plain zero delta rather than a blank.

### DIFF
--- a/frontend/ui/src/app/api/projects/[projectId]/evaluations/compare/route.test.ts
+++ b/frontend/ui/src/app/api/projects/[projectId]/evaluations/compare/route.test.ts
@@ -82,6 +82,7 @@ function run(over: Record<string, unknown> = {}) {
     datasetVersionId: "dv2",
     runNumber: 2,
     candidateVersion: "sonnet",
+    provenance: { declared_model: "claude-sonnet-4", git_commit: "abc1234" },
     status: "completed",
     baselineRunId: null,
     mainScore: 0.5,
@@ -118,6 +119,7 @@ function baselineRun(over: Record<string, unknown> = {}) {
     id: "run_b",
     runNumber: 1,
     candidateVersion: "opus",
+    provenance: null,
     mainScore: 1,
     startedAt: new Date("2026-07-20T00:00:00Z"),
     completedAt: new Date("2026-07-20T00:00:04Z"),
@@ -187,6 +189,10 @@ describe("assembly", () => {
     const b = await body(res);
 
     expect(b.candidate).toMatchObject({ id: "run_c", runNumber: 2, elapsedMs: 6000 });
+    // The declared candidate model is derived from the run's opaque provenance JSON.
+    expect(b.candidate.declaredModel).toBe("claude-sonnet-4");
+    // The baseline declared no model → null, never guessed from its candidate version.
+    expect(b.baseline.declaredModel).toBeNull();
     expect(b.baseline).toMatchObject({ id: "run_b", runNumber: 1, elapsedMs: 4000 });
     expect(b.comparison.available).toBe(true);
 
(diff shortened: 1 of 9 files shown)


dark-sorceror:
@cubic-dev-ai review

XinweiHe (approved):

PR #1775, opened 2026-07-31
feat(scorers): persist the SDK-reported scorer definition

### LINKED ISSUE #1655: Persist the SDK-reported scorer definition through the contract into the run manifest
## Summary

The SDK reports each scorer's definition — its type, and for an LLM judge the model + messages or for a code scorer the language + source, plus output type / description / metadata. Carry that definition through the contract into the persisted per-run manifest so the read-only Scorer detail can render it. The read side already parses these keys; they just need to survive registration.

## Scope

- Declare the definition fields on `ScorerRefSchema` (a non-strict Zod object) so they persist rather than being stripped on parse.
- Cover both definition shapes: LLM judge (`model` + `messages[]` of role/content) and code scorer (`language` + `source`), plus the shared `output_type` / `description` / `metadata`.
- Add a `ScorerMessage` element type (role + content) as the shape of an LLM judge's `messages[]`.
- Mirror the same fields on the Pydantic `ScorerRef` (plus a message model) and regenerate `public.json`.
- Store the full scorer manifest on the run; latest-wins is inherent because the registry reads all runs' manifests newest-first.

## Acceptance criteria

- A reported code scorer persists `scorer_type: code` + `language` + `source`; an LLM judge persists `model` + `messages`.
- Because `ScorerRefSchema` is non-strict, only fields it declares survive the parse — every persisted field must be named in the schema.
- The Pydantic `ScorerRef` uses `extra="ignore"` to mirror Zod's stripping, so identical payloads validate the same on both layers.
- A previously-seen scorer picks up its definition from its latest run.
- Legacy `{ name, version }` scorers still validate and store.
- `public.json` stays in lockstep with the added fields.

## Out of scope

- The scorer registry aggregation that reads these fields (its own issue).
- The Scorer-detail UI that renders them.

dark-sorceror:
## Summary

Fixes: #1655

The SDK reports each scorer's definition — its type, and for an LLM judge the model + messages or for a code scorer the language + source, plus output type / description / metadata. This carries that definition through the reporting contract into the persisted per-run scorer manifest so the read-only Scorer detail can render it. Because `ScorerRefSchema` is a non-strict object that strips unknown keys, every field that must survive registration has to be declared — so this PR declares them on both the Zod and Pydantic layers and regenerates the published schema.

### Tests
- `frontend/ui/src/lib/eval/scorer-registry.test.ts` — covers `scorer-registry`.

## Test plan
- [ ] A reported code scorer persists `scorer_type: code` + `language` + `source`.
- [ ] A reported LLM judge persists `model` + `messages`.
- [ ] Because `ScorerRefSchema` is non-strict, only declared fields survive the parse — verify an undeclared key is stripped.
- [ ] The Pydantic `ScorerRef` (`extra="ignore"`) validates the identical payload the same as Zod.
- [ ] A previously-seen scorer picks up its definition from its latest run.
- [ ] A legacy `{ name, version }` scorer still validates and stores.
- [ ] `public.json` stays in lockstep with the added fields.

### DIFF
--- a/frontend/ui/src/lib/eval/scorer-registry.test.ts
+++ b/frontend/ui/src/lib/eval/scorer-registry.test.ts
@@ -343,4 +343,128 @@ describe("aggregateScorers — value type and shaping edge cases", () => {
     const rows = aggregateScorers(scores, []);
     expect(rows[0].recentErrors.map((e) => e.message)).toEqual(["e4", "e3", "e2"]);
   });
+
+  it("surfaces the SDK-reported definition for a code scorer and an llm_judge scorer", () => {
+    const rows = aggregateScorers(
+      [
+        score({ scorerName: "nc", scorerVersion: "1", numericValue: 1 }),
+        score({ scorerName: "concise", scorerVersion: "1", numericValue: 0.9 }),
+      ],
+      [
+        {
+          scorers: [
+            {
+              name: "nc",
+              version: "1",
+              scorer_type: "code",
+              language: "python",
+              source: "def nc(ctx):\n    return 1.0",
+            },
+            {
+              name: "concise",
+              version: "1",
+              scorer_type: "llm_judge",
+              output_type: "score",
+              model: "claude-sonnet-5",
+              messages: [{ role: "system", content: "Rate the answer 0..1" }],
+            },
+          ],
+        },
+      ],
+    );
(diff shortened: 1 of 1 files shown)


dark-sorceror:
@cubic-dev-ai review

XinweiHe (approved):

PR #1777, opened 2026-07-31
feat(worker): derive per-result cost from the trace on ingest

### LINKED ISSUE #1656: Derive per-result cost from the trace on ingest, scoped to the candidate task
## Summary

The SDK doesn't report per-case cost, so derive it from the trace on ingest (`backend/worker`) and denormalize it onto `EvaluationResult.cost`. Scope the sum to the **candidate task** — exclude the scorer span and its whole subtree — so a scorer that itself calls a model (an LLM judge, or a code scorer that calls a provider) doesn't inflate the candidate's cost with its own judging cost. The scoping helper (`_task_cost_by_trace`) is a pure function of the span set, so it can be exercised without the ingest pipeline.

## Scope

- On ingest, sum provider-usage cost over the eval trace's spans, excluding the `SCORER` subtree at any depth, using the span-kind markers.
- Walk the span tree per trace, prune the `SCORER` span and everything beneath it, then sum `cost` over the remaining candidate-task spans.
- Match the derived total back to the result by trace id and write it onto `EvaluationResult.cost`.
- Make it self-healing — recomputed on every batch so late-arriving spans update the total — and never write `0` (a cost-less trace leaves `cost` null).
- When the SDK later reports an authoritative task cost, defer to it rather than overwrite.

## Acceptance criteria

- A result whose scorer calls a model reports task cost only; the judge's cost is excluded.
- Deep scorer subtrees (scorer → wrapper → LLM call) are fully excluded.
- A cost-less trace leaves `cost` null; the derivation is idempotent across batches.
- Multiple eval traces in one batch stay isolated — one trace's cost never leaks into another's result.
- Once `EvaluationResult.cost` holds an SDK-reported value, later batches do not overwrite it.
- The scoping helper is pure and unit-tested.

## Out of scope

- Span-kind preservation on ingest (its own issue) — this relies on those markers.
- Displaying cost in the run table / run detail (UI issues).

dark-sorceror:
## Summary

Fixes: #1656

The SDK doesn't report per-case cost, so this derives it from the trace on ingest and denormalizes it onto `EvaluationResult.cost`. The sum is scoped to the candidate task — the scorer span and its whole subtree are excluded — so a scorer that itself calls a model (an LLM judge, or a code scorer that calls a provider) doesn't inflate the candidate's cost with its own judging cost. The scoping is a pure function of the span set, so it can be exercised without the ingest pipeline; it exists only because per-result cost is absent today and is designed to defer to an SDK-reported cost once one lands.

## How it works

```
eval trace = EVALUATION(root) → TASK(...) → SCORER(→ its own LLM call)
  → seed the exclusion set from every SCORER span
    → walk the tree, pruning each SCORER span and all its descendants
      → sum cost over the remaining (candidate-task) spans
        → write onto EvaluationResult.cost by trace id; skip when the sum is 0
```

Recomputed on every batch, so late-arriving spans update the total; a cost-less trace leaves `cost` NULL rather than writing a misleading `0`.

### Tests
- `frontend/ui/src/lib/eval/trace-usage.test.ts` — covers `trace-usage`.

### UI
- `frontend/ui/src/features/evaluations/views/compare-case-drawer.tsx` — provides `CompareCaseDrawer`.
- `frontend/ui/src/features/evaluations/views/compare-runs-view.tsx` — provides `CompareRunsView`.

## Test plan
- [ ] A result whose scorer calls a model reports task cost only; the judge's cost is excluded.
- [ ] Deep scorer subtrees (scorer → wrapper → LLM call) are fully excluded.
- [ ] A cost-less trace attributes to `0` in the pure helper while the writer leaves the column NULL.
- [ ] The derivation is idempotent across batches (recompute yields the same total).
- [ ] Multiple eval traces in one batch stay isolated — one trace's cost never leaks into another's result.
- [ ] The scoping helper is pure and unit-tested.

### DIFF
--- a/frontend/ui/src/features/evaluations/views/compare-case-drawer.tsx
+++ b/frontend/ui/src/features/evaluations/views/compare-case-drawer.tsx
@@ -110,7 +110,7 @@ function OpMetric({ label, value }: { label: string; value: React.ReactNode }) {
   );
 }
 
-/** Trace-derived tokens for one side (task + scorer + combined), with pending/unknown. */
+/** Trace-derived tokens for one side (task + scorer, attributed separately), with pending/unknown. */
 function useUsage(projectId: string, traceId: string | null): TraceUsage {
   const q = useTrace(projectId, traceId ?? "", !!traceId);
   return React.useMemo(
@@ -119,10 +119,12 @@ function useUsage(projectId: string, traceId: string | null): TraceUsage {
   );
 }
 
+// Candidate-task tokens only — excludes the scorer (LLM-judge) subtree, matching
+// `row.candidateCost`/`row.baselineCost` above, which the backend derives the same way.
 function tokenText(u: TraceUsage): string {
   if (u.state === "pending") return "Pending";
   if (u.state === "unknown") return "Unknown";
-  return u.combined.totalTokens.toLocaleString("en-US");
+  return u.task.totalTokens.toLocaleString("en-US");
 }
 
 /**
--- a/frontend/ui/src/features/evaluations/views/compare-runs-view.tsx
+++ b/frontend/ui/src/features/evaluations/views/compare-runs-view.tsx
@@ -350,9 +350,11 @@ function sumPaired(
 }
 
 /**
- * Sums trace-derived tokens across cases (candidate/baseline) — only over the subset of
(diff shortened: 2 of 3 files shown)


dark-sorceror:
@cubic-dev-ai review

XinweiHe (approved):

PR #1779, opened 2026-07-31
feat(evaluations): data-fetching hooks over the read models

### LINKED ISSUE #1669: Data-fetching hooks over the derived read models (the UI recomputes nothing)
## Summary

The React Query hooks each surface uses to read the derived read models and the comparison result from the APIs, plus the shared `lib/eval` types — so the UI renders what the server derived and never recomputes comparison or registry math itself. Keeping this in one place keeps the views in sync with the read APIs and gives every surface one loading / error contract to render against.

## Scope

- Per-surface query hooks (runs list, run detail, compare, scorer catalog / detail, datasets, test-case runs) in `features/evaluations/hooks.ts`.
- Consume the derived read models and the comparison payload directly; share `ScorerRegistryRow` and the comparison shapes from `lib/eval`.
- All hooks scoped by project id so switching projects doesn't serve another project's cached rows.
- Align cache keys so a trace already opened from the run detail is reused when the same case opens in diff mode, rather than refetched.
- A single loading / error contract per hook so surfaces render loading / empty / error consistently rather than each inventing its own.
- Comparison and scorer hooks return the server's derived payload unchanged, exposing no client-side transform over it.
- The scorer catalog and detail read from the same registry read model, so the list and detail can't disagree.

## Acceptance criteria

- Every surface reads through a typed hook over the derived APIs.
- The UI does no comparison / registry math — it consumes server output.
- Shared types keep the views in sync with the read APIs, so a read-model field change surfaces as a type error.
- Query keys are project-scoped and aligned so an open trace is reused across the run and diff views.
- Each hook exposes a uniform loading / error shape the surfaces render against.
- A hook returns the derived payload unchanged, with no client-side transform layered on top.

## Out of scope

- The individual surfaces themselves (their own issues) — this is the data layer they share.

dark-sorceror:
## Summary

Fixes: #1669

The React Query hooks each evaluation surface uses to read the derived read models and the comparison result from the APIs, plus the shared types — so the UI renders what the server derived and never recomputes comparison or registry math itself. Keeping this in one place keeps the views in sync with the read APIs and gives every surface one loading / error contract to render against.

### Hooks / data
- `frontend/ui/src/features/evaluations/hooks.ts` — the per-surface query and mutation hooks over the project-scoped read APIs:
  - Datasets: `useDatasets`, `useDataset`, `useCreateDataset`, `useUpdateDataset`, `useDeleteDataset`, `useSaveTestCase`, `useUpdateTestCase`, `useTestCaseRuns`.
  - Evaluations and runs: `useEvaluations`, `useEvaluationRuns`, `useEvaluationRun`, `useDeleteRuns`, `useCreateHumanScore`.
  - Comparison: `useCompareRuns` — returns the server's derived candidate/baseline payload unchanged, shared by the run-detail compare-with and the shareable comparison view so the two can't disagree.
  - Scorers: `useScorers` and `useScorer` — the catalog and the per-scorer family, both over the same registry read model.
  - Trace links: `useTraceEvaluationResults`, `useTraceTestCases` — the trace → evaluation-run and span → dataset-case links, warmed when a trace opens.
  - Every hook is enabled on and keyed by `projectId`; a `getJson` / `sendJson` pair gives one uniform loading / error contract, and mutations invalidate the relevant caches. Query keys are aligned with the trace viewer's `["trace", projectId, traceId]` key so a trace already opened from the run detail is reused when the same case opens in diff mode.

### Frontend
- `frontend/ui/src/features/evaluations/views/evaluations-view.tsx` — provides `RunStatusBadge`, `EvaluationsView`, `formatElapsed`, `ScoreValue`.

## Test plan

- [ ] Confirm every surface reads through a typed hook over the derived APIs (no ad-hoc fetches in views).
- [ ] Confirm the UI does no comparison / registry math — hooks return the server payload unchanged.
- [ ] Confirm a change to a read-model field surfaces as a type error in the consuming views.
- [ ] Confirm query keys are project-scoped so switching projects doesn't serve another project's cached rows.
- [ ] Confirm a trace opened from the run detail is reused (not refetched) when the same case opens in diff mode.
- [ ] Confirm each hook exposes a uniform loading / error shape the surfaces render against.

### DIFF
--- a/frontend/ui/src/features/evaluations/hooks.ts
+++ b/frontend/ui/src/features/evaluations/hooks.ts
@@ -54,13 +54,20 @@ export function useDatasets(
   if (query.limit !== undefined) params.set("limit", String(query.limit));
   const qs = params.toString();
   return useQuery({
-    queryKey: ["datasets", "list", projectId, query.search_query ?? null, query.page ?? 0],
+    queryKey: [
+      "datasets",
+      "list",
+      projectId,
+      query.search_query ?? null,
+      query.page ?? 0,
+      query.limit ?? null,
+    ],
     queryFn: () =>
       getJson<{ data: DatasetRow[]; meta: Meta }>(
         `/api/projects/${projectId}/datasets${qs ? `?${qs}` : ""}`,
       ),
     enabled: !!projectId,
-    placeholderData: (prev) => prev,
+    placeholderData: (prev, prevQuery) => (prevQuery?.queryKey[2] === projectId ? prev : undefined),
   });
 }
 
@@ -71,7 +78,7 @@ export function useDataset(projectId: string, datasetId: string, versionId?: str
     queryFn: () =>
       getJson<DatasetDetailResponse>(`/api/projects/${projectId}/datasets/${datasetId}${qs}`),
     enabled: !!projectId && !!datasetId,
-    placeholderData: (prev) => prev,
+    placeholderData: (prev, prevQuery) => (prevQuery?.queryKey[2] === projectId ? prev : undefined),
   });
 }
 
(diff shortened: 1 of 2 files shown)


dark-sorceror:
@cubic-dev-ai review

XinweiHe (approved):

PR #1780, opened 2026-07-31
feat(evaluations): loading, empty, and error states

### LINKED ISSUE #1670: Loading / empty / error states with SDK-pointing empties and hydration-safe timestamps
## Summary

Make every surface handle loading, empty, and error explicitly, with empty states that point at the SDK, and keep timestamps hydration-safe so the server and client render the same string. Consistent empty pages across the Evaluations tabs round out the surface. The three states are visually distinct so a slow fetch never reads as "no data".

## Scope

- Loading / empty / error states across Datasets, Evaluations, Run detail, Comparison, and Scorers.
- Empty states that tell the user to report a run or save a case from the SDK, not to create one in the UI; the runs empty state carries the SDK snippet.
- Error states that surface the failure and offer a retry rather than collapsing to an empty table.
- Hydration-safe timestamp formatting (independent of ICU / locale data), dropping tabular-nums mismatches between server and client.
- Loading states sized to their populated layout so tables don't jump when data arrives.
- Each empty state names the specific SDK action that would fill it — report a run, save a case, or run a candidate.

## Acceptance criteria

- Each surface renders distinct loading / empty / error states that can't be confused for one another.
- Empty states point at the SDK rather than an in-UI create flow.
- An errored fetch shows the error and a retry, not a false empty.
- Timestamps render identically server- and client-side (no hydration mismatch).
- Loading placeholders match the populated layout so content doesn't shift in.
- Each surface's empty copy names the specific SDK action that would fill it.
- The Evaluations tabs share a consistent empty page.

## Out of scope

- The functional surfaces themselves (their own issues) — this is the states pass over them.

dark-sorceror:
## Summary

Fixes: #1670

Make every evaluation surface handle loading, empty, and error explicitly, with empty states that point at the SDK, and keep timestamps hydration-safe so the server and client render the same string. The three states are visually distinct so a slow fetch never reads as "no data", and each empty state names the specific SDK action that would fill it — report a run, save a case, or run a candidate.

### Shared state primitives
- `frontend/ui/src/features/offline-eval/components/timestamp.tsx` — the hydration-safe absolute timestamp: a deterministic UTC render on the server and first client paint (so the two match exactly), swapping to the app's local-time format once mounted, independent of locale data.

## Test plan

- [ ] Confirm each surface renders distinct loading / empty / error states that can't be confused for one another.
- [ ] Confirm empty states point at the SDK (report a run, save a case, run a candidate) rather than an in-UI create flow.
- [ ] Confirm an errored fetch shows the error and a retry, not a false empty.
- [ ] Confirm timestamps render identically server- and client-side with no hydration mismatch.
- [ ] Confirm loading placeholders match the populated layout so content doesn't shift in.
- [ ] Confirm the Evaluations tabs share a consistent empty page.

### DIFF
--- a/frontend/ui/src/features/offline-eval/components/timestamp.tsx
+++ b/frontend/ui/src/features/offline-eval/components/timestamp.tsx
@@ -1,36 +1,21 @@
-"use client";
-
-import * as React from "react";
-import { cn, parseAsUTC } from "@/lib/utils";
+import { cn } from "@/lib/utils";
 import { formatStamp } from "../utils";
 
 /**
- * Deterministic UTC render of `formatStamp`'s format. Server and browser both
- * produce this identical string, so it is safe for SSR / first client paint.
- */
-function formatStampUTC(iso: string): string {
-  const d = parseAsUTC(iso);
-  const p = (n: number) => String(n).padStart(2, "0");
-  return `${d.getUTCFullYear()}-${p(d.getUTCMonth() + 1)}-${p(d.getUTCDate())} ${p(
-    d.getUTCHours(),
-  )}:${p(d.getUTCMinutes())}:${p(d.getUTCSeconds())}`;
-}
-
-/**
- * Hydration-safe absolute timestamp.
+ * Absolute timestamp, rendered as local time.
  *
- * These screens render hardcoded fixtures during SSR, so `formatStamp` (local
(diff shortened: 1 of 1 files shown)


dark-sorceror:
@cubic-dev-ai review

XinweiHe (approved):

Issue #1748, opened 2026-07-31
fix(worker): traceroot.span.tags is a typed SDK option but the backend silently destroys it


ka1kqi:
## Problem

The TypeScript SDK ships tags as a first-class, typed option:

- `observe(fn, { tags: ['prod'] })` — `tags?: string[]`, serialized onto the span as `traceroot.span.tags`
- `usingAttributes({ sessionId, userId, tags, metadata })` — the docstring example shows `tags: ['prod']`
- exported constants `SPAN_TAGS` / `TRACE_TAGS`

The backend half was never built, and worse than merely unimplemented: `traceroot.span.tags` is listed in the transform's `_KNOWN_ATTRIBUTE_PREFIXES` — the strip-list for attributes "already extracted into dedicated fields" — but **no extraction and no column exist**. So tags are stripped from the metadata blob and then written nowhere. Ingest returns 200; the values are unrecoverable.

Verified end-to-end on staging: spans sent with tags via the SDK's own API round-trip every other field correctly, and tags come back nowhere — not in list, detail, or export; no `tags` in `schemas/public.py`, `schemas/traces.py`, or the committed `openapi/public.json`; no column in any ClickHouse migration.

## Why it matters

A customer following the SDK's typed API loses labeling data permanently and silently. If tags later become filterable, there is no history to backfill — the data was destroyed on write, not hidden.

## Competitor context

Tags are table stakes among direct competitors: Langfuse has trace-level tags settable via SDK/UI and filterable in both UI and public API (`GET /api/public/traces?tags=a&tags=b`, ALL-match); Braintrust has project + schema-inferred tags with filter and group-by; a third OSS agent-observability platform treats `tags` as a reserved metadata key used for trace filtering. Generic-analytics products subsume it under arbitrary event properties.

## Suggested resolution, two steps

1. **Stop the destruction now (one line):** remove `traceroot.span.tags` from `_KNOWN_ATTRIBUTE_PREFIXES` so tags land in the span metadata blob like any unknown attribute. Data becomes recoverable while the feature decision is made — this mirrors the reserved-metadata-key approach.
2. **Feature decision:** implement tags properly — trace-level `Array(String)` column, exposure in list/detail/export, filter in the trace list (fits the existing trace-filter design; the ALL-match `?tags=` semantics above are the reference) — or deliberately retire the option from the SDK so the API stops advertising it. Keeping the current state (typed API, silent discard) is the one wrong option.

## Provenance

[...]

Issue #1752, opened 2026-07-31
perf(traces): trace-list dedup CTE sorts the full matching set with input/output text — peak memory unbounded in tenant size, can OOM shared ClickHouse


ka1kqi:
## Severity: availability, not just latency

Peak memory for a single trace-list page load scales with **(rows matching the WHERE) × (row width, including the `input`/`output` text)** — it is **unbounded in tenant size**, because neither the row count nor the payload width is capped before the sort (see problems A and B below). So this is not a fixed cost; it grows with a tenant's data and with how large their traces' I/O payloads are. On multi-tenant infrastructure that makes a trace-list page load a denial-of-service vector: past some tenant size, a few concurrent page loads exhaust the node and evict every tenant on it.

Concrete data point (one measurement, NOT a production figure): on a local ClickHouse (7.75 GiB host) with a 5M-trace project, two back-to-back page-load probes peaked at ~3.5 GiB each and OOM-killed the server. The absolute number is entirely environment- and payload-dependent — bigger hosts absorb more, larger I/O payloads make it worse, and a production cluster will hit the wall at a different (larger) tenant size. What generalizes is the curve, not the gigabytes: memory has no upper bound as a tenant grows.

## The query

`trace_reader.py` list query. The trace-paging CTE:

```sql
SELECT t.trace_id, ..., t.input, t.output
FROM traces AS t
WHERE project_id = X AND source = 'user' AND [time window]
ORDER BY t.ch_update_time DESC
LIMIT 1 BY t.project_id, t.trace_id      -- dedup, not a row cap
-- then: ORDER BY trace_start_time DESC LIMIT 50 OFFSET n
```

Two compounding problems, both in this one CTE (the span-aggregation half is sound — see #963):

**A. The sort key can't terminate the scan early.** `traces` is `ORDER BY (project_id, toDate(trace_start_time), trace_id)` — day granularity. Queries order by full-millisecond `trace_start_time`, so ClickHouse can't tell which rows within a day are newest and must read *all* matching rows, sort by full timestamp, then take 50. Measured on identical data: ordering by `toDate(trace_start_time)` (matches the key) reads 115k rows / 11 ms; ordering by full `trace_start_time` (what the app does) reads 5.50M rows / 179 ms for the same 50 rows — **48x**, purely from the granularity mismatch.

[...]

PR #1789, opened 2026-08-01
perf(traces): keep input/output out of the trace-paging dedup sort

nosey-dewdrop:
Step 1 of #1752, the query-only half. No migration, no schema change.

## What changed

The `page` CTE deduped with `input`/`output` in its projection, so its `ORDER BY ch_update_time DESC` sorted every row matching the WHERE with the I/O text attached. Any column named in that dedup is materialized for the whole matching set, not for the page, which is why peak memory scaled with (matching rows x payload width) rather than with the 50 rows returned.

The CTE is now two stages:

- `page_keys` dedups, orders and pages on identity and display metadata only. The wide sort is gone.
- `page_io` fetches `input`/`output` for the page traces only.
- `page` joins them, so the final SELECT is unchanged.

## The version-mixing hole, not in the issue

The issue warns about the over-fetch variant and its correctness cliff. There is a second one in any naive split: if a trace has several rows sharing one `ch_update_time`, two independent `LIMIT 1 BY` passes can settle on different rows, and the page would then show one version metadata with another version text. A single CTE could not do that because it picked once.

`page_io` therefore matches on the pair rather than the id:

```sql
WHERE project_id = {project_id:String}
  AND (trace_id, ch_update_time) IN (
      SELECT trace_id, ch_update_time FROM page_keys
  )
```

so it is pinned to the exact row stage one selected.

## What this does not do

Problem A, the day-granularity sort key, is untouched, so the full-timestamp `ORDER BY` still cannot terminate the scan early and deep pagination is still expensive. That is the migration in step 2 of the issue.

## Test plan

Three tests added to `tests/rest/test_trace_reader_filters.py`, using the file existing mocked-client pattern that asserts on the generated SQL:

- `test_page_dedup_does_not_carry_io_text_through_its_sort` — pins the invariant that regressed
- `test_io_text_is_fetched_only_for_the_page_and_pinned_to_the_deduped_row` — page restriction and the pair match
- `test_page_still_exposes_io_columns_to_the_final_select` — the split does not change what the list returns

```
pytest tests/rest/test_trace_reader_filters.py
    12 passed

pytest tests/rest/
    584 passed, 1 skipped
```

The skip is `test_filter_columns_parity.py`, which skips itself on this branch.

I could not reproduce the memory figures: they came from a local ClickHouse with a 5M-trace project, and I have no ClickHouse instance. The tests therefore pin the query shape, which is what the fix changes, rather than the gigabytes. Someone with the seeded instance can confirm the peak drops in the range the issue predicts.

No UI change, so no screenshots.

AI usage: written with an AI assistant. I read and own every line, and the version-mixing hole above is my own finding.

---
## Summary by cubic
Reduces peak memory on trace list paging by removing `input`/`output` from the dedup sort and fetching them only for the page, and preserves `traceroot.span.tags` by keeping them in span metadata. Implements the query-only half of #1752 and fixes data loss from `tags` (#1748).

- **Performance**
  - Split the page CTE into `page_keys` (dedup/order/page on IDs/metadata) and `page_io` (fetch `input`/`output` only for the page).
  - Pin I/O fetch on `(trace_id, ch_update_time)` to prevent version mix when multiple rows share an update time.
  - Final SELECT is unchanged; no schema or migration.

- **Bug Fixes**
  - Stop stripping `traceroot.span.tags` during ingest: remove it from the known-attribute strip list and merge it into `metadata`, including when explicit metadata is set.
  - Add tests to ensure tags persist and the new query shape is enforced.

<sup>Written for commit 0aa56a966b873bca7da2efa366a15f390e77422b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1789?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/rest/services/trace_reader.py
+++ b/backend/rest/services/trace_reader.py
@@ -450,16 +450,22 @@ class TraceReaderService:
         # This avoids scanning/joining every span in the project on each list call,
         # and never groups by the large input/output text columns. See #963.
         query = f"""
-            WITH page AS (
+            WITH page_keys AS (
                 -- Dedup ReplacingMergeTree by latest ch_update_time FIRST (correctness),
                 -- THEN order by start time for pagination.
+                --
+                -- input/output are deliberately absent here. This dedup sorts every
+                -- row matching the WHERE, so any column listed rides through that
+                -- sort: carrying the I/O text made peak memory scale with
+                -- (matching rows x payload width) instead of with the page. The
+                -- text is fetched below for the page's traces only.
                 SELECT
                     trace_id, project_id, name, trace_start_time,
-                    user_id, session_id, input, output
+                    user_id, session_id, ch_update_time
                 FROM (
                     SELECT
                         t.trace_id, t.project_id, t.name, t.trace_start_time,
-                        t.user_id, t.session_id, t.input, t.output
+                        t.user_id, t.session_id, t.ch_update_time
                     FROM traces AS t
                     WHERE {where_clause}
                     ORDER BY t.ch_update_time DESC
@@ -467,6 +473,30 @@ class TraceReaderService:
                 )
                 ORDER BY trace_start_time DESC
                 LIMIT {{limit:UInt32}} OFFSET {{offset:UInt32}}
+            ),
+            page_io AS (
+                -- The heavy columns, for the page's traces only. Pinned to the exact
+                -- (trace_id, ch_update_time) row the dedup above selected, so the
+                -- text cannot come from a different version than the metadata when
+                -- a trace has several rows sharing one update time.
+                SELECT trace_id, input, output
+                FROM (
+                    SELECT trace_id, input, output, ch_update_time
+                    FROM traces
+                    WHERE project_id = {{project_id:String}}
+                      AND (trace_id, ch_update_time) IN (
+                          SELECT trace_id, ch_update_time FROM page_keys
+                      )
+                    ORDER BY ch_update_time DESC
+                    LIMIT 1 BY project_id, trace_id
+                )
+            ),
+            page AS (
+                SELECT
+                    k.trace_id, k.project_id, k.name, k.trace_start_time,
+                    k.user_id, k.session_id, io.input AS input, io.output AS output
+                FROM page_keys AS k
+                LEFT JOIN page_io AS io ON k.trace_id = io.trace_id
             ),
             span_agg AS (
                 SELECT
--- a/backend/worker/otel_transform.py
+++ b/backend/worker/otel_transform.py
@@ -38,7 +38,12 @@ from datetime import UTC, datetime
 from typing import Any
 
 from shared.enums import SpanKind, SpanStatus
-from shared.span_attributes import SPAN_IDS_PATH, SPAN_PATH, SPAN_TREE_ATTRIBUTES
+from shared.span_attributes import (
+    SPAN_IDS_PATH,
+    SPAN_PATH,
+    SPAN_TAGS,
+    SPAN_TREE_ATTRIBUTES,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -95,7 +100,11 @@ _KNOWN_ATTRIBUTE_PREFIXES = {
     "traceroot.span.output",
     "traceroot.span.type",
     "traceroot.span.metadata",
-    "traceroot.span.tags",
+    # traceroot.span.tags is deliberately absent. The SDK ships it as a typed
+    # option, but nothing extracts it and no column holds it, so listing it here
+    # stripped customer-sent labels and wrote them nowhere. It now falls into the
+    # metadata blob like any unextracted attribute, which keeps the values
+    # recoverable until tags get a dedicated field.
     "traceroot.trace.",
     "traceroot.environment",
     # Nothing extracts this any more; it is listed purely to keep a
@@ -978,9 +987,13 @@ def transform_otel_to_clickhouse(
                 # its parent was still open. That hit exactly the spans users
                 # annotate — usually leaves, whose long-lived parents are the ones
                 # still in flight — so the paths ride along with user metadata.
-                span_path_attrs = {
+                # Tags ride along for the same reason: they have no column, and a
+                # span that sets explicit metadata is exactly a span whose author
+                # is labelling things, so the explicit branch would drop them from
+                # the spans most likely to carry them.
+                carried_attrs = {
                     key: span_attrs[key]
-                    for key in SPAN_TREE_ATTRIBUTES
+                    for key in (*SPAN_TREE_ATTRIBUTES, SPAN_TAGS)
                     if span_attrs.get(key) is not None
                 }
                 explicit_metadata = span_attrs.get("traceroot.span.metadata")
@@ -992,7 +1005,7 @@ def transform_otel_to_clickhouse(
                         except (TypeError, ValueError):
                             parsed_explicit = None
                     if isinstance(parsed_explicit, dict):
-                        span_record["metadata"] = json.dumps({**parsed_explicit, **span_path_attrs})
+                        span_record["metadata"] = json.dumps({**parsed_explicit, **carried_attrs})
                     elif isinstance(explicit_metadata, str):
                         # Not a JSON object (free-text or a scalar): store it as
                         # given. Nothing to merge into, so this span has no paths.
--- a/tests/rest/test_trace_reader_filters.py
+++ b/tests/rest/test_trace_reader_filters.py
@@ -168,3 +168,57 @@ def test_unfiltered_list_without_window_stays_unbounded():
     params = svc._client.query.call_args_list[0].kwargs["parameters"]
     assert "start_after" not in params
     assert "t.trace_start_time >=" not in page_sql
+
+
+def test_page_dedup_does_not_carry_io_text_through_its_sort():
+    """The trace-paging dedup must not list input/output.
+
+    That dedup sorts every row matching the WHERE, so any column named in it is
+    materialized for the whole matching set, not for the page. Carrying the I/O
+    text there made peak memory scale with (matching rows x payload width),
+    which is unbounded in tenant size — a page load on a large tenant could
+    exhaust a shared node. The text is fetched separately for the page's traces.
+    """
+    svc = _service_with_mock_client()
+    _drive(svc)
+
+    svc.list_traces(project_id="p1")
+
+    page_sql = svc._client.query.call_args_list[0].args[0]
+    dedup = page_sql.split("page_keys AS (")[1].split("),")[0]
+    assert "input" not in dedup, (
+        "input rides through the pre-page dedup sort again; peak memory is back to "
+        "scaling with the matching set instead of the page"
+    )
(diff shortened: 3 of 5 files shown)


tests/rest/test_trace_reader_filters.py
```diff
@@ -168,3 +168,57 @@ def test_unfiltered_list_without_window_stays_unbounded():
...
+    which is unbounded in tenant size — a page load on a large tenant could
+    exhaust a shared node. The text is fetched separately for the page's traces.
+    """
+    svc = _service_with_mock_client()
+    _drive(svc)
+
+    svc.list_traces(project_id="p1")
+
+    page_sql = svc._client.query.call_args_list[0].args[0]
+    dedup = page_sql.split("page_keys AS (")[1].split("),")[0]
```
ka1kqi, line 188: **Blocking — this test is vacuous.** The split truncates at the first `),` in the region, which is inside the CTE's first comment line (`…FIRST (correctness),`), so `dedup` only ever contains that comment fragment and never reaches the SELECT list. Both `assert "input" not in dedup` and the `output` assert pass tautologically: re-adding `t.input`/`t.output` to `page_keys` — the exact regression this test exists to catch — leaves the suite green.

Note the obvious fix has a trap: splitting on `page_io AS (` captures the whole CTE, but the CTE's own comment says "input/output are deliberately absent here", which would trip `"input" not in`. Suggest stripping `--` comment lines from the captured region first, then asserting — that pins the actual projection.

backend/rest/services/trace_reader.py
```diff
@@ -468,6 +474,30 @@ def list_traces(
                 ORDER BY trace_start_time DESC
```
ka1kqi, line 474: **Blocking — pagination needs a tiebreaker.** `ORDER BY trace_start_time DESC` with LIMIT/OFFSET has no tiebreaker, and ClickHouse inlines this CTE independently at each reference site (`page_io`'s IN, the `page` join, `span_agg` via `page`). `trace_start_time` is millisecond-precision and batch inserts can tie, so when tied traces straddle a page boundary, the evaluation feeding `page_io`'s IN can select a different boundary trace than the evaluation used as the join's left side — that trace then matches nothing in `page_io` and the API silently returns `"input": null` for a trace whose row has text. Pre-split, I/O rode the same dedup rows and could never go missing.

One-line fix:

```sql
ORDER BY trace_start_time DESC, trace_id
```

which makes every evaluation agree, and stabilizes page boundaries across requests as a bonus.

backend/rest/services/trace_reader.py
```diff
@@ -468,6 +474,30 @@ def list_traces(
...
+            page_io AS (
+                -- The heavy columns, for the page's traces only. Pinned to the exact
+                -- (trace_id, ch_update_time) row the dedup above selected, so the
+                -- text cannot come from a different version than the metadata when
+                -- a trace has several rows sharing one update time.
+                SELECT trace_id, input, output
+                FROM (
+                    SELECT trace_id, input, output, ch_update_time
+                    FROM traces
+                    WHERE project_id = {{project_id:String}}
```
ka1kqi, line 486: **Perf — this drops the time bounds.** The WHERE here keeps only `project_id`, discarding the outer `where_clause`'s `trace_start_time` bounds — so `page_io` gets no partition pruning. A tenant with 24 months of history listing yesterday's traces has `page_keys` read ~1 partition, then `page_io` walks key columns across all 24 partitions to resolve ~50 ids (PREWHERE keeps the blob decompression near the matches, but the key-column scan is still full-history). And it executes twice, since `page` is referenced twice. That gives back a chunk of what this PR sets out to win, on the same endpoint.

Fix: repeat the time-bound conditions here (the params are already bound), or constrain with min/max `trace_start_time` from `page_keys`.

backend/rest/services/trace_reader.py
```diff
@@ -468,6 +474,30 @@ def list_traces(
...
+                    ORDER BY ch_update_time DESC
+                    LIMIT 1 BY project_id, trace_id
+                )
+            ),
+            page AS (
+                SELECT
+                    k.trace_id, k.project_id, k.name, k.trace_start_time,
+                    k.user_id, k.session_id, io.input AS input, io.output AS output
+                FROM page_keys AS k
+                LEFT JOIN page_io AS io ON k.trace_id = io.trace_id
```
ka1kqi, line 499: **One-word fix — point `span_agg` at `page_keys`.** Just below this hunk, `span_agg` filters `trace_id IN (SELECT trace_id FROM page)` — but it only needs ids, and because CTEs inline per reference, going through `page` drags an extra evaluation of the join *and* of the unpruned `page_io` scan (see comment above), expanding `page_keys`' dedup sort to ~4 evaluations per call (pre-PR: 2). It also adds a third independent evaluation of the tie-prone pagination, so a boundary trace could additionally get blank `span_count`/`duration`. `IN (SELECT trace_id FROM page_keys)` drops all of that.

backend/rest/services/trace_reader.py
```diff
@@ -468,6 +474,30 @@ def list_traces(
                 ORDER BY trace_start_time DESC
                 LIMIT {{limit:UInt32}} OFFSET {{offset:UInt32}}
             ),
+            page_io AS (
+                -- The heavy columns, for the page's traces only. Pinned to the exact
+                -- (trace_id, ch_update_time) row the dedup above selected, so the
+                -- text cannot come from a different version than the metadata when
```
ka1kqi, line 480: Worth tightening this comment: the pair-pin does prevent mixing across rows with *different* update times (the common case), but for the case it names — several rows sharing one `ch_update_time` — the IN matches *every* tied row, and this CTE's own `LIMIT 1 BY` tie-breaks independently of `page_keys`'. Same-millisecond duplicates are real here (trace batch inserts stamp one `now64(3)` per batch, and a concurrent-root race can produce two same-ms rows with different content), so pre-merge the list can still pair row A's metadata with row B's text — a tear the old single-SELECT couldn't produce. Either add a deterministic tiebreaker to both ORDER BYs or reword the comment to claim only what holds; a false "this cannot happen" is the expensive kind of comment.

backend/rest/services/trace_reader.py
```diff
@@ -468,6 +474,30 @@ def list_traces(
...
+                -- a trace has several rows sharing one update time.
+                SELECT trace_id, input, output
+                FROM (
+                    SELECT trace_id, input, output, ch_update_time
+                    FROM traces
+                    WHERE project_id = {{project_id:String}}
+                      AND (trace_id, ch_update_time) IN (
+                          SELECT trace_id, ch_update_time FROM page_keys
+                      )
+                    ORDER BY ch_update_time DESC
```
ka1kqi, line 490: Dead ordering that reads as load-bearing: within `page_io`, every row that survives the IN pin is an exact tie on `ch_update_time`, so this ORDER BY decides nothing; and `project_id` in the `LIMIT 1 BY` is dead (the WHERE pins one project, and it isn't in the projection). The whole CTE flattens to:

```sql
SELECT trace_id, input, output
FROM traces
WHERE project_id = {project_id:String}
  AND (trace_id, ch_update_time) IN (SELECT trace_id, ch_update_time FROM page_keys)
LIMIT 1 BY trace_id
```

As written, the sort looks like it's doing version-pinning, and the next person to touch this has to re-derive the whole argument to discover it's a no-op.

backend/worker/otel_transform.py
```diff
@@ -992,7 +1005,7 @@ def transform_otel_to_clickhouse(
                         except (TypeError, ValueError):
                             parsed_explicit = None
                     if isinstance(parsed_explicit, dict):
-                        span_record["metadata"] = json.dumps({**parsed_explicit, **span_path_attrs})
+                        span_record["metadata"] = json.dumps({**parsed_explicit, **carried_attrs})
                     elif isinstance(explicit_metadata, str):
```
ka1kqi, line 1009: **Blocking for the tags half — tags are still dropped when explicit metadata isn't a dict.** `carried_attrs` is only merged in the dict branch above. A span that sets `traceroot.span.metadata = "deploy note"` (free text — or a scalar, or a JSON array) plus `tags=["prod"]` takes this `elif`/`else` path: metadata is stored verbatim and the tags are dropped — the same 200-and-gone loss this PR fixes, still reachable. That contradicts the new `SPAN_TAGS` docstring ("has to survive both metadata branches"), and the new tests only cover the dict branch. Either carry the attrs on these branches too, or narrow the docstring/PR claim to the dict case and leave a marker. (The same branches already drop the span-tree path attrs — pre-existing, but worth noting while you're here.)

backend/worker/otel_transform.py
```diff
@@ -978,9 +987,13 @@ def transform_otel_to_clickhouse(
...
                 # still in flight — so the paths ride along with user metadata.
-                span_path_attrs = {
+                # Tags ride along for the same reason: they have no column, and a
+                # span that sets explicit metadata is exactly a span whose author
+                # is labelling things, so the explicit branch would drop them from
+                # the spans most likely to carry them.
+                carried_attrs = {
                     key: span_attrs[key]
-                    for key in SPAN_TREE_ATTRIBUTES
+                    for key in (*SPAN_TREE_ATTRIBUTES, SPAN_TAGS)
```
ka1kqi, line 996: Two smaller things: this tuple (and the dict) is rebuilt per span in the hot ingest loop but only consumed in the dict branch — worth hoisting/moving. More structurally: the `else` branch already expresses the general rule (`not _is_known_attribute(k)` survives), and tags are the second attribute lost to this enumerated list. Building `carried_attrs` from that same predicate — one mechanism instead of two — means the next SDK option can't silently repeat this bug; at minimum, hoist a named module-level `CARRIED_METADATA_ATTRIBUTES` next to `SPAN_TREE_ATTRIBUTES` (whose docstring is now stale about being the full carried set).

ka1kqi (changes requested):
Thanks for this — the design is exactly right (the wide sort was the problem, and the two-stage split with the final re-sort after the join is the correct shape), and the version-mixing observation in the PR description is a subtle catch a naive split would have missed. A handful of things need fixing before merge; the three most important are one-liners (see inline). Two notes that don't fit an inline anchor:

1. **Tags land in storage but stay invisible in the product.** The span metadata panel (`SpanInfoPanel.tsx`) strips every `traceroot.span.*` key, and the trace-detail read re-packs only the span-tree attributes — so after this PR, tags are only observable via direct API/DB reads with the `metadata` field group. If "recoverable, not surfaced" is the intended scope for #1748, please say so explicitly in the PR description (and we'll file the surfacing follow-up); otherwise the closing issue will look unfixed to its reporter.
2. **Commit messages reference issue numbers** ("Refs #1752", "Closes #1748", "Step 1 of #1752 only"). Repo convention keeps issue links in the PR description only — could you reword the commit messages? `closes #N` here in the PR body is the right place and already present.

PR #1823, opened 2026-08-11
fix(ui): restore the persisted date range after retention resolves

### LINKED ISSUE #1822: fix(ui): reload clamps the restored time-range selection to 14d because retention reads as free-plan while loading
## Symptom

On any page with the shared time-range filter (traces, sessions, users, detectors, dashboards), the per-project stored selection survives in-app navigation but not a browser reload: after reload the filter shows "Last 14 days" regardless of the stored value and regardless of the page's actual default (global default is `1d`; only detectors default to `14d`).

## Repro

1. Workspace on a paid plan (retention > 15 days, e.g. Pro/90d).
2. Pick "Last 30 days" (or 60d/90d) on any list page — it persists to localStorage.
3. Navigate to another page in-app → 30d restored correctly.
4. Reload the page (plain F5 is enough) → filter shows "Last 14 days". localStorage still holds `30d`, so step 3 keeps "working" afterwards, which makes the bug look intermittent.

## Root cause

`useRetention` (`frontend/ui/src/lib/hooks/use-retention.ts`) computes `retentionDays` from `workspace?.billingPlan || PlanType.FREE`, so while the project/workspace queries are in flight it reports the free plan's 15 days. The localStorage restore in `useUrlDateFilter` (`frontend/ui/src/lib/hooks/use-url-date-filter.ts`) runs exactly once, in a mount-time layout effect, and passes that transient value to `clampDateFilter` — clamping any stored preset over 15 days down to `14d`, the widest preset within the fallback window. The restore is guarded by a once-only ref, so when the real retention arrives moments later nothing re-adopts the stored value.

In-app navigation is unaffected because the workspace query is already in the react-query cache, so the restore sees the real retention at mount.

## Proposed fix

- Store the raw selection and apply `clampDateFilter` reactively in a derived value (display + `timestamps`), so the clamp re-evaluates when `retentionDays` changes instead of being baked into state at restore time. This also replaces the per-call-site clamps in the initial-state/restore/URL-sync paths.
- Optionally, have `useRetention` return `retentionDays: undefined` while its queries are pending (keeping fail-closed for unrecognized plan strings), so the reload path never briefly renders or queries the wrongly clamped window.

## Video

https://github.com/user-attachments/assets/a7aa15c6-25e8-44bf-afcf-28d7e35f109d

## Affected surfaces

Every consumer of `useListPageState` / `useUrlDateFilter` — trace, session, user, detector list/detail pages, and dashboard detail.

Gujiassh:
## Summary

Fixes #1822.

The hook now keeps the raw per-project date-filter selection in state and derives the retention-clamped option for display and timestamps. When workspace retention changes from the loading fallback to the actual plan, a stored `30d` selection is re-evaluated instead of remaining stuck at `14d`.

## Verification

- `pnpm exec vitest run src/lib/hooks/use-url-date-filter.test.tsx` (11 passed)
- `pnpm exec eslint src/lib/hooks/use-url-date-filter.ts src/lib/hooks/use-url-date-filter.test.tsx` (0 errors)
- `pnpm exec prettier --check src/lib/hooks/use-url-date-filter.ts src/lib/hooks/use-url-date-filter.test.tsx`
- `git diff --check`

The repository-wide TypeScript check and production build are currently blocked by the checkout's missing Prisma generated client; the changed files pass their focused checks.

---
## Summary by cubic
Restores the user's saved date range after retention resolves and prevents a brief clamp to the free plan on reload. The hooks keep the raw pick and derive the clamped window for display and timestamps.

- **Bug Fixes**
  - Report retention as undefined while project/workspace lookups are pending, so restored ranges don’t clamp in the interim.
  - Re-clamp on resolution and fire the change callback when the effective window narrows to reset pagination.
  - Clamp explicit URL selections against the resolved retention and extend tests for pending state, re-clamp timestamps, raw preference persistence, and clamp-driven resets.

<sup>Written for commit e736dc79a2ba1107e11d031d14d87128dc92df13. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1823?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/ui/src/lib/hooks/use-retention.test.tsx
+++ b/frontend/ui/src/lib/hooks/use-retention.test.tsx
@@ -1,0 +1,73 @@
+// @vitest-environment jsdom
+import { renderHook } from "@testing-library/react";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { PlanType } from "@traceroot/core";
+import { useRetention } from "./use-retention";
+
+const state = vi.hoisted(() => ({
+  project: { data: undefined as { workspace_id?: string } | undefined, isPending: true },
+  workspace: { data: undefined as { billingPlan?: string } | undefined, isPending: true },
+}));
+
+vi.mock("@/features/projects/hooks", () => ({
+  useProject: () => state.project,
+}));
+vi.mock("@/features/workspaces/hooks", () => ({
+  useWorkspace: () => state.workspace,
+}));
+
+beforeEach(() => {
+  state.project = { data: undefined, isPending: true };
+  state.workspace = { data: undefined, isPending: true };
+});
+
+describe("useRetention", () => {
+  it("reports retention as unknown while the project lookup is in flight", () => {
+    const { result } = renderHook(() => useRetention("p1"));
+
+    expect(result.current.retentionDays).toBeUndefined();
+  });
+
+  it("reports retention as unknown while the workspace lookup is in flight", () => {
+    state.project = { data: { workspace_id: "w1" }, isPending: false };
+
+    const { result } = renderHook(() => useRetention("p1"));
+
+    expect(result.current.retentionDays).toBeUndefined();
+  });
+
+  it("reports the plan's window once the workspace resolves", () => {
+    state.project = { data: { workspace_id: "w1" }, isPending: false };
+    state.workspace = { data: { billingPlan: PlanType.PRO }, isPending: false };
+
+    const { result } = renderHook(() => useRetention("p1"));
+
+    expect(result.current.retentionDays).toBe(90);
+  });
+
+  it("fails closed for an unrecognized plan string", () => {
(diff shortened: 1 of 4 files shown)


ka1kqi (approved):

ka1kqi:
https://github.com/user-attachments/assets/21ae4887-84fc-4f0a-b11d-9351c9ec33b6

after, sets to correct time filter. Has a flash on dev but may not be a problem on prod

XinweiHe (approved):

Issue #1824, opened 2026-08-11
feat(ui): filter traces by a metadata key, at trace or span scope


lqiu03:
### Problem

Both SDKs let a user attach arbitrary metadata to a trace or a span, and the detail panel renders it — so a user can see that a trace carries tenant: acme or feature: checkout-v2. But there is no way to find the other traces carrying it. Metadata is stored as a ZSTD-compressed JSON string (traces.metadata, spans.metadata), which cannot be filtered without decompressing and parsing every row, so the trace list offers no metadata filter at all.

### Proposed Solution

Filter the trace list by a metadata key and value, matching a key set at either scope, with the available keys discoverable rather than remembered.

Metadata becomes queryable by deriving a Map column from the JSON already stored, rather than by changing what ingestion writes — so nothing about the write path or the SDKs changes, and history is filterable once backfilled. One parameterized keyed field carries every key, because keys are the user's data and the registry is a fixed contract.

### Additional Context

Related references:
- Migrations: `backend/db/clickhouse/migrations/`
- Filter registry and lowering: `backend/rest/services/filters/`
- Filter UI: `frontend/ui/src/features/filters/`
- Span detail panel: `frontend/ui/src/features/traces/components/SpanInfoPanel.tsx`

### Willing to Help Implement?

- [x] I would like to work/help with this

PR #1834, opened 2026-08-11
fix(rest): bound the trace list queries and keep them off the event loop

lqiu03:
Part of #1824

> [!NOTE]
> Stacked on #1833 (`feat/keyed-metadata-lowering`).

## Summary

- Both trace list queries — the page query and the count query — now run with the shared
  `READ_QUERY_SETTINGS`: `readonly`, a 10s execution ceiling, and the GROUP BY spill
  threshold.
- `list_traces` moves off the event loop via `asyncio.to_thread`.

**Why this is the PR that forces it.** The keyed metadata predicate is the first span-level
filter that cannot use the spans no-I/O projection: `metadata_map` is deliberately absent
from it, and a projection only serves a query whose referenced columns it all carries. So
the predicate drops to the base table, whose sort key prunes by time only weakly, and it
runs **twice per request** — once in the page query and once in the count. A scan that wide
has to hit an execution cap rather than run until the client gives up.

The event-loop half is the same fact from the other side: `clickhouse-connect` is
synchronous, the REST service runs a single uvicorn worker on a single replica, so awaiting
the widest scan the dashboard issues inline stalls every other request in the process. There
is no second process to absorb it.

## Type of Change

- [x] fix - A bug fix

## Details

About two dozen production lines. Tests carried with them:

- The page query and the count query each assert they run under the shared settings, and
  that both carry an execution ceiling and the `readonly` flag.
- An off-loop suite with a sentinel that fails if the read is left on the loop, plus a test
  proving the sentinel would actually catch that (a guard that cannot fail is not a guard).
- An exception raised inside the worker thread still maps to a 500.
- A malformed predicate is rejected before any thread is taken.

No API change; no response shape change.

## Notes

- Nothing here is specific to metadata — both list queries get the bounds, including the
  unfiltered path. The keyed predicate is what made the ceiling load-bearing rather than
  precautionary.
- The settings module itself lands in #1832 with the discovery split.

Not applicable — backend only.

## Checklist

---
## Summary by cubic
Bound the trace list page and count queries with shared read settings (10s timeout, readonly) and moved blocking ClickHouse reads off the event loop via `asyncio.to_thread`. No API or response shape changes.

- **Bug Fixes**
  - Applied `READ_QUERY_SETTINGS` to both queries (`max_execution_time=10s`, `readonly=1`).
  - Called `service.list_traces` via `asyncio.to_thread` to keep `clickhouse-connect` off the event loop and avoid stalls on a single `uvicorn` worker.
  - Tests: assert settings on page/count; verify off-loop execution with a loop sentinel; 500 on internal errors; 422 for malformed filters.

<sup>Written for commit 63ea7e8377476bed4f862804bc9ede256a6c891d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1834?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/rest/routers/traces.py
+++ b/backend/rest/routers/traces.py
@@ -1,5 +1,6 @@
 """Trace query endpoints (user-authenticated, not public API)."""
 
+import asyncio
 import logging
 from datetime import datetime
 from typing import Literal
@@ -90,7 +91,13 @@ async def list_traces(
         raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_CONTENT, detail=str(e)) from e
     try:
         service = get_trace_reader_service()
-        result = service.list_traces(
+        # Off the event loop: clickhouse-connect is synchronous, and this is the widest
+        # scan the dashboard issues (a filtered list runs a page query and a count query
+        # back to back, both capped at the shared execution timeout). Awaiting it inline
+        # would stall every other request in the process — the REST service runs a single
+        # uvicorn worker, single replica, so there is no second process to absorb it.
+        result = await asyncio.to_thread(
+            service.list_traces,
             project_id=project_id,
             page=page,
             limit=limit,
--- a/backend/rest/services/trace_reader.py
+++ b/backend/rest/services/trace_reader.py
@@ -4,6 +4,7 @@ import time
 from datetime import UTC, datetime, timedelta
 
 from db.clickhouse import get_clickhouse_client
+from db.clickhouse.query_settings import READ_QUERY_SETTINGS
 from rest.services.filters.translate import Predicate, build_conditions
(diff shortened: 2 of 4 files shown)


lqiu03:
@cubic-dev-ai quickly review plz

XinweiHe (approved):

PR #1835, opened 2026-08-11
feat(ui): filter traces by a metadata key

lqiu03:
Part of #1824

> [!NOTE]
> Stacked on #1834 (`perf/trace-list-query-bounds`). This is the first PR in the stack
> with a user-visible change.

## Summary

The metadata filter end to end, with the key typed by hand.

- Pick **Metadata** in the filter builder, type a key, pick `=` or `contains`, type a value,
  Add filter. The row stays a single line — the popover's minimum width goes up so the
  fifth control fits.
- `Predicate` gains an optional `key` slot. `isValidPredicate` requires a non-empty key of
  at most 256 characters on a `requires_key` field and rejects a stray key on any other
  field, so a hand-edited URL can never put a filter into state that 422s every subsequent
  list request.
- `canonicalizeFilters` includes the key: `metadata.session_id = x` and
  `metadata.user_id = x` select different traces and must not share a query cache entry. An
  absent key stringifies away, so keyless predicates keep the shape their existing cache
  entries were written under.
- Chips read `metadata.tenant = acme`. The key is appended by `predicateLabel` itself, and
  the remove button names `metadata.tenant` rather than just `metadata`, so two metadata
  chips are distinguishable to assistive tech instead of both reading "Remove metadata
  filter".
- **Chips merge per key, not per field.** `upsertPredicate` slots on field *and* key, so
  `metadata.tenant = acme` and `metadata.feature contains checkout` coexist as two
  independent filters rather than one replacing the other — matching the backend, which
  lowers each metadata predicate as its own condition.
- `STATIC_FILTER_FIELDS` gains the metadata entry with `requires_key`. The static registry
  is the client's source of truth for that flag because URL predicates are validated before
  any `/filter-fields` fetch resolves.
- Metadata gets an icon in the shared domain icon registry.
- **Backend (about 20 lines):** removes the temporary `/filter-fields` gate added in
  #1833, adds `FilterField.requires_key` and serializes it, plus its test.
  `/filter-fields` now advertises metadata, which is exactly when the UI can handle it.

## Type of Change

- [x] feat - A new feature

## Details

**Any key the user types is accepted.** The key reaches ClickHouse as a bound parameter
rather than an identifier, which is what makes an arbitrary key safe to query — an
unrecognized key returns no rows, which is a correct empty result and not a reason to refuse
the filter. Nothing in this PR validates a key against a list.

**The key control here is a plain text input** (`aria-label="metadata key"`, capped at the
mirrored `MAX_KEY_LENGTH`). #1837 replaces it with a combobox that suggests the keys
discovered in the active window; the typed-key path it sits on does not change, and
if the suggestion work is dropped this input is what ships permanently.

The 256-character cap is mirrored from the backend deliberately: admitted into state, an
over-long key would persist into the URL and into storage and then 422 every list request
until the user found and removed it by hand.

## Notes

- Ordering: this PR must merge after #1833, since it removes that PR's gate. The
  two are adjacent in the stack for exactly that reason — every merge point stays
  deployable.
- The predicate, registry, chip, and builder logic all carry their own unit suites in this
  diff.

1. Filter popover open with **Metadata** picked: field dropdown, empty key box, operator
   dropdown, value box and Add filter, all on one row.
2. The same row filled in — key `tenant`, operator `=`, value `acme` — just before Add
   filter.
3. After applying: the chip reading `metadata.tenant = acme` in the filter bar, with the
   narrowed trace list and its result count behind it.
4. Two metadata chips at once (`metadata.tenant = acme` and
   `metadata.feature contains checkout`), showing they coexist rather than replace each
   other.
5. A key that was attached at **span** scope only, applied, matching traces in the list —
   the either-scope behavior, which is the part a reviewer cannot see from the diff.

## Checklist

---
## Summary by cubic
Add metadata key filtering to the trace list. Users can filter by `metadata.<key>` with `=` or `contains`, and the UI/URL enforce backend caps on value length and filter count to avoid 422s.

- **New Features**
  - Filter builder: shows a “metadata key” input when `metadata` is selected (trimmed, Enter applies, max 256 chars). Value is free text; operators: `=` and `contains`. Popover min width increased so the row stays on one line.
  - Chips and labels: `predicateLabel` appends the key (`metadata.session_id = ...`); remove buttons name `metadata.<key>` so multiple metadata chips are distinguishable.
  - Canonicalization and upserts: cache keys include `key`; `upsertPredicate` merges per field+key so different metadata keys coexist, same key replaces. Refuses adding a 21st chip.
  - Registry/API/icons: `STATIC_FILTER_FIELDS` adds `metadata` as `KEYED_MAP` with `requires_key`; `FIELD_ICONS`/`DOMAIN_ICONS` add a `metadata` icon. Backend `GET /traces/filter-fields` now advertises `metadata` and serializes `requires_key`; the temporary keyed-field gate is removed.

- **Validation and limits**
  - `Predicate` supports an optional `key`; keyed fields require a non-empty key and stray keys are rejected.
  - String values are capped at 1024 chars (text `=`/`contains` and each `in` element); the text input clamps this. Keys are capped at 256 chars.
  - Active filters are capped at 20: parse/serialize truncate to 20; `upsertPredicate` rejects additions past the cap but still allows edits that replace an existing one.

<sup>Written for commit 5fd2294e357bbc49ee2b45339cc70eb3e0c312ba. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1835?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/rest/routers/traces.py
+++ b/backend/rest/routers/traces.py
@@ -151,14 +151,9 @@ async def get_filter_fields(
                 "value_source": c.value_source,
                 "enum_values": list(c.enum_values),
                 "integer": c.is_integer,
+                "requires_key": c.requires_key,
             }
             for c in filter_columns.FILTER_COLUMNS
-            # Temporary: a keyed field needs a key control the filter builder does not
-            # have yet. Advertising it now would let a deployed builder render it as a
-            # plain text field and emit a keyless predicate, which the translator rejects
-            # with a 422 — and a 422 on the filter parameter sinks the whole trace list.
-            # Removed by feat/metadata-filter-ui, which ships the key control.
-            if not c.requires_key
         ]
     }
 
--- a/backend/rest/schemas/traces.py
+++ b/backend/rest/schemas/traces.py
@@ -139,6 +139,10 @@ class FilterField(BaseModel):
     # True for integer-typed numeric fields (tokens/latency/errors), so the UI can
     # restrict their inputs to whole numbers.
     integer: bool = False
+    # True for the one parameterized field (metadata), whose predicate carries a key as
+    # well as an operator and a value. The builder renders a key control only for these,
+    # so the one-row filter shape holds for every other field.
+    requires_key: bool = False
 
 
 class FilterFieldsResponse(BaseModel):
--- a/frontend/ui/src/components/icons/domain-icons.ts
+++ b/frontend/ui/src/components/icons/domain-icons.ts
@@ -2,6 +2,7 @@ import type { LucideIcon } from "lucide-react";
 import {
   ArrowRight,
   Bot,
+  Braces,
   BotMessageSquare,
   Box,
   CircleAlert,
@@ -61,6 +62,8 @@ export const DOMAIN_ICONS = {
   workspace: LayoutGrid,
   detector: Eye,
   environment: Globe,
+  // Free-form key/value data the user attached to a trace, not a traceroot field.
+  metadata: Braces,
   dashboard: LayoutDashboard,
   // Neutral "unknown field" fallback for filter/widget dropdowns. Kept
   // decoupled from `model` even though both currently render as Box — if the
--- a/frontend/ui/src/features/filters/filter-builder.test.tsx
+++ b/frontend/ui/src/features/filters/filter-builder.test.tsx
@@ -50,6 +50,19 @@ const MODEL: FilterFieldDef = {
   enum_values: [],
 };
 
+const METADATA: FilterFieldDef = {
+  field: "metadata",
+  label: "Metadata",
+  type: "text",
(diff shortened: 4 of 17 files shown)


lqiu03:
@cubic-dev-ai quickly review

XinweiHe (approved):

PR #1836, opened 2026-08-11
feat(rest): expose metadata keys for filter discovery

lqiu03:
Part of #1824

> [!NOTE]
> Stacked on #1835 (`feat/metadata-filter-ui`).

## Summary

`GET /projects/{project_id}/traces/metadata-keys` → `{"keys": [{"value", "count"}]}`, the
discovery answer behind the metadata filter's key control.

- **Both scopes.** A trace-level key can never reach a span, so the two key spaces are
  disjoint and scanning one table alone would leave the other's keys unsuggestable while a
  filter on either is equally valid.
- The two halves are `UNION ALL`ed **below** the aggregation, so one `GROUP BY` sums a key's
  trace and span occurrences and the ordering and the cap apply to the merged answer.
  Capping each half first would drop a key that is frequent overall but top of neither list.
- The key fan-out (`arrayJoin(mapKeys(metadata_map))`) sits in its own layer **above** the
  ReplacingMergeTree dedup. Inside the deduped sub-query, `LIMIT 1 BY` applies after the
  SELECT expressions are evaluated, so every row would contribute exactly one of its keys.
- Window-bounded through the same `_resolve_scan_window` / `_window_scan` pair the
  distinct-value scans use: a defaulted lookback rather than an all-time scan, an exclusive
  upper bound symmetric with the list's window, detector self-traces excluded.
- Capped at 100, cached for 30s in its own cache namespace, and run off the event loop —
  it arrayJoins over `mapKeys` on both tables' base rows, so it is the heaviest of the
  discovery reads, and the cache does not protect the first caller.
- Retention clamping, rate-limit bucket, and project access exactly as the sibling
  discovery endpoints.

## Type of Change

- [x] feat - A new feature

## Details

**Suggestion is not permission.** Unlike `/filter-values/{field}` there is no field to
resolve and nothing to reject: the key binds as a query parameter, so this endpoint only
suggests, and a key absent from its answer stays filterable by typing it. That is stated in
the endpoint docstring so a future reader does not mistake it for a whitelist.

The response reuses `FilterValueCount` (`value` / `count`) rather than defining a parallel
row type — a key with an occurrence count is the same shape as a categorical value with one,
and the same suggestion list renders both.

Tests cover the endpoint (both scopes contributing, frequency order, the cap, the cache, the
window default, retention clamping) and that the read runs off the event loop.

## Notes

- No UI consumer in this PR; the combobox that calls it is #1837. The trace list's
  column picker is explicitly not a consumer — it offers fixed fields only, and metadata
  reaches the list as a single cell rather than a column per key.

Not applicable — backend only, no UI consumer yet.

## Checklist

### DIFF
--- a/backend/rest/routers/traces.py
+++ b/backend/rest/routers/traces.py
@@ -26,6 +26,7 @@ from rest.routers.deps import RateLimitedProjectAccess
 from rest.schemas.traces import (
     FilterFieldsResponse,
     FilterValuesResponse,
+    MetadataKeysResponse,
     SpanIOResponse,
     TraceDetailResponse,
     TraceListResponse,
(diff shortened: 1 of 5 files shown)


lqiu03:
@cubic-dev-ai quickly review

XinweiHe (approved):

PR #1837, opened 2026-08-11
feat(ui): suggest metadata keys in the filter builder

### LINKED ISSUE #1824: feat(ui): filter traces by a metadata key, at trace or span scope

lqiu03:
Fixes #1824

> [!NOTE]
> Stacked on #1836 (`feat/metadata-key-discovery`). Last PR of the #1824 stack — this one
> carries the `Fixes`.

## Summary

Keys become discoverable rather than remembered, which is the clause of #1824 the typed-key
PR left open.

- `MetadataKeyCombobox` replaces the plain key input: it **suggests** the keys discovered in
  the active window and **accepts any key typed**. There is no custom-key mode, no validation
  that rejects an unsuggested key, and nothing in it can disable Add filter.
- The text box *is* the key. Typing edits the predicate's key directly and doubles as the
  suggestion search, so a key typed and never confirmed against the list is still the key
  that gets filtered on — a separate search box would silently discard it.
- Suggestions come from `useMetadataKeys` over the active window, with the same 30s
  staleness and the same window-keyed cache as the categorical value dropdown, so picking a
  suggestion can never return zero rows purely because the key was observed outside the
  range on screen.
- The backend's frequency order is preserved through filtering — no second ranking model
  here. An exact match stops narrowing, so reopening the list after picking `session_id`
  still offers the other keys rather than only the one already in the box.
- Enter dismisses the list and applies the row; the typed key needs no confirmation.
- Empty list shows the same "No options" line the categorical value dropdown has always
  shown. A key list still loading and one that came back empty look alike there too, and an
  empty list blocks nothing.
- `getMetadataKeys` client function plus `MetadataKey` / `MetadataKeysResponse` types.
  `MetadataKey` aliases the distinct-value row rather than copying it, since the backend
  serves both from one model.
- The dropdown panel and empty-state classes are extracted so the key combobox drops the
  same surface as the field and value dropdowns.

## Type of Change

- [x] feat - A new feature

## Details

Focus handling is the fiddly part and is commented in place: focus stays in the text box
while the list is open and is not pulled back on close, because returning focus would refire
`onFocus` and reopen the list immediately.

`useMetadataKeys` takes an `enabled` flag, but that is not where the laziness comes from —
every call site passes it true, and the fetch stays lazy because the combobox only mounts
for a metadata predicate. That is the opposite of `useFilterValues` beside it, where
`enabled` is the real gate on an always-mounted control, so the difference is documented on
the hook.

## Notes

- With this, #1824 is complete: metadata is filterable at either scope, by a key the user
  can discover instead of recall, with an unlisted key still typable.
- Nothing about the typed-key path changes; this swaps the control the key is typed into.

1. Metadata picked, key box focused, suggestion list open showing the discovered keys with
   their occurrence counts, in frequency order.
2. The same list after typing a partial key — narrowed to the matching keys, order
   preserved.
3. A key chosen from the list, filled into the box, with the operator and value completed
   and Add filter ready.
4. A key typed that is **not** in the suggestion list, applied — chip present and the list
   narrowed. Suggestion is not permission.
5. The list reopened after a key was picked, still offering the other keys rather than
   narrowing to the one already chosen.

## Checklist

---
## Summary by cubic
Adds a metadata key combobox to the filter builder that suggests discovered keys from the active time window while still accepting any typed key. Completes #1824 by making metadata keys discoverable instead of remembered.

- New Features
  - Replaced key input with `MetadataKeyCombobox` using `useMetadataKeys(projectId, startAfter, endBefore, enabled)` (30s cache); preserves backend frequency order; exact match stops narrowing; shows counts; Enter applies; typed keys always valid.
  - Added API `getMetadataKeys` and types `MetadataKey`/`MetadataKeysResponse` (keys list only); request supports `start_after`/`end_before`; shared dropdown/empty-state styles extracted and reused; tests cover combobox behavior, empty-state parity, focus/enter handling, window scoping/enablement, and API URL building.

<sup>Written for commit a646a22d3da4719f713b12feebe6499e14004246. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1837?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/ui/src/features/filters/filter-builder.test.tsx
+++ b/frontend/ui/src/features/filters/filter-builder.test.tsx
@@ -2,14 +2,41 @@
 import { afterEach, describe, it, expect, vi } from "vitest";
 import { render, cleanup, screen, fireEvent } from "@testing-library/react";
 import { FilterBuilder } from "./filter-builder";
-import type { FilterFieldDef } from "./registry";
+import { ValueDropdown } from "./filter-controls";
+import { isValidPredicate } from "./predicate";
+import type { FilterFieldDef, FilterValue } from "./registry";
 
-// Stub the distinct-values hook; keep it as a spy so we can assert the window bounds
-// are threaded through to it for a distinct-query categorical field.
+// Spies, not bare stubs, so tests can assert the window bounds each hook is asked about.
 const mockUseFilterValues = vi.hoisted(() => vi.fn(() => ({ values: [], isLoading: false })));
-vi.mock("./hooks", () => ({ useFilterValues: mockUseFilterValues }));
+// A discovery row IS the categorical distinct-value row (`{value, count}`), so the stub
+// cannot spell the field a way the endpoint does not.
+const mockUseMetadataKeys = vi.hoisted(() =>
+  vi.fn(
+    (
+      _projectId: string,
+      _startAfter?: string,
+      _endBefore?: string,
+      _enabled?: boolean,
+    ): { keys: { value: string; count: number }[]; isLoading: boolean } => ({
+      keys: [],
+      isLoading: false,
+    }),
+  ),
+);
+vi.mock("./hooks", () => ({
+  useFilterValues: mockUseFilterValues,
+  useMetadataKeys: mockUseMetadataKeys,
+}));
 
-afterEach(cleanup);
+/** Discovery answers with these keys, in the order given (frequency-descending). */
+function suggestKeys(keys: FilterValue[]) {
+  mockUseMetadataKeys.mockReturnValue({ keys, isLoading: false });
+}
+
+afterEach(() => {
+  cleanup();
+  mockUseMetadataKeys.mockReturnValue({ keys: [], isLoading: false });
+});
 
 const STATUS: FilterFieldDef = {
   field: "status",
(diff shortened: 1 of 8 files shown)


lqiu03:
@cubic-dev-ai quickly review plz

XinweiHe (approved):

Issue #1844, opened 2026-08-11
Offline-eval UI redesign: N-run comparison, simplified run detail, version-scoped dataset rows


XinweiHe:
## Summary

The offline-eval UI surfaces (Experiments list, Run detail, **Run Comparison**, and the Dataset-detail row panel) were iterated heavily and several earlier decisions were reverted in favor of a simpler, more consistent design. This issue tracks that redesign and its rollout.

The change is **frontend-only** (no SDK changes) and is net-large-deletion: it removes earlier experiments in favor of leaner surfaces.

## What changed

**Experiments list**
- Selectable rows + a bulk **Actions** menu (Compare / Delete) in the top-right, left of the date filter.
- Renamed/added columns to match a clean, consistent layout (Experiment Name, Run Name, Cost, Avg Cost, Duration, Avg Duration); dataset version shown as a time-sortable id.
- Compare is now driven from row selection (2+ runs) and guards against mixing different datasets.

**Run Comparison (new)**
- Replaces the fixed two-run *baseline vs candidate* view with an **N-run** comparison, lined up by dataset-row id (the intersection of rows present in every run), so runs on different dataset *versions* still align correctly.
- Baseline is chosen inline from a picker; each non-baseline run shows a green/red **improvement/regression delta** vs the baseline, in both the aggregate header and the per-row cells.
- Cells expand to full value on hover; per-run values are colour-keyed.

**Run detail**
- Removed the identity header and the inline "compare with" (comparison now lives on the list).
- Per-case results table with one column per scorer and horizontal scroll.

**Dataset detail — row panel**
- The row's **Experiments** tab now mirrors the Experiments-list columns and is **scoped to the dataset version being viewed** (runs that measured a different version of the row are not shown).

**Cleanup**
- Removed the trace input/output +/- text-diff feature and the standalone two-run comparison API endpoint/engine artifacts that the new design no longer needs.

## Rollout — 3-PR stack on top of #1782

1. **Experiments list + N-run Run Comparison** — row selection, bulk actions, the new compare page + data hooks; removes the old two-run compare machinery.
2. **Run detail simplification + trace-diff cleanup** — strips the run-detail header/inline-compare, adds scorer columns; deletes the now-dead trace diff plumbing (depends on #1).
3. **Dataset detail: version-scoped row Experiments tab** — the version-filtered runs table + the supporting read endpoint.

Each PR is intentionally deletion-heavy and independently green.

## Notes
[...]

PR #1782, opened 2026-07-31
test(evaluations): view smoke-test suite

dark-sorceror:
## Summary

Fixes: #1671

Mount each real evaluation surface and assert its key states and core interactions so the UI can't silently regress. This is the frontend counterpart to the backend suite and what lets every UI change land safely. The tests exercise the real views against the read-model shapes, so a drift between the APIs and the UI fails a test rather than shipping. The suite runs in CI alongside the rest of the frontend tests.

### Tests
- `frontend/ui/src/features/evaluations/evaluations.smoke.test.tsx` — mounts the server-backed Datasets and Evaluations pages and asserts loading / empty / error / populated states, the SDK-pointing empty copy, and the run-centric grouping / latest / history behavior.
- `frontend/ui/src/features/evaluations/run-detail-compare.smoke.test.tsx` — picking another run of the same evaluation computes the comparison via the shared engine, shows the comparison banner, drives the metrics and per-cell diffs and the `≠ baseline` marks, and clearing restores the plain view with the change column removed.
- `frontend/ui/src/features/evaluations/run-detail-trace.smoke.test.tsx` — opening a result opens the real ingested trace directly when available, and shows the pending (retry) state while it is still ingesting.
- `frontend/ui/src/features/evaluations/compare-runs.smoke.test.tsx` — the shareable comparison page against a fixture with a regressing case: baseline-left orientation, the aggregate verdict, and the per-case regression surfaced.
- `frontend/ui/src/features/evaluations/scorer-detail.smoke.test.tsx` — the read-only scorer detail for both an LLM judge (model + prompt) and a code scorer (highlighted source), plus an undeclared scorer's honest em-dash gaps.
- `frontend/ui/src/features/evaluations/save-test-case-drawer.smoke.test.tsx` — the full capture flow (dataset picker with a New-dataset option, selected span, read-only recorded output, the expected-outcome field, metadata).
- `frontend/ui/src/features/evaluations/save-test-case-format.smoke.test.tsx` — per-field seed formats: the mode a captured value opens in is a property of the field.
- `frontend/ui/src/features/evaluations/compare-derive.test.ts` — covers `compare-derive`.
- `frontend/ui/src/features/evaluations/datasets-detail.smoke.test.tsx` — covers `datasets-detail.smoke`.
- `tests/db/test_clickhouse_migrate.py` — covers `clickhouse_migrate`.

### UI
[...]

PR #1845, opened 2026-08-11
feat(evaluations): row-selection compare + N-run Run Comparison; simplify run detail

XinweiHe:
Part **1 of 3** of the offline-eval UI redesign (tracking: #1844), stacked on top of #1782.

## What
- **Experiments list**: selectable rows + a bulk **Actions** menu (Compare / Delete); avg cost & duration columns; dataset version shown as a time-sortable id; a cross-dataset selection is refused with a warning toast.
- **Run Comparison** (new): replaces the fixed two-run baseline/candidate view with an **N-run** comparison, lined up by dataset-row id (the intersection of rows every run measured). Baseline is chosen inline; each non-baseline run shows a green/red improvement/regression delta in both the aggregate header and the per-row cells; cells reveal full text on hover.
- **Run detail**: drops the identity header + inline "compare with" (comparison now lives on the list) and gains one column per scorer.

## Removes
The old two-run compare machinery — the `/evaluations/compare` route, the `compare-derive` engine helpers, the case drawer — and their tests. Net large deletion.

## Stack
1. **this PR** — list + N-run compare + run detail
2. trace-diff cleanup
3. dataset-detail Experiments tab

Frontend only; no SDK changes.

---
## Summary by cubic
Replaces the two-run baseline/candidate view with an N-run comparison driven from the Experiments list, and simplifies Run Detail. Frontend-only; removes the old compare API and case drawer for a cleaner workflow.

- **New Features**
  - Experiments list: row selection + bulk Actions (Compare, Delete), avg cost/duration, dataset version shown as a time-sortable snowflake ID, cross-dataset selection blocked with a toast.
  - Run Comparison: N-run grid aligned by dataset-row id with per-run stacked metric values; pick baseline inline; green/red deltas in header and cells; Input collapses when runs agree; shareable route `.../evaluations/compare?runs=<id,id,…>&baseline=<id>`.
  - Run Detail: identity header and inline “compare with” removed; one column per scorer; leaner table; input/outputs reveal full text on hover.
  - Delete run dialog component; list uses delete-only actions via an optional `onEdit` in the actions menu.

- **Refactors**
  - Removed the old two-run compare machinery: `app/api/.../evaluations/compare` route, `compare-derive`, case drawer, and related tests.
  - Simplified types and hooks; adopted list-driven comparison and `useQueries`.
  - Added `lib/eval/snowflake.ts` for derived, time-sortable dataset-version IDs.

<sup>Written for commit 58d7cf506759f67b4e35025e6ccf5cf4ba8d63cb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1845?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


dark-sorceror:
On two of the HIGH items in the Cubic review — both look **already addressed / stale** against the current HEAD (`58d7cf506`):

**Compare truncation** (`compare-runs-view.tsx`): it isn't silent — the view already renders an amber banner when a run exceeds the run-detail result cap, stating explicitly that the duration/cost/scorer aggregates and deltas are *partial* (the `truncatedRuns` guard, `caseCount > results.length`). Full-result pagination or a hard-fail would be a fair enhancement, but the current disclose-and-continue behavior isn't a silent data-loss bug.

**Delete → 405** (`hooks.ts` / `evaluations-view.tsx`): a `DELETE` handler does exist at `app/api/projects/[projectId]/evaluations/runs/[runId]/route.ts` (`export async function DELETE`, present since `ea6aedcfd`), and `useDeleteRuns` issues its `DELETE` to that exact URL — so run deletion doesn't 405. This one reads as a cross-file route-detection miss.

Flagging so these two aren't treated as open blockers on the stack; happy to look again if I've missed a path.

PR #1884, opened 2026-08-13
fix(detectors): evaluate every trigger field at the trace list's grain

lqiu03:
Part **1 of 3** of the detector trigger-field stack. Independent of the other two — this one stands alone against main. #1885 and #1886 are stacked on it.

A detector trigger could be configured on any of the seven filterable fields, but the summary the evaluator matched against was built from **the root span only**. Every condition on a non-root field read as missing, and a missing field answers `True` only for `!=` and `False` for everything else. The result is **a detector that is configured, enabled and never fires** — silently, with nothing in the logs to suggest it.

This builds the trace summary at the trace list's grain instead:

- Wall-clock latency across the trace, not the root span's duration.
- ERROR-span count and per-trace sums for cost and tokens.
- Metadata as a membership set folded from the trace-scope map and every span-scope map, so a key attached at either scope matches — the same answer the trace list's keyed-map lowering gives.

Backend only. No UI or API surface changes; the two PRs above it build on this.

---

Stack: **#1884** (evaluator) → **#1885** (registry) → **#1886** (editor)

### DIFF
--- a/backend/worker/detector_tasks.py
+++ b/backend/worker/detector_tasks.py
@@ -15,7 +15,9 @@ import asyncio
 import hashlib
 import json
 import logging
+import operator
 import uuid
+from decimal import Decimal
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


lqiu03:
Reopened as #1887 from the upstream branch so the stack chains to main. The fork-based head here broke the link to #1885.

Issue #1877, opened 2026-08-12
[Epic] Agent detector reads — detectors, findings, and RCA through the shared registry


ka1kqi:
Give the in-app agent (and API-key clients) read access to everything a user can read in the UI about detectors and their output — the catalog, per-detector configuration, findings, and root-cause analyses — as the first capability family added end-to-end through the shared tool registry (#1736).

## Why

The agent could already browse traces and sessions, but the product's own diagnosis layer — detectors and their findings/RCA — was invisible to it. Wiring these reads through the registry means one curated definition per tool serves the agent (internal surface, service auth) and API-key clients (public surface) with no per-surface handler code, exercising exactly the acceptance path #1736 established.

## Sub-issues (each ≈ one PR, stacked in order)

- [ ] #1875 — feat(agent): drive detector and finding reads from the shared registry
- [ ] #1876 — feat(api): get_detector — one detector's full configuration for the agent and API clients

## Acceptance

- The agent answers "what detectors do I have", "what does detector X check for", "any findings today", and "what's wrong with trace T and why" from live project data, through registry-driven dispatch.
- The same registry entries serve the public API; public params remain a subset of each internal twin's, CI-enforced.

## Out of scope (tracked later, not filed yet)

- Pagination: DECIDED (2026-08-18) — typed lists stay windowed (limit + time range, uniform across traces/sessions/detectors/findings); no offset/cursor primitive. Exact bulk enumeration defers to the future governed query surface. #1919 closed with rationale; shelf implementation at closed PR #1936 (feat/detector-list-pagination @ 7c73bceb).
- Generated response types for registry entries — the agent formatters currently cast responses to `any`; extending the tools-package generator with response schemas would let codegen catch response-shape drift
- Detector *runs* read (executions including non-firing ones — needs a new public endpoint design; the internal runs endpoint requires a detector id)
- Detector counts / window summaries
- Consolidating the detector read path (single canonical read layer) and any write/mutation tools — the write path forces the config-ownership decision and is deliberately deferred

Satyam296:
Hey @ka1kqi, saw the two sub-issues + pagination already have PRs up (#1912/#1913/#1936) so looks well in hand! Of the deferred items at the bottom, the generated response types for registry entries (replacing the `any` casts in the agent formatters) looks like a concrete, scoped chunk is that open for someone else to pick up, or do you have a specific design in mind for it already?

ka1kqi:
[...]

PR #1913, opened 2026-08-13
feat(api): get_detector — one detector's full configuration for the agent and API clients

### LINKED ISSUE #1876: feat(api): get_detector — one detector's full configuration for the agent and API clients
Part of the agent detector reads epic #1877. Builds on the detector/finding read wiring (#1875).

## What

A `get_detector` read exposing one detector's full configuration end-to-end — to the agent and to API-key clients — so the agent can answer "what does this detector check for" and "why didn't it fire" from real config instead of guessing:

- prompt (detection criteria), output schema, sample rate
- enabled / RCA-enabled flags, detection model/provider/source
- trigger conditions (the run-gating rules), creation/update times

## How

- **Schema + service:** `DetectorDetail` extending the catalog item; a reader lookup on the Postgres catalog with a join to the detector's optional trigger row.
- **Routes:** public `GET /api/v1/public/detectors/{detector_id}` (curated into the tool registry, 404/500 error contract documented in the schema) plus the internal project-scoped twin, registered after the static `/findings` routes so the literal segment can never be shadowed by the path parameter (regression-tested on both surfaces).
- **Registry + agent:** regenerated registry entry, internal binding, parity-test row, `formatDetectorDetail` formatter (prompt truncated, JSON blobs bounded, "no trigger" stated explicitly), tool ordered next to `list_detectors`, prompt copy updated.
- **Drift guard:** a test asserting every column the detector reader's raw SQL references still exists in the Prisma schema, so a catalog migration breaks tests instead of runtime reads.

## Acceptance

- Agent flow: `list_detectors` → `get_detector` → `list_findings` / `get_finding` works against a live project; the same registry entry serves API-key clients on the public surface.
- Requesting the literal path `/detectors/findings` still hits the findings list on both surfaces; missing detector is a clean 404; reader failure a sanitized 500 (both pinned by tests).
- Diff coverage ≥ 80% on both sides.

ka1kqi:
## What

Adds a `get_detector` read end-to-end: one detector's full configuration — prompt (detection criteria), output schema, sample rate, enabled/RCA flags, detection model/provider/source, and trigger conditions — exposed to the agent and to API-key clients through the same registry pipeline. This lets the agent answer "what does this detector check for" and "why didn't it fire" from real config, and closes one of the read gaps deferred by the first detector-reads PR.

Also includes the agent-prompt hardening that came out of live testing:
- the detector **template catalog** in the system prompt, so coverage-gap questions get "add a detector with the hallucination template" instead of proposals to build external tooling
- a **live-data rule**: re-query current counts instead of answering from earlier results, and attribute deltas to new data arriving rather than inventing filter explanations

## How

- **Schema + service:** `DetectorDetail` extending the catalog item; `DetectorReaderService.get_detector` (Postgres, `LEFT JOIN detector_triggers`).
- **Routes:** public `GET /api/v1/public/detectors/{detector_id}` (curated `x-tool`, 404/500 error contract documented in the schema) plus the internal project-scoped twin via the shared handler-body module. Both registered after the static `/findings` routes; shadowing is regression-tested on both surfaces.
- **Registry:** regenerated `public.json` + `registry.generated.ts`, `INTERNAL_BINDINGS` entry, parity-test row.
- **Agent:** `formatDetectorDetail` (prompt shown, JSON blobs bounded, "no trigger" stated explicitly), tool at position 5 next to `list_detectors`, prompt copy.
- **Drift guard:** `tests/rest/test_detector_schema_drift.py` asserts every column the reader's raw SQL references still exists in the Prisma schema, turning silent catalog-migration drift into a test failure.

## Verification

- Full backend suite green (incl. OpenAPI artifact `--check`); tools 27/27, agent 59/59; tsc clean; diff coverage vs main: backend 93%+, frontend 100%.
- Live-verified against a locally seeded project: config rendering (including the null-model → "default (system)" path), route shadowing, date-bounded finding queries, and template-aware coverage answers.

## Notes

- Stacked on the detector/finding reads PR (base branch of this PR); merge that first.
- PR #1843 (current-date injection) was **merged into this branch** by a maintainer, so its date line + test ship as part of this diff — intentional stack completion, not scope creep.
- origin/main (offline-eval epic + detector-trigger work) has been merged in; the curation dict, OpenAPI artifact, and registry were regenerated as the union.
- Known overlap with #1862 (metadata key discovery) in the pinned tool lists and registry artifacts; whichever lands second rebases.
- Follow-up candidate: the prompt's template list is hand-duplicated from the UI catalog — a sync assertion would prevent drift (one stale entry was already caught and removed in review).

Closes #1876. Part of epic #1877.

## Video

https://github.com/user-attachments/assets/449f29c5-6303-4761-a5f8-041c1d4166b0

---
## Summary by cubic
Adds get_detector so agents and API clients can fetch one detector’s full configuration, and updates the agent prompt/tools to use it. Old: clients could only list detectors and evaluation traces appeared in production lists. New: GET one detector with full config; evaluation writes live behind the same host via a gateway; trace/session/user lists exclude evaluations by default with an opt‑in flag. Side effects: ClickHouse adds is_evaluation; the worker runs Celery beat for evaluation cost backfill.

**Detector detail + agent updates**
- Endpoints: public GET /api/v1/public/detectors/{detector_id} and the project‑scoped internal twin, registered after static /findings and /traces, sharing the read rate‑limit bucket. Maps missing to 404 and reader errors to a sanitized 500; OpenAPI documents both and exposes an x‑tool entry.
- Service/schema: DetectorDetail adds prompt, output_schema, sample_rate, RCA, detection model/provider/source, trigger_conditions, and updated_at; reader does a LEFT JOIN on detector_triggers; require_detector maps None→404 and exceptions→500 and runs off the event loop. Tests pin public/internal parity and guard the raw SQL against Prisma schema drift.
- Tools + prompt: `@traceroot-ai/tools` generates and binds get_detector (plus internal binding); the agent binds it and renders config with truncation and explicit missing pieces. The system prompt lists the detector template catalog from the UI, adds a live‑data re‑query rule, injects the current UTC date, and drops the stale “intent” entry; tests assert catalog sync and the exact date line.

**Rollout**
- Apply DB migrations (ClickHouse 011_add_is_evaluation; Prisma offline‑eval + detector defaults).
- Rebuild and redeploy the worker image to run Celery with --beat.

<sup>Written for commit 477a9c7122d13cc983650782490581324c8164cd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1913?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ka1kqi:
**Heads-up for whoever rebases/merges this against the CLI-login / user-credential auth work — `get_detector`'s auth needs reconciling; it won't merge cleanly as-is.**

This PR's public `get_detector` uses `auth: StampedAuth` (`from rest.routers.public.deps import StampedAuth`), which today wraps `authenticate_api_key` — i.e. **key-only**. The in-flight user-credential stack changes that surface in two ways:

1. **Rename.** That key-only alias is renamed `StampedAuth` → `KeyStampedAuth`, and a new `DualStampedAuth` (API key **+** user session token **+** CLI access JWT, via `authenticate_public_caller`) is introduced. So after that stack lands, `StampedAuth` no longer exists and this import/annotation breaks.

2. **Consistency + the actual capability gap.** The other public detector reads (`list_detectors`, `list_findings`, `get_finding`, `get_finding_by_trace`) are migrated to `DualStampedAuth`. If `get_detector` stays key-only, you'd have — in the same `detectors_read.py` — four detector endpoints that accept a CLI/agent JWT and one (`get_detector`) that 401s it. Since this endpoint is explicitly "for the agent and API clients" and is added to the curated registry tool surface, that's a live gap: an agent/CLI holding a user credential could `list_detectors` but not `get_detector`.

**Reconciliation at merge time (either merge order):**
- Switch `get_detector` from `StampedAuth` → **`DualStampedAuth`** so it matches its siblings and accepts user credentials/JWT.
- Give its tests the dual-auth treatment — `test_public_detectors_read.py` here imports `authenticate_api_key`; add a session/JWT case like the sibling reads have.

The generated registry entry itself needs no change — it's auth-agnostic (name/method/path/inputSchema); only the endpoint's dependency does.

XinweiHe (approved):

PR #1557, opened 2026-07-14
fix(pricing): add missing Codex model prices

Gujiassh:
Fixes #1545

## Summary

- add the five Codex pricing entries missing from the standard model price table
- verify all six priced Codex models, including the existing table entry, against OpenAI's published API pricing
- cover bare, `openai/`-prefixed, and dated model IDs with exact pricing assertions
- leave the Spark research preview unpriced because it is not listed in the published API pricing table

## Type of Change

- [x] fix - A bug fix

## Details

The missing Codex model IDs returned no price and produced null trace costs. This change converts OpenAI's published per-million-token prices to the per-token units used by the table.

The Spark variant is a research preview without a published API price, so it is intentionally left unpriced instead of inheriting another model's rate.

Pricing source: https://developers.openai.com/api/docs/pricing

## Screenshots / Recordings (if applicable)

Not applicable; this changes backend cost lookup data and tests only.

## Checklist

- [x] I have read the [CONTRIBUTING.md](../CONTRIBUTING.md)
- [x] I have added/updated tests where applicable
- [x] I have added screenshots or recordings for UI changes (if applicable)
- [x] I have updated documentation where necessary

## Validation

- `.venv/bin/pytest tests/worker/tokens/test_pricing.py -q` (207 passed)
- `.venv/bin/ruff format --check tests/worker/tokens/test_pricing.py`
- `.venv/bin/ruff check tests/worker/tokens/test_pricing.py`
- `jq empty frontend/packages/core/src/standard-model-prices.json`
- `git diff --check`

PR #1914, opened 2026-08-13
feat: add Grok 4.6 to xAI model catalog

### LINKED ISSUE #1873: feat(byok): support Grok 4.6, new xAI frontier model, in TraceRoot BYOK modules
### Problem

xAI released Grok 4.6 (grok-4.6) on August 12, 2026 as its new frontier model for coding, knowledge work, and STEM, but it is missing from our curated xAI catalog. As a result, users see it as “(Unsupported)” in Settings > Model Providers > xAI.

 
### Proposed Solution

Add \{ id: "grok-4.6", label: "grok-4.6" \} to the xai entry in ADAPTER_MODELS (frontend/packages/core/src/llm-providers.ts) at index 0 so it lands newest-first and becomes the default-selected xAI model on Add Provider > xAI, matching the ordering convention used for grok-4.5 in #1485. label should equal id per #626.

Add a case in frontend/packages/core/src/__tests__/llm-providers.test.ts and verify it appears in the ModelProvidersTab.tsx dropdown and model selector without triggering the unsupported-model warning banner.

This is a catalog-only change: do not edit SYSTEM_MODELS, PROVIDER_PRIORITY, or ADAPTER_API_PROTOCOL. xAI remains on openai-completions against https://api.x.ai/v1, so no per-model apiProtocol override is needed.

### Alternatives Considered

None. This follows the existing curated-catalog pattern; no free-text fallback is desired for standard BYOK providers.

The grok-4.6-fast variant is deliberately out of scope. We do not currently carry -fast SKUs for any provider, and adding one is a separate decision from tracking the frontier model

 

### Additional Context

  - Announcement: https://x.ai/news/grok-4-6
  - Pricing: $2/MTok input, $0.50/MTok cached input, $6/MTok output below 200K prompt tokens;
    in the 200K+ long-context band those double to $4 / $1 / $12 and apply to every token in
    the request. The fast variant is 2x.
  - 500K context window.
  - Best practice is to follow the implementation and QA pattern from #1485 (grok-4.5) and
    PR https://github.com/traceroot-ai/traceroot/pull/610

- [ ] I would like to work/help with this

kashish00208:
Fixes #1873 

## Summary

Adds `grok-4.6` to the curated xAI model catalog so it is recognized as a supported xAI model instead of appearing as `(Unsupported)` in the provider settings.

This follows the existing implementation pattern used for Grok 4.5 and the curated BYOK model dropdown.

## Changes

* Added `{ id: "grok-4.6", label: "grok-4.6" }` to the xAI entry in `ADAPTER_MODELS`.
* Placed Grok 4.6 at index `0` so it appears newest-first and is the default-selected xAI model.
* Verified that Grok 4.6 appears in the xAI model dropdown/model selector.
* Verified that selecting Grok 4.6 does not trigger the unsupported-model warning.

## Scope

This is intentionally a catalog-only change.

* `SYSTEM_MODELS` was not modified.
* `PROVIDER_PRIORITY` was not modified.
* `ADAPTER_API_PROTOCOL` was not modified.

## Testing

* Verified the xAI model selector displays `grok-4.6`.
* Verified the unsupported-model warning is not shown for `grok-4.6`.

## Context

This follows the existing Grok 4.5 implementation pattern and the curated model catalog approach for BYOK providers.

## Type of Change

- [ x] feat - A new feature

---
## Summary by cubic
Adds `grok-4.6` to the curated `xai` model catalog and sets it as the default BYOK selection; previously it appeared as Unsupported and unpriced.

- Inserts `grok-4.6` at index 0 of the `xai` list and adds a `standard-model-prices.json` entry with input/output/cacheRead rates.
- Adds tests covering model list membership, id resolution (plain, `xai/`-prefixed, dated variants, and rejecting neighboring ids), published rates, and cost calculation.
- No migrations, config changes, or updates to `SYSTEM_MODELS` or `PROVIDER_PRIORITY`.

<sup>Written for commit 824e3fefca1c782a1e32aaa86d005ba031196d10. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1914?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lqiu03:
Hi @kashish00208 , I'm working on code-reviewing your pr. A few comments:
1. Adding the id and label in src/llm-providers.ts is not enough, please in src/standard-model-prices.json update your proposed LLM's id, model name, provider, token prices, etc
2. Please add a test case as requested in issue #1873 
3. Please add end-to-end testing screenshots for your change. You can do so by self-hosting your change locally and test connections with the LLM you're proposing adding.
Screenshot 2026-08-13 at 4 03 34 PM

frontend/packages/core/src/llm-providers.ts
```diff
@@ -149,6 +149,12 @@ export const PROVIDER_PRIORITY: LLMAdapter[] = [
 // Curated model catalog per adapter — used for dropdown selection in BYOK provider settings.
 // Adapters NOT listed here (azure, amazon-bedrock, openrouter) use free-text input.
 export const ADAPTER_MODELS: Partial<Record<LLMAdapter, LLMModelDef[]>> = {
+  xai: [
+    { id: "grok-4.6", label: "grok-4.6" },
```
lqiu03: Not sure why delete l195-199 and moving it to l152-157, is it necessary? Simplest change is add `{ id: "grok-4.6", label: "grok-4.6" }`

lqiu03 (changes requested):

kashish00208:
@lqiu03 Thanks for the review and for pointing these out. I misunderstood the scope of the change and only updated ADAPTER_MODELS, including moving grok-4.6 to the top. The move was not necessary ,  I’ll keep the existing models in place and simply add grok-4.6 as the newest entry.  I’ll push the  requested updates.

frontend/packages/core/src/standard-model-prices.json
```diff
@@ -1095,5 +1095,17 @@
       "cacheRead": 0.0000003,
       "cacheWrite": null
     }
+  },
+  {
+    "id": "tr-grok-4-6",
+    "modelName": "grok-4.6",
```
dark-sorceror, line 1199: This is the catalogue's first xAI entry and nothing tests it. Add resolution cases plus an absolute-rate assertion pinning `2e-06` / `6e-06` / `5e-07` — #1557 is the shape to copy.

dark-sorceror (changes requested):
Prices match xAI's published table — $2 / $6 / $0.50 with no cache-write column — and specificity is clean in both directions, checked against a simulated post-#1342 catalogue too. One thing below.

Note the >=200k-token tier doubles all three rates, which the catalogue has no way to express. Same standing gap as the rest of the catalogue, not a blocker on this PR.

frontend/packages/core/src/standard-model-prices.json
dark-sorceror: Added in 2a489f732, since the branch had gone quiet.

`TestGrok46ModelIds` now covers id resolution for the bare, `xai/` prefixed and dated forms, a specificity guard against `grok-4`, `grok-4.5`, `grok-4.60` and `grok-4.6-fast`, the whole-catalogue uniqueness check, an absolute-rate assertion pinning 2e-06 / 6e-06 / 5e-07 with `cacheWrite is None`, and a cost smoke test.

Mutation-checked: a uniform 1.5x drift on grok-4.6 reds `test_absolute_rates` and nothing else. Rates re-verified against the provider docs at $2 input, $6 output, $0.50 cached, with no cache-write column.

Named `GROK_4_6_CASES` and `TestGrok46ModelIds` rather than the `XAI_MODEL_CASES` and `TestXAIModelIds` that another open PR introduces. If both used the same class name, a conflict resolved by keeping both sides would import fine while the second definition silently shadowed the first, dropping a whole test class with no error. With distinct names the two files auto-merge clean, and only the price JSON conflicts.

dark-sorceror (approved):
The missing coverage is added and verified.

Rates match the published table at $2 input, $6 output, $0.50 cached input, with no cache-write column, so `cacheWrite: null` is right. Specificity is clean in both directions, checked against a simulated post-merge catalogue as well: `grok-4` does not absorb `grok-4.6`, and `grok-4.6` absorbs nothing it should not. The new absolute-rate assertion is non-vacuous, since a uniform 1.5x drift reds exactly that one test. 233 tests pass.

Worth flagging for whoever merges: this and the other open xAI pricing PR conflict in `standard-model-prices.json` only, because both append at the same tail anchor. The test file auto-merges clean. Resolve that one by hand rather than accepting both sides wholesale, since a naive splice leaves a missing comma and invalid JSON.

The 200k-plus prompt tier doubles all three published rates and the catalogue schema cannot express a tiered rate. That is a standing gap across every entry rather than anything this PR introduces.

XinweiHe (approved):

XinweiHe:
Let's fix the merge conflict before mering.

dark-sorceror:
Your merge resolution left `standard-model-prices.json` invalid, so the whole `tests/worker/tokens/` suite was failing to collect: 4 failed, 267 errors, since the `real_cache` fixture parses that file.

The splice dropped four lines where the conflict boundary cut through the middle of an object, fusing the `grok-4.6` and `grok-4.20` entries: `"cacheWrite": null`, the `}` closing `prices`, the `},` closing the entry, and the `{` opening the next one. Restored in d2aed3ea, which is the four lines and nothing else. Sorry for pushing to your branch without asking first.

Verified after the repair: 101 entries, `merged minus grok-4.6` equals main exactly, no duplicate `id`, `modelName` or `matchPattern`, and both `TestXAIModelIds` and `TestGrok46ModelIds` collect and pass with no shadowing. CI is green.

Still blocked by a changes-requested review from August about a block move you fixed the next day. That needs the reviewer or an admin to dismiss it.

dark-sorceror:
resolve merge conflicts

kashish00208:
> resolve merge conflicts

Appreciate you stepping in to patch the malformed JSON in standard-model-prices.json and verifying the suite run. Definitely no problem with pushing the repair commit to the branch.

Since CI is fully green now, could an admin or @lqiu03  dismiss the outdated review request from August so this is clear to merge?

PR #1936, opened 2026-08-17
feat(api): opaque keyset-cursor pagination for the detector list reads

### LINKED ISSUE #1919: feat(api): pagination contract for public list reads — no way to page past the first N results
Follow-up from the detector reads epic #1877 (surfaced in review of the read stack).

## Problem

The public list reads (detector findings/detectors, and the same response shape on sessions/traces) return `PaginationMeta` with a hardcoded `page: 0` plus `limit` and `total`. A caller can see that results were truncated (`total > len(data)`) but has **no way to request the next page** — no offset, no cursor. Fine for the in-app agent's browse-then-drill flow at current volumes, but any API-key client (the CLI foremost) listing findings on a busy project silently loses everything past the first `limit` rows.

## Direction to decide

Results are ordered newest-first with time-window params already in the contract, so two candidate shapes:

1. **Keyset cursor (recommended starting point):** formalize `end_before` as the paging cursor — respond with `meta.next_end_before` (the oldest returned timestamp) and document "pass it back to get the next page". Cheap to implement, stable under concurrent inserts, and already latently supported by the query shape. Needs a tie-break rule for identical timestamps.
2. **Offset paging:** honor a real `page`/`offset` param and echo it in meta. Simpler mental model, but drifts under live inserts and costs deep-scan penalties in ClickHouse.

Whichever is chosen should be applied consistently across the public list reads and stamped into the OpenAPI curation so generated registry surfaces (CLI/MCP/agent) inherit it.

## Acceptance

- A documented way to fetch page 2+ of every public list read, with a truncation signal that tells the caller how to continue.
- Registry entries expose the paging param(s); the parity test keeps the internal twins compatible.
- Existing callers (agent tools) keep working unchanged with paging left unused.

ka1kqi:
## What

Closes the "can't page past the first N results" gap on the detector list reads, for the agent and API-key clients alike: both `list_detectors` and `list_findings` (public + internal surfaces) now accept an opaque `cursor` param and return `meta.next_cursor` while further pages exist.

## How

- **Codec** (`backend/rest/pagination.py`): unpadded base64url of `(timestamp, id)`; decode raises on any malformed input, uniformly surfaced as `422 "Invalid cursor"` before the reader runs.
- **Keyset queries**: total ordering `ORDER BY (timestamp, id) DESC` on both lists; strict tuple/row-comparison predicate so tied timestamps can't skip or duplicate rows across pages (ties are real — detection batches land within one millisecond). Postgres row comparison for the catalog; ClickHouse tuple comparison applied **post-dedup** for findings so a re-ingested finding version can't resurface at a page boundary. Count queries stay cursor-free — `total` remains the filter-wide count, constant across pages.
- **Contract**: `CursorPaginationMeta` subclass (shared `PaginationMeta` untouched — trace/session responses unaffected); `next_cursor` emitted iff the page came back full (an exact-multiple final page yields one empty follow-up — consumers treat an empty page as terminal).
- **Propagation**: OpenAPI artifact + registry regenerated, so the agent's and future CLI/MCP tool schemas inherit `cursor` automatically; agent list formatters append "More results available — call again with cursor=…" and the system prompt teaches the loop.

## Verification

Backend suite green (16 new contract tests across both surfaces: full-page cursor emission of the last row's key, partial-page absence, decoded passthrough, invalid-cursor 422 with the reader never invoked; SQL-shape tests pin predicate placement and count/list scoping). Artifact `--check` clean; registry drift test green; agent/tools suites green; diff coverage vs the stack base 100% on both sides.

## Notes

- Stacked on the get_detector PR (base branch of this PR) — merge that first. Third PR of the detector-reads stack.
- Traces/sessions list reads adopt the same contract in a follow-up; this PR deliberately scopes to the detector surface where the epic lives.

Closes #1919. Part of epic #1877.

---
## Summary by cubic
Adds opaque keyset-cursor pagination to `list_detectors` and `list_findings` on both agent and public API surfaces. Previously pages ended at the first N results; now responses include `meta.next_cursor` and requests accept `cursor`. Invalid cursors return 422. Closes #1919. Part of epic #1877.

- Review notes
  - Codec: unpadded base64url of `(timestamp, id)`; treat the token as opaque. Decode errors are uniformly surfaced as 422 “Invalid cursor” before the reader runs.
  - Queries: total ordering `ORDER BY (timestamp, id) DESC`. Postgres row comparison for detectors, ClickHouse tuple comparison applied post-dedup for findings so re-ingested versions cannot resurface at page boundaries. `total` remains the filter-wide count and ignores `cursor`.
  - Contract: new `CursorPaginationMeta` adds `next_cursor`, emitted only for full pages; an exact-multiple final page yields one empty follow-up.
  - Routers: add `cursor` query param, decode early, and pass the keyset to the reader.
  - OpenAPI/registry: schemas regenerated; tool registry exposes `cursor`. Agent prompt and formatters append a pagination hint only when `next_cursor` is set.
  - Tests: codec round-trip and error cases, router 422 path, predicate placement and ordering, and schema changes are covered.
  - Stacking: based on the get_detector PR; merge that first.

- Rollout
  - Clients must handle `meta.next_cursor` and repeat the same request with `cursor=<value>` to fetch subsequent pages. Treat `cursor` as opaque. Expect a possible empty terminal page when the final page size is an exact multiple of `limit`. No changes required for `total` usage.

<sup>Written for commit 7c73bceb8d14b6c72c908ead473381369e506f2e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1936?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/rest/openapi/public.json
+++ b/backend/rest/openapi/public.json
@@ -117,7 +117,7 @@
         "type": "object"
       },
       "CursorPaginationMeta": {
-        "description": "Pagination meta plus ``next_cursor``, present only while more pages exist.",
+        "description": "Pagination meta plus ``next_cursor``, set when the page came back full; the following page may be empty.",
         "properties": {
           "limit": {
             "title": "Limit",
--- a/backend/rest/openapi/public.json
+++ b/backend/rest/openapi/public.json
@@ -116,6 +116,41 @@
         "title": "CompleteRunResponse",
         "type": "object"
       },
+      "CursorPaginationMeta": {
+        "description": "Pagination meta plus ``next_cursor``, present only while more pages exist.",
+        "properties": {
+          "limit": {
+            "title": "Limit",
+            "type": "integer"
+          },
+          "next_cursor": {
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Next Cursor"
+          },
+          "page": {
+            "title": "Page",
+            "type": "integer"
+          },
+          "total": {
+            "title": "Total",
+            "type": "integer"
+          }
+        },
+        "required": [
+          "page",
+          "limit",
+          "total"
+        ],
+        "title": "CursorPaginationMeta",
+        "type": "object"
+      },
       "DetectorDetail": {
         "description": "A detector's full configuration (Postgres ``detectors`` + optional trigger).\n\n``trigger_conditions`` comes from ``detector_triggers.conditions`` and is\nNone when the detector has no trigger row (it then runs on every sampled\ntrace).",
         "properties": {
@@ -723,7 +758,7 @@
             "type": "array"
           },
           "meta": {
-            "$ref": "#/components/schemas/PaginationMeta"
+            "$ref": "#/components/schemas/CursorPaginationMeta"
           }
         },
         "required": [
@@ -744,7 +779,7 @@
             "type": "array"
           },
           "meta": {
-            "$ref": "#/components/schemas/PaginationMeta"
+            "$ref": "#/components/schemas/CursorPaginationMeta"
           }
         },
         "required": [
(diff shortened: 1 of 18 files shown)


ka1kqi:
Closing unmerged by decision rather than by defect: after review discussion we're not adding a pagination primitive to the typed list endpoints at all. The lists keep the uniform `limit` + `start_after`/`end_before` windowed model (which covers every current consumer — agent, UI, future CLI list commands); exact bulk enumeration is deferred to the future governed query surface, where resumption is the caller's own keyset predicate instead of a per-endpoint API contract.

The branch is kept as a reviewed, green, ready-made stopgap in case an API-key consumer needs exact enumeration before that surface exists: `feat/detector-list-pagination` @ `7c73bceb` (opaque keyset cursor, tie-safe under same-millisecond timestamps, 100% diff-covered). The keyset ordering/tie analysis in it also documents how to enumerate `detector_findings` correctly from any future query surface.

PR #1954, opened 2026-08-19
fix(ui): show the trace-id copy button even without a headerIdentity

### LINKED ISSUE #1947: Copy button missing on offline-evaluation trace ids
## Summary

Every trace surface shows a copy button next to the trace id — except when the header has no identity block, which is **exactly the case for every offline-evaluation trace**. The trace-viewer header only renders the id `CopyButton` when `headerIdentity` is set (`frontend/ui/src/features/traces/components/TraceViewerPanel.tsx:369`); an eval trace opens with no `headerIdentity` (`frontend/ui/src/features/evaluations/views/run-detail-view.tsx:286-288`), so you can read the trace id but have no way to copy it — you have to hand-select a long hash out of a header that offers no affordance for it. Because the copy control is gated on `headerIdentity`, the gap is invisible everywhere else and only shows up on eval traces.

## How to replicate

1. Open an evaluation run and click a case row to open its trace (`run-detail-view.tsx`), or open any trace whose header carries only a `traceId` with no identity block.
2. Observe: no copy button beside the trace id.
3. Open a trace that does have a `headerIdentity` (a normal ingested trace with identity) — the copy button is present. Same header, different affordance.

## Root cause

The `CopyButton` render is conditional on `headerIdentity` rather than on the presence of a `traceId`:
- `TraceViewerPanel.tsx:369` — copy button rendered only inside the `headerIdentity` branch.
- `run-detail-view.tsx:286-288` — eval traces are opened with no `headerIdentity`, so they never reach that branch.

A canonical copy control already exists and is used elsewhere (`components/ui/copy-button.tsx`, e.g. `SpanInfoPanel.tsx:151`); it's simply not wired for the identity-less header.

## Impact

- Eval trace ids cannot be copied at all — the primary way to correlate an eval case with its trace in logs, the SDK, or a URL.
- Inconsistent with every other id surface in the app, which all offer one-click copy.

## Acceptance criteria

- Opening any trace — including an eval trace with no `headerIdentity` — shows a working copy button beside the trace id, using the shared `CopyButton`.
- Traces that already have a `headerIdentity` are unchanged.
- A test asserts the copy control renders for an identity-less trace header.

mohdUwaish59:
Fixes #1947

## Summary

TraceViewerPanel's header already falls back to "Trace" + traceId when no headerIdentity is supplied — which is every offline-eval trace, and in fact every trace in production, since no caller besides offline-eval's run-detail-view ever supplies headerIdentity. The copy button was gated on headerIdentity alone instead of on whether there's actually an id to copy, so those ids had no copy affordance at all.

## Type of Change

- [ ] feat - A new feature
- [x] fix - A bug fix

## Details

Extracted headerLabel/headerValue/headerCopyTitle as a single source of truth for the header: the display and the copy control now read the same values, so they can't diverge — which is the exact bug class this issue was about. Replaced a toLowerCase()-based title template with an explicit branch (avoids "Copy test case id id" for a hypothetical multi-word label).

Tests: added a click-to-copy assertion (navigator.clipboard.writeText called with the exact displayed id) for both the headerIdentity and default-header cases — the previous version only checked the button existed, not that it copied the right value. The default-header test lives under "header actions" (the production contract), not "offline-eval extensions", since it guards the absence of that extension.

## Checklist

---
## Summary by cubic
Fixes #1947 by always showing the `TraceViewerPanel` copy button and copying the displayed trace or header identity. Previously, the button was hidden when no `headerIdentity` was provided. Clipboard failures are now caught, logged, and shown with a failure icon instead of becoming unhandled rejections or false successes.

- The header display and copy control share the same label, value, and tooltip.
- `CopyButton` resets its status after two seconds, cancels stale reset timers, cleans them up on unmount, and ignores results from superseded clicks.
- Tests cover default and custom header identities, successful and rejected writes, failure states, timer behavior, and out-of-order clipboard writes.

<sup>Written for commit 672059a14b4f7fbe3b66ce5c3c131f1dd22e1b59. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1954?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/ui/src/components/ui/copy-button.tsx
+++ b/frontend/ui/src/components/ui/copy-button.tsx
@@ -19,10 +19,17 @@ const CopyButton = React.forwardRef<HTMLButtonElement, CopyButtonProps>(
     const [copied, setCopied] = React.useState(false);
 
     const handleCopy = async () => {
-      await navigator.clipboard.writeText(value);
-      setCopied(true);
-      onCopy?.();
-      setTimeout(() => setCopied(false), 2000);
+      try {
+        await navigator.clipboard.writeText(value);
+        setCopied(true);
+        onCopy?.();
+        setTimeout(() => setCopied(false), 2000);
+      } catch (err) {
+        // A rejected write (denied permission, insecure context, etc.) must
+        // not become an unhandled promise rejection — and copied must stay
+        // false, since nothing was actually copied.
+        console.error("Failed to copy to clipboard:", err);
+      }
     };
 
     return (
--- a/frontend/ui/src/components/ui/copy-button.tsx
+++ b/frontend/ui/src/components/ui/copy-button.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import * as React from "react";
-import { Copy, Check } from "lucide-react";
+import { Copy, Check, X } from "lucide-react";
 import { Button, type ButtonProps } from "./button";
 import { cn } from "@/lib/utils";
 
@@ -11,24 +11,47 @@ export interface CopyButtonProps extends Omit<ButtonProps, "onClick"> {
   iconClassName?: string;
 }
 
+type CopyStatus = "idle" | "copied" | "failed";
+const STATUS_RESET_DELAY_MS = 2000;
+
 /**
- * Copy button that shows a check icon after copying
+ * Copy button that shows a check icon after copying, or a failed icon if the
+ * clipboard write was rejected (denied permission, insecure context, etc.).
  */
 const CopyButton = React.forwardRef<HTMLButtonElement, CopyButtonProps>(
   ({ value, onCopy, className, iconClassName, variant = "ghost", size = "sm", ...props }, ref) => {
-    const [copied, setCopied] = React.useState(false);
+    const [status, setStatus] = React.useState<CopyStatus>("idle");
+    const resetTimeoutRef = React.useRef<ReturnType<typeof setTimeout> | null>(null);
+
+    // Clear any pending reset on unmount so it never fires setState after
+    // the button is gone.
+    React.useEffect(() => {
+      return () => {
+        if (resetTimeoutRef.current) clearTimeout(resetTimeoutRef.current);
+      };
+    }, []);
+
+    const scheduleReset = () => {
+      // A prior click's reset must not fire mid-way through this one — e.g.
+      // copy, fail, copy again inside 2s would otherwise let the first
+      // timeout revert the second click's "copied" state early.
+      if (resetTimeoutRef.current) clearTimeout(resetTimeoutRef.current);
+      resetTimeoutRef.current = setTimeout(() => setStatus("idle"), STATUS_RESET_DELAY_MS);
+    };
 
     const handleCopy = async () => {
       try {
         await navigator.clipboard.writeText(value);
-        setCopied(true);
+        setStatus("copied");
         onCopy?.();
-        setTimeout(() => setCopied(false), 2000);
+        scheduleReset();
       } catch (err) {
-        // A rejected write (denied permission, insecure context, etc.) must
-        // not become an unhandled promise rejection — and copied must stay
-        // false, since nothing was actually copied.
+        // A rejected write must not become an unhandled promise rejection,
+        // and must not be mistaken for success — surface it as its own icon
+        // state instead of silently leaving the idle Copy icon.
         console.error("Failed to copy to clipboard:", err);
+        setStatus("failed");
+        scheduleReset();
       }
     };
 
@@ -41,8 +64,10 @@ const CopyButton = React.forwardRef<HTMLButtonElement, CopyButtonProps>(
         onClick={handleCopy}
         {...props}
       >
-        {copied ? (
+        {status === "copied" ? (
           <Check className={cn("h-3.5 w-3.5 text-green-600", iconClassName)} />
+        ) : status === "failed" ? (
+          <X className={cn("h-3.5 w-3.5 text-destructive", iconClassName)} />
         ) : (
           <Copy className={cn("h-3.5 w-3.5", iconClassName)} />
         )}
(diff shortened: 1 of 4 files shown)


frontend/ui/src/features/traces/components/TraceViewerPanel.interactions.test.tsx
```diff
@@ -243,6 +247,20 @@ describe("TraceViewerPanel header actions", () => {
     fireEvent.click(close!);
     expect(onClose).toHaveBeenCalled();
   });
+
+  it("copies the trace id with no headerIdentity — the default header contract (#1947)", () => {
```
dark-sorceror: Remove the issue number reference.

frontend/ui/src/features/traces/components/TraceViewerPanel.tsx
```diff
@@ -359,20 +365,16 @@ export function TraceViewerPanel({
...
-            )}
+            <span className="shrink-0 text-sm font-medium">{headerLabel}</span>
+            <span className="truncate font-mono text-xs text-muted-foreground">{headerValue}</span>
+            {/* Copy affordance for the header id — always shown (see #1947),
+                reading the same headerLabel/headerValue as the display above
+                so the copied text can never diverge from what's shown. */}
+            <CopyButton
+              value={headerValue}
+              className="h-6 w-6 shrink-0 text-muted-foreground hover:text-foreground"
+              title={headerCopyTitle}
```
dark-sorceror, line 378: Why was the original `title={`Copy ${headerIdentity.label.toLowerCase()} id`}` dropped?

dark-sorceror (changes requested):
Also fixing an existing regression in `frontend/ui/src/components/ui/copy-button.tsx:22`: add a try/catch, so a rejected write becomes a handled rejection and the button changes state properly still.

frontend/ui/src/features/traces/components/TraceViewerPanel.interactions.test.tsx
mohdUwaish59: Okay, I removed the issue number

frontend/ui/src/features/traces/components/TraceViewerPanel.tsx
mohdUwaish59: Yeah right, missed it in my own review. I reverted it to to the original  `Copy ${label.toLowerCase()}` id template.

mohdUwaish59:
> Also fixing an existing regression in `frontend/ui/src/components/ui/copy-button.tsx:22`: add a try/catch, so a rejected write becomes a handled rejection and the button changes state properly still.

done. Wrapped in try/catch; copied stays false on a rejected write since nothing was actually copied, error is logged instead of becoming an unhandled rejection. Also added the component's first  test covering both the success and rejected-write paths. Pushed

frontend/ui/src/features/traces/components/TraceViewerPanel.tsx
```diff
@@ -359,20 +367,16 @@ export function TraceViewerPanel({
...
-            {headerIdentity && (
-              <CopyButton
-                value={headerIdentity.value}
-                className="h-6 w-6 shrink-0 text-muted-foreground hover:text-foreground"
-                title={`Copy ${headerIdentity.label.toLowerCase()} id`}
-              />
-            )}
+            <span className="shrink-0 text-sm font-medium">{headerLabel}</span>
+            <span className="truncate font-mono text-xs text-muted-foreground">{headerValue}</span>
+            {/* Copy affordance for the header id — always shown (see #1947),
```
dark-sorceror: Can we drop the issue number reference here

dark-sorceror (changes requested):
One more nit: add `expect(container.querySelector(".text-green-600")).toBeNull()` to the rejection case.

frontend/ui/src/features/traces/components/TraceViewerPanel.tsx
mohdUwaish59: Done, both the changes and pushed.

frontend/ui/src/components/ui/copy-button.tsx
```diff
@@ -19,10 +19,17 @@ const CopyButton = React.forwardRef<HTMLButtonElement, CopyButtonProps>(
...
     const handleCopy = async () => {
-      await navigator.clipboard.writeText(value);
-      setCopied(true);
-      onCopy?.();
-      setTimeout(() => setCopied(false), 2000);
+      try {
+        await navigator.clipboard.writeText(value);
+        setCopied(true);
+        onCopy?.();
+        setTimeout(() => setCopied(false), 2000);
```
dark-sorceror: Timeout is never cleared, hold it in a `useRef` so it clears on next click with `useEffect` cleanup

frontend/ui/src/components/ui/copy-button.tsx
```diff
@@ -19,10 +19,17 @@ const CopyButton = React.forwardRef<HTMLButtonElement, CopyButtonProps>(
...
+      } catch (err) {
+        // A rejected write (denied permission, insecure context, etc.) must
+        // not become an unhandled promise rejection — and copied must stay
+        // false, since nothing was actually copied.
```
dark-sorceror: Let's add a failed state for the icon here

dark-sorceror (dismissed):

frontend/ui/src/components/ui/copy-button.tsx
mohdUwaish59: done

frontend/ui/src/components/ui/copy-button.tsx
mohdUwaish59: Done. also updated copied from boolean to enum "idle" | "copied" | "failed" such that copied and failed at once is impossible

dark-sorceror:
LGTM

PR #1955, opened 2026-08-19
fix(ui): standardize Save action and disable until changed (#1949)

### LINKED ISSUE #1949: Standardize the Save action and disable it until there are changes
## Summary

The primary "edit and commit" button is done differently on every edit surface in the eval/dataset UI — both its **label** (Save / Save changes / Create / Publish update / Save review / Mark ready) and its **behavior**. The disable-Save-until-something-changes (dirty-state) pattern exists in exactly one place; every other edit surface leaves Save enabled even when nothing has changed, so the same action looks and behaves differently depending on where you are.

## How to replicate

1. Open the **dataset edit** panel — the commit button reads **"Save"** (`dataset-panels.tsx:209`) and is enabled even if you change nothing (`canSave` only checks that the name is non-empty, `:154`).
2. Open the **test-case editor** — the same kind of action reads **"Save changes"** (`test-case-editor-modal.tsx:186`) and is disabled until a field actually changes (`hasChanges`, `:83-88`).
3. Open the **save-result→dataset** drawer (`save-result-to-dataset-drawer.tsx`, button `:288`) and the **review panel** (`review-panel.tsx:281`) — both enable their commit button regardless of edits.

## Root cause

There's no shared convention for the commit action's label or its enabled state:

| Surface | button (file:line) | label | dirty check? |
|---|---|---|---|
| Dataset edit | `dataset-panels.tsx:209` | **Save** | **NO** (`canSave` = name non-empty, `:154`) |
| Test-case editor | `test-case-editor-modal.tsx:186` | **Save changes** | **YES** (`hasChanges`, `:83-88`) |
| Save-result→dataset | `save-result-to-dataset-drawer.tsx:288` | Publish update / Save new case | **NO** (`:178`) |
| Review panel | `review-panel.tsx:281` | Save review | **NO** (`disabled={saving}`, `:280`) |

## Impact

- The same commit action reads "Save" in one place and "Save changes" in another.
- Three edit surfaces let a user "save" without changing anything, producing no-op writes / redundant dataset versions and a false sense that something happened.

## Acceptance criteria

- Every edit-commit button uses one label — plain **"Save"**.
- **Save is disabled by default** — on open, before any edit, it is disabled, and it enables only once a field actually changes (a no-op save is impossible on every edit surface), mirroring `test-case-editor-modal`'s `hasChanges`.

Gaurav-meena95:
### Summary
Standardizes the primary edit-and-commit button across dataset and evaluation edit surfaces (`DatasetEditPanel`, `TestCaseEditorModal`, and `SaveResultToDatasetDrawer`):
1. Every edit-commit button now consistently uses the label "Save" (and "Saving…" while pending).
2. The Save button is disabled by default on initial open and only enables once an editable field has actually changed (`hasChanges`), preventing unintended no-op writes and redundant version publishing.

Closes #1949

### Changes
- `dataset-panels.tsx`: Add `hasChanges` check in `DatasetEditPanel` to disable Save until name or description changes.
- `test-case-editor-modal.tsx`: Standardize button label from `"Save changes"` to `"Save"`.
- `save-result-to-dataset-drawer.tsx`: Add `hasChanges` dirty check and standardize button label to `"Save"`.
- Updated test suites across `dataset-panels.smoke.test.tsx`, `datasets-detail.smoke.test.tsx`, and `save-result-to-dataset-drawer.smoke.test.tsx`.

### Validation
- All 17 test files and 205 tests in `src/features/evaluations` and `src/features/offline-eval` pass (100%).
- ESLint passed with 0 errors.

before edit 
image

after edit 
image

before edit

image

after edit

image

---
## Summary by cubic
Standardizes the primary edit action as a single “Save” button and disables it until something actually changed, closing #1949. Previously buttons varied (“Save changes”, “Publish update”, “Save new case”) and allowed no-op writes; now they show “Save”/“Saving…” and only enable when the persisted value would differ.

- Replaces labels with “Save” (and “Saving…”) across `DatasetEditPanel`, `TestCaseEditorModal`, and `SaveResultToDatasetDrawer`, and drops the unused `ACTION_META` CTA labels.
- Dirty checks compare each field as persisted: input via `encodeEditedText` (whitespace-only edits count since the case id hashes exact bytes; reformatting structured input doesn’t), expected trimmed, and metadata canonically so reformatting isn’t a change. Metadata with no canonical form (an unpaired UTF-16 surrogate) blocks Save with an error; create mode requires non-empty input, and seeded metadata that can’t canonicalize doesn’t lock out other field edits.
- Adds a leaf `canonical.ts` for client-side canonicalization; `case-id.ts` and `versions.ts` re-export from it.
- In the result drawer, only “Update source case” is gated on changes (dataset selection, input, expected, or toggling “use candidate as expected”); save-new-case and duplicate are always savable.
- Updates smoke tests to assert the “Save” label, disabled-by-default behavior, and the per-field dirty checks.

<sup>Written for commit 525e59e4118a8d05f8ed83db550ebc336d27d935. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1955?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/ui/src/features/evaluations/components/save-result-to-dataset-drawer.tsx
+++ b/frontend/ui/src/features/evaluations/components/save-result-to-dataset-drawer.tsx
@@ -177,8 +177,19 @@ export function SaveResultToDatasetDrawer({
     },
   });
 
+  const hasChanges =
+    action === "update_existing_case"
+      ? (meta.picksDataset && targetDatasetId !== sourceDatasetId) ||
+        input.trim() !== result.input.trim() ||
+        (useCandidateAsExpected
+          ? (result.candidateOutput ?? "").trim() !== (result.expectedOutput ?? "").trim()
+          : expected.trim() !== (result.expectedOutput ?? "").trim())
+      : true;
+
   const canSave =
-    !save.isPending && (!meta.picksDataset || datasets.some((d) => d.id === targetDatasetId));
+    !save.isPending &&
+    (!meta.picksDataset || datasets.some((d) => d.id === targetDatasetId)) &&
+    hasChanges;
 
   return (
     <Drawer open={open} onOpenChange={onOpenChange}>
@@ -302,7 +313,7 @@ export function SaveResultToDatasetDrawer({
             Cancel
           </Button>
           <Button size="sm" disabled={!canSave} onClick={() => save.mutate()}>
-            {save.isPending ? "Saving…" : meta.cta}
+            {save.isPending ? "Saving…" : "Save"}
           </Button>
         </DrawerFooter>
       </DrawerContent>
--- a/frontend/ui/src/features/evaluations/components/save-result-to-dataset-drawer.tsx
+++ b/frontend/ui/src/features/evaluations/components/save-result-to-dataset-drawer.tsx
@@ -24,6 +24,7 @@ import {
 } from "@/components/ui/drawer";
 import { useToast } from "@/components/ui/toast";
 import { EditableValueBlock } from "@/features/offline-eval/components";
+import { encodeEditedText } from "@/lib/eval/json-value";
 import { useDatasets } from "../hooks";
 
 /** Read-only fields never edit, so onChange is a no-op. */
@@ -186,7 +187,7 @@ export function SaveResultToDatasetDrawer({
   const hasChanges =
     action === "update_existing_case"
       ? (meta.picksDataset && targetDatasetId !== sourceDatasetId) ||
-        input !== result.input ||
+        encodeEditedText(result.input, input) !== encodeEditedText(result.input, result.input) ||
         (useCandidateAsExpected
           ? (result.candidateOutput ?? "") !== (result.expectedOutput ?? "")
           : expected !== (result.expectedOutput ?? ""))
--- a/frontend/ui/src/features/evaluations/components/save-result-to-dataset-drawer.tsx
+++ b/frontend/ui/src/features/evaluations/components/save-result-to-dataset-drawer.tsx
@@ -177,13 +177,19 @@ export function SaveResultToDatasetDrawer({
     },
   });
 
+  // Only "update" republishes an existing case, so only it needs a dirty gate; save/duplicate
+  // always create a case and are savable as they stand. Every field is compared the way it is
+  // PERSISTED — this drawer POSTs `input` and `expected` verbatim, and a case's id is a hash of
+  // its exact input (`stableCaseId` over `canonicalJson`, which does not trim) — so nothing here
+  // trims. A whitespace-only edit really does change the published case; gating it away as a
+  // no-op would leave the reviewer unable to save a change the backend would have accepted.
   const hasChanges =
     action === "update_existing_case"
       ? (meta.picksDataset && targetDatasetId !== sourceDatasetId) ||
-        input.trim() !== result.input.trim() ||
+        input !== result.input ||
         (useCandidateAsExpected
-          ? (result.candidateOutput ?? "").trim() !== (result.expectedOutput ?? "").trim()
-          : expected.trim() !== (result.expectedOutput ?? "").trim())
+          ? (result.candidateOutput ?? "") !== (result.expectedOutput ?? "")
+          : expected !== (result.expectedOutput ?? ""))
       : true;
 
   const canSave =
--- a/frontend/ui/src/features/evaluations/components/save-result-to-dataset-drawer.tsx
+++ b/frontend/ui/src/features/evaluations/components/save-result-to-dataset-drawer.tsx
@@ -42,27 +42,24 @@ export interface ResultForDataset {
 
 const ACTION_META: Record<
   ResultDatasetAction,
-  { title: string; blurb: string; cta: string; picksDataset: boolean }
+  { title: string; blurb: string; picksDataset: boolean }
 > = {
   update_existing_case: {
     title: "Update source case",
     blurb:
       "Publishes a new immutable version of this run's dataset with this case updated. Runs already recorded keep pointing at the version they used.",
-    cta: "Publish update",
     picksDataset: false,
   },
   save_new_case: {
     title: "Save as a new case",
     blurb:
       "Adds a brand-new case to the chosen dataset. Blocked if it would just recreate this result's own source case — use “Update source case” for that.",
-    cta: "Save new case",
     picksDataset: true,
   },
   duplicate_as_variant: {
     title: "Duplicate as a variant",
     blurb:
       "Adds a new case tagged as a variant of the source case. Always allowed — useful for exploring a tweak without touching the original.",
-    cta: "Duplicate case",
     picksDataset: true,
   },
 };
--- a/frontend/ui/src/features/evaluations/components/save-result-to-dataset-drawer.tsx
+++ b/frontend/ui/src/features/evaluations/components/save-result-to-dataset-drawer.tsx
@@ -180,6 +180,9 @@ export function SaveResultToDatasetDrawer({
   // its exact input (`stableCaseId` over `canonicalJson`, which does not trim) — so nothing here
   // trims. A whitespace-only edit really does change the published case; gating it away as a
   // no-op would leave the reviewer unable to save a change the backend would have accepted.
+  // Trimming `input` on the way in is not an option either: that id derivation is byte-parity
+  // with the TS/Python SDKs, so a trimming UI would give `" hi "` a different `tc_` id than the
+  // same input pushed from an SDK, and a re-publish would duplicate the case, not upsert it.
   const hasChanges =
     action === "update_existing_case"
       ? (meta.picksDataset && targetDatasetId !== sourceDatasetId) ||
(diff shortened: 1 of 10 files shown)


ka1kqi:
@Gaurav-meena95 please resolve merge conflicts here

ka1kqi:
@dark-sorceror

frontend/ui/src/features/evaluations/components/save-result-to-dataset-drawer.tsx
```diff
@@ -172,7 +172,13 @@ export function SaveResultToDatasetDrawer({
...
   });
 
-  const canSave = !save.isPending && (!meta.picksDataset || !!targetDatasetId);
+  const hasChanges =
+    (meta.picksDataset && targetDatasetId !== sourceDatasetId) ||
+    input !== result.input ||
+    useCandidateAsExpected ||
+    (!useCandidateAsExpected && expected !== (result.expectedOutput ?? ""));
+
+  const canSave = !save.isPending && (!meta.picksDataset || !!targetDatasetId) && hasChanges;
```
dark-sorceror: Drawer has three actions and only one is an edit-and-commit. On "update the source case" is an edit and commit, but "save new" and "duplicate" create a case, so they are savable as is.

dark-sorceror: Keep the membership check; do not restore `!!targetDatasetId`.

dark-sorceror (dismissed):

frontend/ui/src/features/evaluations/components/save-result-to-dataset-drawer.tsx
Gaurav-meena95: Resolved the merge conflicts and updated hasChanges so it only applies when updating an existing case, while keeping the dataset membership check from main. Ready for re-review!

frontend/ui/src/features/evaluations/components/save-result-to-dataset-drawer.tsx
```diff
@@ -177,8 +177,18 @@ export function SaveResultToDatasetDrawer({
     },
   });
 
+  const hasChanges =
+    action === "update_existing_case"
+      ? (meta.picksDataset && targetDatasetId !== sourceDatasetId) ||
+        input !== result.input ||
```
dark-sorceror, line 189: Add `.trim()`

frontend/ui/src/features/evaluations/components/save-result-to-dataset-drawer.tsx
```diff
@@ -177,8 +177,18 @@ export function SaveResultToDatasetDrawer({
...
+        useCandidateAsExpected ||
+        (!useCandidateAsExpected && expected !== (result.expectedOutput ?? ""))
```
dark-sorceror: ditto

frontend/ui/src/features/evaluations/components/test-case-editor-modal.tsx
dark-sorceror: ditto

frontend/ui/src/features/evaluations/save-result-to-dataset-drawer.smoke.test.tsx
```diff
@@ -176,10 +193,12 @@ describe("SaveResultToDatasetDrawer", () => {
       body: { error: "This result already came from this case — use update instead." },
     });
     mount("save_new_case");
+    // Change input so Save enables while still targeting ds1 (originating dataset)
+    fireEvent.change(screen.getByLabelText("Input"), { target: { value: "Different question" } });
```
dark-sorceror: Do we still need these two lines since now that create actions are exempt?

frontend/ui/src/features/evaluations/components/save-result-to-dataset-drawer.tsx
```diff
@@ -177,8 +177,18 @@ export function SaveResultToDatasetDrawer({
     },
   });
 
+  const hasChanges =
+    action === "update_existing_case"
+      ? (meta.picksDataset && targetDatasetId !== sourceDatasetId) ||
+        input !== result.input ||
+        useCandidateAsExpected ||
```
dark-sorceror: agree with cubic

dark-sorceror (dismissed):
Also resolve the cubic reviews once you fix them

dark-sorceror:
Also feel free to resolve review comments in general if they are resolved and replied

Gaurav-meena95:
Hi @dark-sorceror @ka1kqi,

I have resolved all merge conflicts by rebasing against latest `main`, and addressed all review feedback:
- Added `.trim()` across `test-case-editor-modal.tsx` and `save-result-to-dataset-drawer.tsx`.
- Handled candidate output change detection to only mark dirty when it produces an effective change over expected output.
- Reverted the unrelated `docker-compose.yml` change.
- Cleaned up the redundant smoke test lines.

All 12 test suites (100 tests) in `src/features/evaluations` pass cleanly. Ready for review!

dark-sorceror:
Taking a look soon!

frontend/ui/src/features/evaluations/components/test-case-editor-modal.tsx
```diff
@@ -82,9 +82,9 @@ export function TestCaseEditorModal({
   // In edit mode, keep Save disabled until a field actually changes (a create is always savable).
   const hasChanges =
     mode.kind !== "edit" ||
-    input !== mode.input ||
-    expected !== (mode.expected ?? "") ||
-    metadata !== metadataToText(mode.metadata);
+    input.trim() !== mode.input.trim() ||
```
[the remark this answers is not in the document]
dark-sorceror: Resolved in 3f8e652

Valid, and the fix is the second option you listed — dropping `.trim()` from the input comparison rather than trimming on save. A case's id is `stableCaseId` over `canonicalJson(input)`, which is `JSON.stringify` with no trimming and must stay byte-identical to the TS and Python SDKs, so the UI must not normalize input on the way out; that also makes a whitespace-only input edit a genuinely different case, which Save has to allow rather than gate away as a no-op. `expected` and `metadata` keep their trimmed comparisons because they are persisted trimmed and parsed respectively, and the same untrimmed treatment was applied to the equivalent check in `save-result-to-dataset-drawer.tsx`, with a regression test on each surface.

dark-sorceror:
LGTM

Gaurav-meena95:
Thanks for taking a look @dark-sorceror 

Since branch protection requires an official approving review, could you please submit a formal review choosing "Approve" via the "Files changed" -> "Review changes" button? The plain text "LGTM" comment doesn't unlock the merge button.

XinweiHe:
Thanks for this — CI is green and the direction is right. A few things to fix before it can go in.

**Blocker: a stored row can become permanently uneditable.**
`metadataError` runs `canonicalJson(parsed)` over metadata seeded from storage (`test-case-editor-modal.tsx:75`). Metadata written by an SDK with an unpaired surrogate is valid JSON, and the PATCH route stores it as JSONB without re-canonicalizing — so `canSave` stays false forever, and Input/Expected edits are blocked too. Validate only what the user typed, not the seed. The assertion at `datasets-detail.smoke.test.tsx:404` currently locks in the dead-Save state.

**Should fix**
1. Dirty checks trim *both* sides (`dataset-panels.tsx:155`, `test-case-editor-modal.tsx:131`), so normalizing whitespace in a stored name/description/expected can never be saved — even though the PATCH would persist the trimmed value. Trim the editor side only.
2. Create mode is unconditionally dirty: `+ Row` → Save with every field blank publishes a version with an empty input (`CreateTestCaseRequestSchema.input` has no `.min(1)`). Given the PR's goal, the create path is the one that most needs a gate.
3. `input` is compared raw but persisted through `encodeEditedText`, so a pure reformat of a structured input enables Save and publishes a byte-identical version — the redundant publish #1949 asks to prevent.
4. The `catch` around `canonicalJson` is unqualified, so any throw (e.g. deep recursion on pasted metadata) is reported as "invalid Unicode (an unpaired surrogate)". Narrow it to `LoneSurrogateError` and rethrow the rest.

**Nits**
- `metrics.ts:42` still carries a private `canonicalJson` that diverges from the newly extracted one (default `.sort()`, no surrogate guard) — worth deleting in favour of the shared module.
- `canonical.ts` now ships a regex lookbehind into a `"use client"` bundle. Fine for current browsers, just flagging it.

Heads-up: this touches `dataset-panels.smoke.test.tsx`, which #1957 also edits.

Gaurav-meena95:
Thanks for the detailed review @XinweiHe  I've addressed all the points:

- **Blocker (uneditable stored row):** Metadata is now only validated if modified by the user (`metadata !== seedMetadataText`). Stored rows with unpaired surrogates remain editable for `input`/`expected`.
- **Trim editor side only:** Updated dirty checks for dataset name, description, and test case expected to trim only the editor value so whitespace normalizations can be saved.
- **Create mode gate:** Gated `+ Row` to require `input.trim().length > 0` before enabling Save.
- **Input comparison parity:** Compared `input` through `encodeEditedText` to prevent redundant publishes when reformatting structured JSON.
- **Narrowed catch:** Narrowed the `canonicalJson` error handling to `LoneSurrogateError` and rethrown any other unexpected errors.
- **Metrics cleanup:** Replaced the private `canonicalJson` in `metrics.ts` with the shared module.

All checks and test suites (`npm run test`, `format:check`, `lint`) are passing locally.

Issue #1795, opened 2026-08-02
Add a parity check so provider/framework lists can't drift from the sources of truth


dark-sorceror:
## Summary

The model-provider and agent-framework lists are hand-maintained in the docs and the in-app getting-started picker, and they're supposed to track what the SDK and BYOK config actually support — but nothing enforces it. That's how the current inconsistencies accumulated (Groq/Bedrock/Azure hidden, xAI/Moonshot/Z.AI/DeepSeek missing from docs/app, provider counts of 5 vs 10 vs 6). Add a test (or codegen) that verifies each display list against its source of truth.

## The lists and their sources of truth

- **Instrumentation integrations** — SDK `Integration` enum (`traceroot-py/traceroot/instrumentation/registry.py:19-38`) + TS `instrumentModules` (`traceroot-ts/packages/traceroot/src/types.ts:51-66`).
- **BYOK model providers** — `ADAPTER_CONFIG` (`traceroot/frontend/packages/core/src/llm-providers.ts:239-297`).

## The display surfaces that must match

- Docs: `traceroot/docs/integrations/overview.mdx` cards + `docs.json` nav + the `integrations/*.mdx` pages.
- App getting-started picker: `traceroot/frontend/ui/src/features/traces/components/GettingStarted/integrations.tsx` (the file comment already says "Keep this in sync with docs/integrations/overview.mdx" — but nothing enforces it).

## Impact

Without an enforced check, the lists diverge every time a provider/framework is added or removed — exactly what happened here. A parity test turns "please remember to update both" into a build failure.

## Acceptance criteria

- A test asserts each display list is consistent with its source of truth: every displayed instrumentation integration exists in the SDK enum (and has a docs page); every displayed BYOK provider exists in `ADAPTER_CONFIG`; and nothing supported-and-intended-for-display is missing.
- The docs `overview.mdx` cards, `docs.json` nav, and the `integrations/*.mdx` pages stay in 1:1 correspondence (already true today — lock it with a test).
- Preferably, the app picker and/or docs list are generated from a shared manifest rather than hand-copied, so "keep in sync" is structural, not a comment.
- The check fails loudly on drift (a provider added to `ADAPTER_CONFIG` or the SDK enum but not surfaced, or displayed but unsupported).

Satyam296:
@dark-sorceror Could you please assign me this issue ?

dark-sorceror:
@Satyam296 assigned

Issue #1630, opened 2026-07-27
[Epic] Icon consolidation — finish the domain-icon registry rollout


ka1kqi:
Domain concepts (tokens, cost, model, user, session, trace, …) were each drawn with a lucide icon picked ad-hoc per file, so one concept drifted to several glyphs across the app. The registry work introduced `frontend/ui/src/components/icons/domain-icons.ts` as the single source of truth, migrated ~25 surfaces onto it, and added a `no-restricted-imports` guard so the registry cannot be bypassed.

These are the pieces deliberately left out of that change, either because they needed their own design decision or because they were cross-cutting refactors that did not belong in a feature PR.

### Scope

- [ ] #1627 — success / failure / pending state entries, plus the deprecated-alias hole they leave in the lint guard
- [ ] #1628 — a distinct fallback glyph, so unmapped fields stop reading as "model"
- [ ] #1629 — test debt: duplicated mock scaffolding, and icon changes that are covered but not asserted

### Ordering

#1627 is the highest value: it closes a live case of one component being drawn under two spellings, and it is the only item that removes a documented carve-out from the lint guard. #1628 is a small self-contained fix. #1629 is pure hygiene and can go last.

### Not in scope

Some lucide glyphs stay deliberately unrestricted because each has a genuine second meaning in this app: `Eye` (password visibility), `Hash` (Slack channel), `Clock` (durations), `ArrowRight` (call-to-action arrows), `Sparkle` ("magic"), `CircleAlert` (generic warnings). Those are correct as-is and should not be migrated.

shankywho:
hey @ka1kqi, can you assign me this issue

Issue #1629, opened 2026-07-27
test(ui): de-duplicate page-test mock scaffolding and strengthen icon assertions


ka1kqi:
Part of #1630.

Three related bits of test debt surfaced while reviewing the domain-icon registry work. None is urgent; grouping them so they are fixed together.

### Duplicated retention mock block

The same six-field `useRetention` + `PricingDialog` mock is copy-pasted across five files:

- `frontend/ui/src/app/projects/[projectId]/detectors/page.test.tsx`
- `frontend/ui/src/app/projects/[projectId]/detectors/[detectorId]/page.test.tsx`
- `frontend/ui/src/app/projects/[projectId]/sessions/page.test.tsx`
- `frontend/ui/src/app/projects/[projectId]/users/page.test.tsx`
- `frontend/ui/src/app/projects/[projectId]/traces/page.user-filter.test.tsx`

A shared test helper would stop the next page test rediscovering which fields the hook returns. Deliberately not done during the registry work to avoid diverging from the established pattern mid-change.

### Duplicated page-test scaffolding

`traces/page.user-filter.test.tsx` duplicates roughly 60 lines of mock scaffolding from its sibling `traces/page.prefetch.test.tsx`, which already uses mutable module-level state for exactly this "vary one input across cases" need. Making `useSearchParams` read a mutable variable there would let the badge case move in and the extra file be deleted.

Its `useRetention` / `PricingDialog` mocks are also redundant, since its `QueryClientProvider` wrapper lets the real hook run.

### Icon changes that are covered but not asserted

Coverage counts these lines, but the tests would still pass with the icon deleted:

- `frontend/ui/src/features/traces/components/SessionDetailPanel.test.tsx` asserts badge **text** only, across four migrated icon lines
- The onboarding project-step glyph change (`Layers` → `FolderKanban`) has no assertion at all

`SpanTreeView.render.test.tsx` and `SpanTimelineView.test.tsx` show the pattern to copy: assert `badge.querySelector("svg")` is present.

Gaurav-meena95:
Hi @ka1kqi, I'd be happy to work on this test cleanup as well. I've recently worked on the related dashboard/icon changes, so I already have the context. Could you please assign this issue to me? Thanks!

Gaurav-meena95:
Hi @ka1kqi  i have raised the pr test(ui): de-duplicate page-test mock scaffolding and strengthen icon…- #1633
#1633

Issue #1555, opened 2026-07-14
Docs integration cards' logos without a dark variant don't match in dark mode


dark-sorceror:
## Summary

On the docs Integrations overview, each integration card shows a light-mode logo and only swaps to a dark variant when a `logoDark` is provided. Many cards have no `logoDark`, so their light logos (often dark-colored marks meant for a white background) render unchanged on the dark card — low-contrast or barely visible — while the cards that do provide a dark variant look crisp. The result is an inconsistent, mismatched grid in dark mode.

## How to replicate

1. Open the docs Integrations overview page in **dark mode**.
2. Cards with a dark logo (e.g. Python, TypeScript, OpenAI, Anthropic) render correctly.
3. Cards without one (e.g. Agno, CrewAI, DSPy, LangChain, LlamaIndex, Gemini, Mistral) show their light-mode logo on the dark background — washed out / hard to read and visually inconsistent with the rest.

<img width="651" height="734" alt="Image" src="https://github.com/user-attachments/assets/0e3e1c3d-cd1c-4984-8909-ba35c974f586" />

## Impact

Cosmetic, but it's the docs' first impression of the integration surface: a mix of crisp and washed-out logos reads as unpolished, and some marks are nearly illegible in dark mode.

## Acceptance criteria

- Every integration card's logo is legible and visually consistent on the dark card background.
- Logos that need a dark variant get a `logoDark` (dark/inverted asset); logos that already read well on both grounds are left as-is.
- New integration cards have a clear convention for when a `logoDark` is required.

Rajesh-Nagarajan-11:
Please assign me this issue @dark-sorceror

dark-sorceror:
@Rajesh-Nagarajan-11 assigned

kashifalikhan36:
@Rajesh-Nagarajan-11 are you working on this issue, it's more then 2 days bro ?

Rajesh-Nagarajan-11:
Yes I'm working on this

PR #1970, opened 2026-08-19
fix(ui): provider brand marks legible in dark mode + de-duplicate SVG boilerplate

### LINKED ISSUE #1967: xAI provider icon is invisible in dark mode — hard-coded black fill, and nothing stops the next one
## Summary

Provider brand icons in the model-providers (BYOK) settings tab don't all render legibly in dark mode. `XAIIcon` hard-codes `fill="#000"` on a monochrome mark with no background tile, so the xAI/Grok icon is **effectively invisible on dark backgrounds**, while its theme-aware siblings (OpenAI, Anthropic, OpenRouter) use `fill="currentColor"` and adapt correctly.

Two structural problems in the same file make this class of bug easy to reintroduce. All 10 icon components repeat the same `<svg>` boilerplate, so any cross-cutting fix is a 10-way edit. And `PROVIDER_ICONS` is typed `Record<string, FC<IconProps>>` even though a canonical adapter list — `LLMAdapter` — already exists in `@core/llm-providers`, so the registry can silently fall out of sync with the adapters the product actually supports.

## How to replicate

1. Switch the app to **dark mode**.
2. Go to **Settings → Workspace → Model Providers**.
3. Add or view an **xAI (Grok)** BYOK provider (adapter `xai`).
4. The xAI icon is invisible / near-invisible (black-on-dark) while the other provider icons in the same list render fine.

## Root cause (confirmed)

1. **`XAIIcon` hard-codes a black fill.** `frontend/ui/src/components/icons/provider-icons.tsx:188`
   ```tsx
   <path d="M6.469 8.776L16.512 23h-4.464L2.005 8.776H6.47z…" fill="#000" />
   ```
   A monochrome brand mark with no background tile and a literal black fill has no dark-mode adaptation. Its theme-aware siblings instead use `fill="currentColor"` and inherit a legible text color — `OpenAIIcon:11`, `AnthropicIcon:24`, `OpenRouterIcon:164`.

2. **`MoonshotIcon` (`:194`) and `ZAIIcon` (`:209`)** each paint their own dark `rect` tile (`:201`), so they stay legible in both themes but sit as dark-on-dark blocks against a dark page — low contrast rather than invisible. Worth an explicit keep-or-change decision in the same pass.

3. **10× duplicated `<svg>` boilerplate.** Every component repeats `<svg viewBox="0 0 24 24" className={cn("h-4 w-4", className)} xmlns aria-hidden="true">`. There is no single place to apply a theme-related default, which is why the `fill="#000"` outlier survived — it reads as "just another icon" rather than as a deviation.

4. **The icon registry is not keyed off the canonical adapter list.** A single source of truth already exists:
   - `LLMAdapter` const + type — `frontend/packages/core/src/llm-providers.ts:9`
   - but `PROVIDER_ICONS` is `Record<string, FC<IconProps>>` — `provider-icons.tsx:225`
   - and the dispatcher looks up `adapter.toLowerCase()` against it — `provider-icons.tsx:244`

   Because the map's key type is `string`, adding a value to `LLMAdapter` (and to the `adapter` value list documented at `frontend/packages/core/prisma/schema.prisma:131`, which is a plain `String` column, not a DB enum) compiles cleanly with no icon. The new provider silently renders the gray letter-circle fallback (`:259`).

### Why this isn't caught

- The icons are decorative (`aria-hidden="true"`) and always paired with a text label, so an invisible icon breaks nothing functionally — it just looks broken, and only in dark mode, so it passes light-mode review.
- No theme or visual-regression coverage exists over the icon set, and no test asserts that every `LLMAdapter` value resolves to an icon.
- The registry's `string` key type means TypeScript cannot flag the gap either.

## Impact

- **Broken-looking dark-mode UI:** the xAI provider appears to have no icon — a blank slot — in the model-providers tab on dark themes. This is on a settings page users hit while entering API keys, so it reads as an unfinished product.
- **Latent drift:** adding an adapter without an icon degrades silently to the letter-circle, with neither the compiler nor a test flagging it.
- **Maintenance drag:** 10× duplicated `<svg>` boilerplate makes every icon-wide edit repetitive and easy to get subtly wrong — the current bug is an instance of exactly that.

## Notes

The four marks currently commented `(from lmnr reference)` — `AnthropicIcon:24`, `AzureIcon:38`, `GoogleIcon:103`, `AWSIcon:126` — carry path data byte-identical to Laminar's `frontend/components/ui/icons.tsx`, and Laminar is Apache-2.0. The remaining six are commented `(from lobehub/lobe-icons)` (MIT). Since this work rewrites every icon component anyway, re-sourcing the four from lobehub/lobe-icons makes the set single-sourced under one license and lets the attribution collapse into a single accurate file header. That is a licensing/consistency cleanup that belongs with the refactor, not a separate change — and if a given mark can't be cleanly re-sourced, its credit stays.

## Acceptance criteria

- Every provider icon renders legibly in **both light and dark mode**; monochrome marks with no background tile inherit `currentColor`. Verified against the model-providers tab in dark mode, with the xAI/Grok mark visible.
- An explicit, recorded decision on the `Moonshot` / `Z.AI` dark tiles — kept as-is or made theme-aware, not left unexamined.
- The repeated `<svg>` boilerplate is extracted into one shared wrapper; each icon component declares only its own paths/gradients.
- `PROVIDER_ICONS` is keyed by the `LLMAdapter` type so a missing icon is a compile error, plus a test asserting every `LLMAdapter` value resolves to an icon (or to an intentional fallback).
- The icon set is single-sourced, with one file-header attribution that accurately describes where the marks come from.
- Light-mode rendering is unchanged for every icon not re-sourced; the only intended visual change is dark-mode legibility.

dark-sorceror:
## Summary

Fixes: traceroot-ai/traceroot#1967
Supersedes: #1915

`XAIIcon` painted a monochrome mark with `fill="#000"` and no background tile, so the xAI/Grok icon **disappeared on dark backgrounds** in the model-providers (BYOK) settings tab — the provider row showed a blank slot where every other provider showed an icon. Upstream's mono mark uses `currentColor`; the hard-coded black was introduced when the path was copied in.

Fixing the one fill would have left the shape of the bug in place, so this also removes the two things that let it through: ten icon components each repeating the same `<svg>` opener with nowhere to put a shared default, and a `PROVIDER_ICONS` map typed `Record<string, FC>` that couldn't tell the compiler when an adapter had no icon.

## What changed

`frontend/ui/src/components/icons/provider-icons.tsx`:

- **Dark-mode fix.** `XAIIcon` uses `fill="currentColor"`, matching `OpenAIIcon` / `AnthropicIcon` / `OpenRouterIcon`. xAI/Grok is monochrome by brand, so inheriting text colour is the upstream treatment, not a brand deviation.
- **Shared `ProviderSvg` wrapper.** Owns `viewBox`, `className={cn("h-4 w-4", className)}`, `xmlns` and `aria-hidden`. Each icon now declares only its own paths, gradients and fills.
- **Registry keyed by `LLMAdapter`.** `PROVIDER_ICONS` goes from `Record<string, FC<IconProps>>` to `Record<LLMAdapter, FC<IconProps>>`. `LLMAdapter` already existed in `@traceroot/core` as the canonical adapter list — the registry just wasn't using it. Imported as `import type`, so nothing extra reaches the client bundle. The dispatcher still widens for the actual lookup, since `adapter` is a free-form string at runtime and the letter-circle fallback has to stay reachable.
- **`Moonshot` / `Z.AI` tiles kept.** Both are white-on-dark by brand and carry their own `rect`, so they are legible either way; changing them would alter light mode too. Deliberate keep.

New: `frontend/ui/src/components/icons/provider-icons.test.tsx` — 14 tests covering adapter coverage, the no-pure-black rule, and the fallback path.

## Attribution — the comments were wrong

Four marks were commented `(from lmnr reference)`. Checked against `@lobehub/icons@5.16.0` (MIT), path by path:

| Mark | Result |
|---|---|
| Anthropic | byte-identical to `Anthropic/Mono` |
| Azure | byte-identical to `Azure/Color` + `Azure/Mono` |
| AWS Bedrock | byte-identical to `Bedrock/Mono` (including gradient stops) |
| Google Gemini | **no match** — a simplified four-point star |

So three of the four were lobehub's MIT icons all along, and the `(from lmnr)` note simply named the wrong source. Gemini was the only genuinely divergent mark; it is re-sourced to upstream's current four-colour spark, which also brings it in line with Google's present brand mark. Three more marks (DeepSeek, xAI, Moonshot) were verified identical to their lobehub equivalents; OpenAI, OpenRouter and Z.AI do not match v5.16.0 and are left untouched.

The ten inline `(from …)` notes are replaced by one file header crediting `@lobehub/icons` (MIT, © LobeHub) and noting that each mark is its provider's trademark.

**Visual impact:** Gemini is the only mark whose rendering changes. Everything else is byte-identical to `main` apart from the xAI fill.

## Notes for reviewers

- **Light mode is unchanged except for Gemini.** The only other intended visual change is xAI in dark mode.
- **`schema.prisma:131` is untouched.** `ModelProvider.adapter` is a plain `String` with its valid values in a trailing comment, not a DB enum. Converting it is a migration and out of scope; `LLMAdapter` is the enforceable source of truth on the TypeScript side, which is where the bug lives.
- **`ModelProvidersTab.tsx` is untouched.** Both `ProviderIcon` call sites (`:385`, `:436`) keep the same props.
- **Related, deliberately not folded in:** #1555 is the same class of bug (a brand mark assuming a light ground) on the docs integration cards — different surface, different mechanism (`logoDark` asset vs `currentColor`), no shared code. #1795 proposes a parity check for provider *display lists* against `ADAPTER_CONFIG`; the registry typing here covers the *icon* list against `LLMAdapter`. If #1795 later generalises into one drift-check suite, this test should move into it.
- The new test follows the existing `// @vitest-environment jsdom` + Testing Library pattern. If #1629 (under epic #1630) settles on a different icon-assertion convention, this should adopt it. #1630 itself covers `domain-icons.ts`, not this file.

## Test plan

- [x] `vitest run src/components/icons/` — 14/14 pass
- [x] Full `frontend/ui` suite — 208 files / 1970 tests pass
- [x] `tsc --noEmit` output byte-identical to `origin/main` (6 pre-existing errors in unrelated test files, none introduced)
- [x] Drift guard verified: deleting `moonshot` from the registry fails with `TS2741: Property 'moonshot' is missing in type … but required in type 'Record<LLMAdapter, FC<IconProps>>'`
- [x] `eslint src/components/icons/` clean; `prettier --check` clean
- [x] Provenance verified programmatically against `@lobehub/icons@5.16.0`, not by eye
- [ ] Visual check in a browser: dark mode shows the xAI mark; light mode unchanged apart from Gemini (screenshots pending)

---
## Summary by cubic
Fixes dark‑mode invisibility of the xAI/Grok provider icon and removes duplicated SVG setup across provider icons. Previously xAI used a hard‑coded black fill and disappeared on dark backgrounds; it now inherits `currentColor`. Gemini now uses the current four‑color spark; all other marks are visually unchanged.

- Shared `ProviderSvg` wrapper centralizes sizing and common SVG attributes; each icon now declares only its own paths/gradients/fills.
- `PROVIDER_ICONS` is keyed by `LLMAdapter` from `@traceroot/core` to prevent missing icons; runtime lookup stays case‑insensitive with the letter‑circle fallback for unknown adapters.
- Adds tests guarding adapter coverage, banning pure‑black fills unless the icon carries its own tile, and verifying fallback behavior.
- Moonshot and `Z.AI` keep their dark tiles by brand; light mode is unchanged except for Gemini.
- Icon attribution corrected to `@lobehub/icons`; per‑icon notes replaced by a single file header.

<sup>Written for commit debcbea66e11bd259478e65e75760043e3a22cfe. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1970?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/ui/src/components/icons/provider-icons.test.tsx
+++ b/frontend/ui/src/components/icons/provider-icons.test.tsx
@@ -1,0 +1,66 @@
+// @vitest-environment jsdom
+import { afterEach, describe, expect, it } from "vitest";
+import { cleanup, render } from "@testing-library/react";
+import { LLMAdapter } from "@traceroot/core/llm-providers";
+import { ProviderIcon } from "./provider-icons";
+
+afterEach(cleanup);
+
+const ADAPTERS = Object.values(LLMAdapter);
+
+/** The letter-circle fallback is the only mark rendered with a <text> node. */
+function isFallback(container: HTMLElement): boolean {
+  return container.querySelector("text") !== null;
+}
+
+describe("ProviderIcon", () => {
+  it("covers every supported adapter with a real brand mark", () => {
+    // Guards the registry against drift: a new LLMAdapter with no icon would
+    // otherwise degrade silently to the letter-circle in the settings UI.
+    const missing = ADAPTERS.filter((adapter) => {
+      const { container } = render(<ProviderIcon adapter={adapter} />);
+      return isFallback(container);
+    });
+
+    expect(missing).toEqual([]);
+  });
+
+  it.each(ADAPTERS)("renders %s without a hard-coded black fill", (adapter) => {
+    const { container } = render(<ProviderIcon adapter={adapter} />);
+    const svg = container.querySelector("svg");
+
+    expect(svg).not.toBeNull();
+
+    // A monochrome mark painted pure black is invisible in dark mode. Marks
+    // that intentionally carry their own background tile are exempt, since the
+    // black is the tile rather than the glyph.
+    const hasOwnTile = svg!.querySelector("rect") !== null;
+    if (hasOwnTile) return;
+
+    const blackFills = Array.from(svg!.querySelectorAll("[fill]")).filter((el) => {
+      const fill = el.getAttribute("fill")?.toLowerCase();
+      return fill === "#000" || fill === "#000000" || fill === "black";
+    });
+
+    expect(blackFills).toEqual([]);
+  });
+
+  it("falls back to a letter circle for an unknown adapter", () => {
+    const { container } = render(<ProviderIcon adapter="not-a-provider" />);
+
+    expect(isFallback(container)).toBe(true);
+    expect(container.querySelector("text")?.textContent).toBe("N");
+  });
+
+  it("matches adapters case-insensitively", () => {
+    const { container } = render(<ProviderIcon adapter="OpenAI" />);
+
+    expect(isFallback(container)).toBe(false);
+  });
+
+  it("forwards className onto the rendered svg", () => {
+    const { container } = render(<ProviderIcon adapter="anthropic" className="h-6 w-6" />);
+
+    expect(container.querySelector("svg")?.getAttribute("class")).toContain("h-6");
+  });
+});
--- a/frontend/ui/src/components/icons/provider-icons.tsx
+++ b/frontend/ui/src/components/icons/provider-icons.tsx
@@ -1,47 +1,56 @@
 "use client";
 
-import { FC } from "react";
+import { FC, ReactNode, SVGProps } from "react";
+import type { LLMAdapter } from "@traceroot/core/llm-providers";
 import { cn } from "@/lib/utils";
+
+// Provider brand marks adapted from @lobehub/icons (MIT, © LobeHub).
+// Each mark is the provider's own trademark and follows its brand guidelines.
+// Monochrome marks use `currentColor` so they stay legible on light and dark
+// backgrounds; only marks that are brand-colored by definition hardcode a fill.
 
 interface IconProps {
   className?: string;
 }
 
+type ProviderSvgProps = IconProps &
+  Pick<SVGProps<SVGSVGElement>, "fill" | "fillRule"> & {
+    children: ReactNode;
+  };
+
+/**
+ * Shared wrapper owning every attribute the brand marks have in common, so an
+ * icon component declares only its own paths, gradients and fills.
+ */
+const ProviderSvg: FC<ProviderSvgProps> = ({ className, children, ...svgProps }) => (
+  <svg
+    viewBox="0 0 24 24"
+    className={cn("h-4 w-4", className)}
+    xmlns="http://www.w3.org/2000/svg"
+    aria-hidden="true"
+    {...svgProps}
+  >
+    {children}
+  </svg>
+);
+
 // OpenAI spiral
 export const OpenAIIcon: FC<IconProps> = ({ className }) => (
-  <svg
-    viewBox="0 0 24 24"
-    className={cn("h-4 w-4", className)}
-    xmlns="http://www.w3.org/2000/svg"
-    fill="currentColor"
-    aria-hidden="true"
-  >
+  <ProviderSvg className={className} fill="currentColor">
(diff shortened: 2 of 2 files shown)


XinweiHe:
@cubic review

XinweiHe:
Needs some before and after screenshots.

PR #1995, opened 2026-08-25
fix(alerts): carry the rule's filters into the Slack notification

GabrielDrapor:
https://github.com/user-attachments/assets/12522ad9-0a84-478e-94ad-51c5a3d54420

## Problem

Following "View spans" from an alert's Slack message landed on the trace list with only the evaluated window applied — the rule's filters were dropped, so a filtered rule looked like an unfiltered one. The prose had the same gap: two rules on the same measure that differ only by filter produced identical messages.

Reproduced with a rule `count > 1` filtered on `span_kind = AGENT`: the link carried `date_filter=custom&start&end` and nothing else.

## Fix

**Notification carries the filters** (`@traceroot/core`, `@traceroot/slack`, worker)
- The notification payload now includes `rule.filters` (older queued jobs render as unfiltered).
- The message states them: `Where \`span_kind = AGENT\` and \`metadata[tenant] contains acme\`.` — every filter, whether or not the link can carry it.
- The link carries whichever filters the trace list can express, as `?filters=` predicates (`alertFiltersToTracePredicates`): `=` → `in: [value]`; keyed `metadata` keeps its key with `eq`/`contains`.

**Trace list gains the alert's span dimensions** (`backend/rest/services/filters/columns.py`, `features/filters/registry.ts`)
- `span_kind`, `status` and `name` (Span name) join `model_name`/`environment` as `SPAN_MEMBERSHIP` categorical fields, so those alert filters can ride in the link. `filter-values` resolves them through the registry, so the value dropdown works without further changes.

**Semantics, stated honestly**
- Membership lowers as "trace has ≥1 span where …", looser than the span-grain predicate the alert evaluated, and the list shows traces rather than the counted spans — so the link is now labelled **View traces**.
- `contains` on a categorical field (the list only takes `in`) and `is_root` (no list column) stay in the prose only.
- A `metadata[key]` predicate maps too, one notch wider still: the alert matched the pair on spans only, while the trace list also accepts it on the trace row (`widget_registry.py` vs `translate.py`'s trace-row OR arm), so the link can include a trace tagged only at trace scope. Kept because it still narrows the list to the rule's tenant/user/flag, which beats an unfiltered list; noted in `alertFiltersToTracePredicates`.
- A span-grain drill-down that runs the alert's own widget query would close the gap completely; that's a separate feature, proposed as a follow-up.

**Saved filter rows read as loading, not empty** (`FilterRow`, `AlertFilters`)
- Opening "View alert" showed the saved filter as an empty field/op/value row until the widget schema resolved (several seconds in dev), which reads as "no filter". `FilterRow` takes `fieldsLoading`; a saved row renders as its raw predicate with a spinner until the registry answers. The dashboards widget builder is unaffected unless it opts in.
- Once the registry has answered, a saved field it does not know (retired, or never offered on this view) renders as its raw predicate with an "Unknown field" marker and stays removable, instead of the same empty row.

## Verification

- Unit: core, slack, worker, ui filters/alerts, backend filter/registry parity suites — all green; new cases cover the link, the prose, escaping, the mapping rules, and the loading row.
- Local e2e on `make dev-autoreload`: the filtered rule alerted, Slack received `[ALERT] … Where \`span_kind = AGENT\`.` with a **View traces** link; opening it shows the `span kind = AGENT` chip and the 1-minute window.

## Notes for the stack

- Base is the stack tip (`feat/alerts-ui-routes-pr`) because the change spans three of its PRs (#1891 blocks, #1900 delivery, #1901 scheduler).
- The tip does not yet include the review fixups sitting on the leaf branches (e.g. `0221fcfb` on #1891 also touches `alert-blocks.ts`); expect a small conflict there when those are folded in — happy to rebase.
- Observed while testing, not changed here: `dashboards.py` calls `run_widget_query` / `get_distinct_*_values` synchronously inside async handlers, so a ClickHouse query blocks the event loop and concurrent requests (e.g. `widgets/schema`) wait on it. The trace list already moved this off the loop in #1834; the widget endpoints should do the same.

https://claude.ai/code/session_01JcKGf8Zx34mmyFP8yMBrQX

---
## Summary by cubic
Slack alerts now include a rule’s filters in both the blocks and fallback text and link to a filtered trace list as “View traces”; previously the link only applied the time window and read “View spans.” Saved filter rows render as loading or “Fields unavailable” instead of empty, and unknown fields stay labeled and removable.

**Bug Fixes**
- Carry rule filters in the notification payload; older queued jobs render unfiltered.
- Map only list-expressible filters into the deep link (`?filters=`): `=` → one-value `in`; keyed `metadata` uses `eq`/`contains`; drop unsupported or incomplete rows; stringify numeric values.
- Add `span_kind`, `status`, and `name` to the trace list as `SPAN_MEMBERSHIP` categorical `in` fields.
- Use trace-membership semantics (“trace has ≥1 matching span”), so the deep link targets traces.
- Keep filters without a list counterpart (e.g., `contains` on categorical fields, `is_root`) in prose only.
- Match `metadata[key]` at trace or span scope, so drill-down can be slightly wider than evaluation.
- Protect mrkdwn with backtick-safe code spans and escape prose without corrupting the encoded link.
- Show saved filter rows as text with a spinner until the field schema resolves; on schema error, render the predicate under “Fields unavailable” and keep it removable; once answered, flag unknown fields even with an empty registry.
- Pass `fieldsLoading`/`fieldsUnavailable` from alerts and the widget builder so saved rows never read as empty or unknown before the registry resolves.

**Dependencies**
- Regenerate the public OpenAPI and the tools registry to include the new trace-list filter fields.

<sup>Written for commit 56bc64f8401a84f46c61308afba5c54adc137702. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/1995?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


XinweiHe:
1. a short video would be really help here.
2. for cubic reviews, either fix them, or if they are not as relevant left with a comment and resolve them.

XinweiHe (approved):

Issue #2044, opened 2026-08-27
`gpt-5.6-terra` is over-billed and missing from the published docs


dark-sorceror:
## Summary

`gpt-5.6-terra` is priced at **$2.50 / $15** per 1M tokens; the provider publishes **$2 / $12**. Every span on the model is over-billed by 25% on both input and output. **All four rates are wrong by exactly the same 1.25x factor**, not just the two headline ones:

| rate | committed $/1M | published $/1M | ratio |
|---|---|---|---|
| input | 2.5000 | 2.0000 | **1.25x** |
| output | 15.0000 | 12.0000 | **1.25x** |
| cacheRead | 0.2500 | 0.2000 | **1.25x** |
| cacheWrite | 3.1250 | 2.5000 | **1.25x** |
 The model is also missing from the system-model list and both published docs pages.

## How to replicate

Price a `gpt-5.6-terra` span and compare the reported cost against the published rate — it is high by the factor above.

## Root cause (confirmed)

`frontend/packages/core/src/standard-model-prices.json` carries `input: 2.5e-06`, `output: 1.5e-05`. Nothing ties a catalogue entry to the published rate, so the drift is silent — the mechanism described in #1597.

**Why tests did not catch it.** `TestGpt56CachePricing` (`tests/worker/tokens/test_pricing.py:234-245`) asserts cache prices as *ratios* of input — `cacheWrite` at 1.25×, `cacheRead` at 0.10× — never absolute values. A wrong base price stays internally consistent and passes. Concretely: because the drift here is *uniform*, `cacheRead / input` is exactly 0.10 and `cacheWrite / input` is exactly 1.25 on the committed values, so both invariants hold perfectly. A catalogue scaled by any single factor satisfies every assertion that currently touches this model — three resolution cases (`test_pricing.py:68-70`) and the two ratio checks. For reference, the sibling entries get their absolute-price assertions from #2051 (`sol`) and #2052 (`luna`), with #2053 still missing one for `cyber`.

## Surfaces

| Surface | State |
|---|---|
| `standard-model-prices.json` | present, mispriced |
| `llm-providers.ts` → `ADAPTER_MODELS` (BYOK dropdown) | present, under a now-stale "Limited partner preview as of 2026-07-09" comment |
[...]

PR #2051, opened 2026-08-27
fix(pricing): correct gpt-5.6-sol rates, resolve the bare gpt-5.6 alias

### LINKED ISSUE #2043: `gpt-5.6-sol` is over-billed, its documented alias prices as null, and it is missing from the published docs
## Summary

`gpt-5.6-sol` is priced at **$5 / $30** per 1M tokens; the provider publishes **$4 / $20**. Every span on the model is over-billed by 25% on input and 50% on output. Separately, the provider documents `gpt-5.6` as an alias for this model, and no catalogue pattern matches the bare alias, so a call using it resolves to no price at all. The model is also missing from the system-model list and both published docs pages.

## How to replicate

Price a `gpt-5.6-sol` span and compare against the published rate — the reported cost is high by the factors above.

For the alias, resolve `gpt-5.6` against the catalogue patterns: it matches no entry and yields a null price, which renders as absent rather than as an error. `gpt-5.6-sol` resolves correctly.

## Root cause (confirmed)

**Price.** `frontend/packages/core/src/standard-model-prices.json` carries `input: 5e-06`, `output: 3e-05`. Nothing ties a catalogue entry to the published rate, so drift is silent — the mechanism described in #1597.

**Alias.** The pattern is `^(openai\/|azure\/)?(gpt-5\.6-sol)(-[\d-]+)?$`. The variant suffix is mandatory, so the bare `gpt-5.6` cannot match.

**Why tests did not catch either.** `TestGpt56CachePricing` (`tests/worker/tokens/test_pricing.py:234-245`) asserts cache prices as *ratios* of input — `cacheWrite` at 1.25×, `cacheRead` at 0.10× — never absolute values. A wrong base price stays internally consistent and passes. The resolution cases at `test_pricing.py:64-67` cover the prefixed and dated forms but not the bare alias.

## Surfaces

| Surface | State |
|---|---|
| `standard-model-prices.json` | present, mispriced, pattern excludes the alias |
| `llm-providers.ts` → `ADAPTER_MODELS` (BYOK dropdown) | present, but under a comment reading "Limited partner preview as of 2026-07-09 — not yet accessible via self-serve API keys", which is now out of date |
| `llm-providers.ts` → `SYSTEM_MODELS` | absent — **not necessarily a gap**: system models run on TraceRoot's own API keys, so listing a frontier model here means TraceRoot pays its inference. Out of scope for this issue. |
| `docs/ai-agent/byok.mdx` | **absent** — the OpenAI row stops at `gpt-5.5` |
| `docs/integrations/internal-models.mdx` | **absent** — still calls `gpt-5.5` "OpenAI's most capable model" |
| `test_pricing.py`, `model-resolver.test.ts` | resolution cases present; no alias case, no absolute-price assertion |

## Impact

- Every recorded `gpt-5.6-sol` cost is overstated. The error reads as plausible — high, not impossible — so it is unlikely to be noticed from the UI.
- A caller using the documented `gpt-5.6` alias sees no cost at all.
- The published BYOK docs understate the supported set by a generation.
- Correcting the catalogue fixes future pricing only; already-recorded spans keep the wrong figure.

## Acceptance criteria

- `input` and `output` match the published $4 / $20, with `cacheRead` and `cacheWrite` recomputed from the corrected input so the existing ratio tests still hold.
- `gpt-5.6` resolves to this entry, with a resolution case alongside the existing prefixed and dated ones.
- **A test asserts the base prices against their published values.** This is the load-bearing item: the current suite pins internal ratios only and would have passed throughout this drift.
- The model appears in `byok.mdx` (it is already offered via BYOK), and the stale "limited partner preview" comment is corrected. **`SYSTEM_MODELS` is deliberately out of scope** — that list runs on TraceRoot's keys and adding a frontier model to it is a cost decision, not a bug fix.
- Whether already-recorded spans are repriced is decided explicitly and recorded.

mohdUwaish59:
gpt-5.6-sol was priced at $5/$30 per 1M tokens against a published $4/$20 (25% over on input, 50% on output). Separately, the provider documents gpt-5.6 as an alias for this model, but the catalogue pattern required the "-sol" suffix, so the alias resolved to no price at all.

- standard-model-prices.json: input/output corrected to $4/$20, cacheRead/ cacheWrite recomputed from the same 0.10x/1.25x ratios; matchPattern now also matches the bare gpt-5.6 alias.
- llm-providers.ts: removed the stale "limited partner preview" comment — the model is already in the self-serve BYOK dropdown.
- docs/ai-agent/byok.mdx: added gpt-5.6-sol to the OpenAI row.
- test_pricing.py: added the gpt-5.6 alias to OPENAI_MODEL_CASES; added TestGpt56SolPublishedPrices asserting the absolute input/output rate — the existing TestGpt56CachePricing only checks internal ratios, so a wrong base price stayed internally consistent and passed throughout this drift.

Repricing decision: already-recorded gpt-5.6-sol spans are NOT repriced. This fix corrects pricing going forward only. No backfill exists in this codebase for retroactively repricing spans when the catalogue changes (backfill_eval_result_costs in ingest_tasks.py is unrelated — it fills in missing costs from an ingest race, not catalogue drift), and writing one against production billing data is a materially different, riskier piece of work than a catalogue fix.

Fixes #2043

## Type of Change

- [ ] feat - A new feature
- [x] fix - A bug fix

## Checklist

---
## Summary by cubic
Fixes gpt-5.6-sol pricing to match the published $4/$20 per 1M tokens and resolves the bare `gpt-5.6` alias to the same price.

- Corrects input/output to $4/$20 and recomputes cache rates from the same 0.10x/1.25x ratios.
- Expands the price match pattern to accept `gpt-5.6` without the `-sol` suffix.
- Removes the stale "limited partner preview" comment.
- Adds a test asserting absolute rates, so a wrong base price no longer goes undetected by ratio checks.

Already-recorded spans are not repriced; this affects pricing going forward only.

<sup>Written for commit 8bcbff2742798964247408289a04403ee835935b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/2051?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Saisharathchandranandnetha:
Thanks for addressing #2043.

I noticed the same documentation consistency issue raised by Cubic:

> `gpt-5.6-sol` is now the first entry in the **“Default Models”** table, but it is currently only in `ADAPTER_MODELS` and not `SYSTEM_MODELS`, so `gpt-5.5` remains the actual system default.

I agree the documentation should either keep `gpt-5.5` first or clearly distinguish **BYOK model availability** from the **system-model default**.

Otherwise, the changes appear to cover the main requirements from #2043.

tests/worker/tokens/test_pricing.py
```diff
@@ -65,6 +65,8 @@ def real_cache() -> list[dict]:
     ("openai/gpt-5.6-sol", "gpt-5.6-sol"),
     ("azure/gpt-5.6-sol", "gpt-5.6-sol"),
     ("gpt-5.6-sol-2026-07-09", "gpt-5.6-sol"),
+    # gpt-5.6 is the provider-documented alias for gpt-5.6-sol (#2043).
```
dark-sorceror: Drop the issue ref the PR body already links it

tests/worker/tokens/test_pricing.py
```diff
@@ -245,6 +247,19 @@ def test_cache_read_is_90_percent_discount(self, real_cache, model_name):
         assert entry["prices"]["cacheRead"] == pytest.approx(entry["prices"]["input"] * 0.10)
 
 
+class TestGpt56SolPublishedPrices:
+    """Ratio checks alone let a wrong base price stay internally consistent and
+    pass (#2043) — gpt-5.6-sol was billed at $5/$30 per 1M tokens against a
```
dark-sorceror: ditto

dark-sorceror (changes requested):
Prices and the alias fix both verified correct. Two comment refs to drop.

tests/worker/tokens/test_pricing.py
mohdUwaish59: done

tests/worker/tokens/test_pricing.py
mohdUwaish59: done

dark-sorceror:
LGTM

dark-sorceror (approved):

XinweiHe:
Thanks for the thorough work here — the test rationale is exactly right, and the "ratio checks let a wrong base price stay internally consistent" framing is the real lesson from this class of bug.

That said, I want to hold this before merge, because I can't reconcile the target price with what OpenAI publishes.

## The $4 / $20 figure

`gpt-5.6-sol` looks like it is **correctly priced at $5 / $30 today**, and was not part of the 2026-07-30 price cut.

That cut reduced Luna by 80% and Terra by 20%, and left Sol unchanged. What convinces me is the cross-source consistency rather than any single page:

- every source showing Terra at **$2.50 / $15** also shows Luna at **$1.00 / $6.00** — pre-cut snapshots
- every source showing Terra at **$2.00 / $12** also shows Luna at **$0.20 / $1.20** — post-cut snapshots
- **both groups show Sol at $5 / $30.** It's the one figure that doesn't move across that split, and I can't find $4 / $20 published anywhere, at any point in time.

If that's right, this PR would take Sol from correct to a **20% under-bill on input and 33% on output** — the inverse of the bug it's fixing — and `TestGpt56SolPublishedPrices` would pin the wrong number in place, so it stops reading as drift.

I'd rather be wrong here than merge it, so: **can you link the source for $4 / $20?** If OpenAI publishes that and my reading is off, I'll happily eat this comment and we merge.

To be clear, this isn't on you — #2043 specifies `$4 / $20` in its acceptance criteria and you implemented the issue faithfully. The bad figure entered upstream of this PR, which is itself an instance of what the issue describes: nothing in the loop ties a catalogue entry back to the provider's published rate.

## Two smaller things

**The `gpt-5.6` alias.** #2043 states the provider documents bare `gpt-5.6` as an alias for `gpt-5.6-sol`, but I couldn't find that documented either — the model pages I checked treat Sol/Terra/Luna as distinct ids without defining a bare-name default. Since the pattern change routes an undocumented name to the most expensive tier in the family, this needs a citation too. (The regex itself is sound — I checked that `(-[\d-]+)?` is digits-only, so `-terra`/`-luna` can't be swallowed, and no other catalogue entry matches bare `gpt-5.6`.)

**The dropped comment in `llm-providers.ts`.** `// Limited partner preview as of 2026-07-09 — not yet accessible via self-serve API keys` sits above all three `gpt-5.6-*` entries, not just `sol`. And "it's already in the self-serve BYOK dropdown" is circular as a justification — that comment was annotating exactly that list. If Sol went self-serve but Terra/Luna didn't, let's narrow the comment rather than delete it.

## Where this leaves us

- **Sol (this PR):** blocked pending the price citation.
- **Terra (#2123):** confirmed correct — $2 / $12 verified against live provider pricing. Merging that one.
- **Luna (#2052):** catalogue has $1 / $6 against a published $0.20 / $1.20 — a **5x over-bill** running in production right now. That's the most urgent of the three.

Also worth noting for whoever lands second: this PR and #2123 conflict in `byok.mdx` and at the shared test-class anchor in `test_pricing.py`. #2044 suggests collapsing these into one parametrised `PUBLISHED_RATES` table instead of a per-model test class, which would remove the collision for the rest of the family — worth doing as part of whichever lands first.

XinweiHe:
I was wrong. Withdrawing my objection — **this PR is correct on both counts.**

I checked OpenAI's model page directly this time ([developers.openai.com/api/docs/models/gpt-5.6-sol](https://developers.openai.com/api/docs/models/gpt-5.6-sol)):

> **Text tokens — Per 1M tokens**
> Input **$4.00** · Cached input **$0.40** · Output **$20.00**
>
> "GPT-5.6 Sol costs $4 per million input tokens and $20 per million output tokens, a 20% reduction in input pricing and a 33% reduction in output pricing."
>
> "**The `gpt-5.6` alias routes requests to GPT-5.6 Sol.**"

So `$4 / $20` is right, and the bare-alias behaviour is documented exactly as #2043 described. Every number in the diff checks out against that page — cacheRead `4e-7` = the published $0.40 cached input, and cacheWrite `5e-6` = the documented 1.25x of uncached input.

**Where I went wrong:** I found the 2026-07-30 cut (Luna −80%, Terra −20%, Sol unchanged) and concluded from several sources agreeing on $5/$30 that Sol had never moved. It moved later — there was a second, Sol-specific cut on 2026-08-21 that every page I'd read predated. My "both groups show Sol at $5/$30" argument was consistent and confidently reasoned, and simply wrong: I'd sampled only stale snapshots and treated their agreement as confirmation. I should have gone to the provider's own model page first instead of triangulating secondary sources. Apologies for the round trip.

My third point — the `// Limited partner preview` comment — I'll also drop. Sol and Luna both show as `Default` models with standard per-tier rate limits, so the comment was stale for the whole family and removing it is right.

Two things I noticed on that page that are **not** objections to this PR, just worth capturing separately:

1. **The $4/$20 is explicitly promotional** — "available at least through November 21, 2026." The catalogue has no notion of an expiry, so this will silently become an under-bill if it reverts. Worth an issue with a November reminder rather than anything here.
2. **Long-context surcharge is unmodelled.** "Prompts with >272K input tokens are priced at 2x input and 1.5x output for the full request." With a 1,050,000-token context window that's reachable, and neither the catalogue nor `cost_from_buckets` represents it — so any span over 272K input is under-billed across the whole gpt-5.6 family. Separate issue, separate PR.

Thanks for your patience, and sorry for the delay this caused — merging.

PR #2053, opened 2026-08-28
feat(pricing): add gpt-5.6-cyber to OpenAI BYOK catalog and pricing table (#2046)

### LINKED ISSUE #2046: `gpt-5.6-cyber` is unpriced and absent from every surface
## Summary

`gpt-5.6-cyber` — cybersecurity-focused GPT-5.6 variant — appears in no surface of the product. A span recorded against it resolves to no price and renders as costless, and the model cannot be selected anywhere in the UI.

## How to replicate

Resolve `gpt-5.6-cyber` against the catalogue patterns: no entry matches, so `getModelPricing` returns `null` and the cost column reads as absent rather than as an error. The model also does not appear in the BYOK dropdown, the system-model list, or either docs page.

## Root cause (confirmed)

The model is simply not in `frontend/packages/core/src/standard-model-prices.json`, so there is no pattern to match and no price to return. It was published after the catalogue was last updated by hand, and nothing reconciles the catalogue against the provider's published model list — the mechanism described in #1597.

A missing entry fails quietly by design: an unmatched model returns `null` rather than raising, so nothing surfaces the gap.

## Surfaces

| Surface | State |
|---|---|
| `standard-model-prices.json` | **absent** — no price |
| `llm-providers.ts` → `ADAPTER_MODELS` (BYOK dropdown) | **absent** |
| `llm-providers.ts` → `SYSTEM_MODELS` | **absent** |
| `docs/ai-agent/byok.mdx` | **absent** |
| `test_pricing.py` | no resolution case |

## Impact

- Any usage of this model is recorded at no cost, so spend on it is invisible in per-trace, per-eval and usage reporting. A missing price is less dangerous than a wrong one — it reads as absent rather than authoritative — but it is silently wrong all the same.
- The model cannot be chosen through BYOK or as a system model, so the product understates what it supports.
- Whether this model belongs in the catalogue at all is worth confirming: pricing is not published on the models page alongside the frontier tier, so the rate must be sourced before an entry can be added.

## Acceptance criteria

- Confirm the published input/output rates from an authoritative source before adding anything — do not infer them from a sibling model.
- A catalogue entry exists with a pattern covering the bare id and the `openai/` and `azure/` prefixed forms, consistent with the neighbouring GPT-5.6 entries.
- A resolution test covers the bare and prefixed forms.
- A test asserts the base prices against their published values, so the entry cannot drift silently the way the rest of the GPT-5.6 family did.
- The model is added to the BYOK dropdown and both docs pages, or a note records why it is deliberately excluded.

Gaurav-meena95:
### Summary

Closes #2046

Adds OpenAI's **`gpt-5.6-cyber`** (released under the Daybreak Red security program) to the BYOK model catalog and standard pricing table. Previously, spans recorded against this model resolved to no price ($0.00 cost) and could not be selected in the BYOK model settings dropdown.

### Published Rates

Rates are sourced directly from OpenAI's Daybreak Red announcement:
- **Input:** $12.50 / 1M tokens (`$0.0000125` / token)
- **Output:** $75.00 / 1M tokens (`$0.000075` / token)
- **Cache Read:** $1.25 / 1M tokens (`$0.00000125` / token, 90% discount matching GPT-5.6 family)
- **Cache Write:** 1.25× input rate (`$0.000015625` / token)

### Changes Made

1. **Pricing Catalog (`frontend/packages/core/src/standard-model-prices.json`):**
   - Added `tr-gpt-5-6-cyber` with match pattern `(?i)^(openai\/|azure\/)?(gpt-5\.6-cyber)(-[\\d-]+)?$` and base + cache pricing.
2. **BYOK Catalog (`frontend/packages/core/src/llm-providers.ts`):**
   - Added `gpt-5.6-cyber` to `ADAPTER_MODELS.openai` alongside the other GPT-5.6 preview models.
3. **Unit Tests (`tests/worker/tokens/test_pricing.py`):**
   - Added resolution test cases for bare, `openai/`, and `azure/` prefixed forms to `OPENAI_MODEL_CASES`.
   - Parameterized `gpt-5.6-cyber` in `TestGpt56CachePricing` to assert cache-write (1.25×) and cache-read (90% discount) calculations.

### Validation

- [x] All 237 Python pricing tests passed:
  ```bash
  uv run pytest tests/worker/tokens/test_pricing.py
  
  
  
Before :- 

<img width="3024" height="1964" alt="Image" src="https://github.com/user-attachments/assets/c9106823-30ed-46fd-be87-e68c7f4e6de5" />

After :- 

<img width="3024" height="1964" alt="Image" src="https://github.com/user-attachments/assets/299ef1b7-0877-4ccb-b5ac-06ea14d7961e" />
  

<!-- This is an auto-generated description by cubic. -->
---
## Summary by cubic
Adds `gpt-5.6-cyber` to the OpenAI BYOK model catalog and standard pricing table. Previously this model resolved to a $0.00 cost and couldn't be selected in the BYOK settings dropdown.

**Pricing**
- Input: $12.50 / 1M tokens; output: $75.00 / 1M tokens.
- Cache read: $1.25 / 1M tokens (90% discount); cache write: 1.25× input rate.
- Adds resolution and cache pricing tests for bare, `openai/`, and `azure/` prefixed forms.
- Pins the model's input/output rates to the provider-published values so a wrong base price can't pass via ratio checks.
- Removes the stale preview-access note from the catalog.

<sup>Written for commit e9945d078d69c7cc241759163f54cf07e9c4b3d9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/2053?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg"><img alt="Review in cubic" src="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"></picture></a>

<!-- End of auto-generated description by cubic. -->

### DIFF
--- a/frontend/packages/core/src/llm-providers.ts
+++ b/frontend/packages/core/src/llm-providers.ts
@@ -155,7 +155,6 @@ export const PROVIDER_PRIORITY: LLMAdapter[] = [
 // Adapters NOT listed here (azure, amazon-bedrock, openrouter) use free-text input.
 export const ADAPTER_MODELS: Partial<Record<LLMAdapter, LLMModelDef[]>> = {
   openai: [
-    // Limited partner preview as of 2026-07-09 — not yet accessible via self-serve API keys.
     { id: "gpt-5.6-sol", label: "gpt-5.6-sol" },
     { id: "gpt-5.6-terra", label: "gpt-5.6-terra" },
     { id: "gpt-5.6-luna", label: "gpt-5.6-luna" },
--- a/frontend/packages/core/src/llm-providers.ts
+++ b/frontend/packages/core/src/llm-providers.ts
@@ -159,6 +159,7 @@ export const ADAPTER_MODELS: Partial<Record<LLMAdapter, LLMModelDef[]>> = {
     { id: "gpt-5.6-sol", label: "gpt-5.6-sol" },
     { id: "gpt-5.6-terra", label: "gpt-5.6-terra" },
     { id: "gpt-5.6-luna", label: "gpt-5.6-luna" },
+    { id: "gpt-5.6-cyber", label: "gpt-5.6-cyber" },
     { id: "gpt-5.5", label: "gpt-5.5" },
     { id: "gpt-5.5-pro", label: "gpt-5.5-pro" },
     { id: "gpt-5.4", label: "gpt-5.4" },
--- a/frontend/packages/core/src/standard-model-prices.json
+++ b/frontend/packages/core/src/standard-model-prices.json
@@ -205,6 +205,18 @@
     }
   },
   {
+    "id": "tr-gpt-5-6-cyber",
+    "modelName": "gpt-5.6-cyber",
+    "matchPattern": "(?i)^(openai\\/|azure\\/)?(gpt-5\\.6-cyber)(-[\\d-]+)?$",
+    "provider": "openai",
+    "prices": {
+      "input": 0.0000125,
+      "output": 0.000075,
+      "cacheRead": 0.00000125,
+      "cacheWrite": 0.000015625
+    }
+  },
+  {
     "id": "tr-gpt-5-5",
     "modelName": "gpt-5.5",
     "matchPattern": "(?i)^(openai\\/|azure\\/)?(gpt-5\\.5)(-[\\d-]+)?$",
--- a/tests/worker/tokens/test_pricing.py
+++ b/tests/worker/tokens/test_pricing.py
@@ -71,6 +71,9 @@ OPENAI_MODEL_CASES = [
     ("gpt-5.6-luna", "gpt-5.6-luna"),
     ("openai/gpt-5.6-luna", "gpt-5.6-luna"),
     ("azure/gpt-5.6-luna", "gpt-5.6-luna"),
+    ("gpt-5.6-cyber", "gpt-5.6-cyber"),
+    ("openai/gpt-5.6-cyber", "gpt-5.6-cyber"),
+    ("azure/gpt-5.6-cyber", "gpt-5.6-cyber"),
     ("gpt-5.5", "gpt-5.5"),
     ("openai/gpt-5.5", "gpt-5.5"),
     ("azure/gpt-5.5", "gpt-5.5"),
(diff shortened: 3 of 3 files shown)


tests/worker/tokens/test_pricing.py
```diff
@@ -234,12 +237,16 @@ def test_latest_openai_model_calculates_cost(self, real_cache):
 class TestGpt56CachePricing:
     """gpt-5.6 is the first OpenAI family with non-null cacheWrite pricing."""
 
-    @pytest.mark.parametrize("model_name", ["gpt-5.6-sol", "gpt-5.6-terra", "gpt-5.6-luna"])
+    @pytest.mark.parametrize(
```
dark-sorceror, line 252: These compare cache rates to input, so they hold for any input as long as the ratios move with it. Add an assertion pinning input and output to the published $12.50 / $75 absolutely.

frontend/packages/core/src/llm-providers.ts
```diff
@@ -159,6 +159,7 @@ export const ADAPTER_MODELS: Partial<Record<LLMAdapter, LLMModelDef[]>> = {
     { id: "gpt-5.6-sol", label: "gpt-5.6-sol" },
     { id: "gpt-5.6-terra", label: "gpt-5.6-terra" },
     { id: "gpt-5.6-luna", label: "gpt-5.6-luna" },
+    { id: "gpt-5.6-cyber", label: "gpt-5.6-cyber" },
```
dark-sorceror, line 161: Added to the self-serve dropdown directly under a comment saying these models aren't self-serve accessible. That line is stale — worth dropping it here.

frontend/packages/core/src/standard-model-prices.json
```diff
@@ -204,6 +204,18 @@
       "cacheWrite": 0.00000125
     }
   },
+  {
+    "id": "tr-gpt-5-6-cyber",
+    "modelName": "gpt-5.6-cyber",
+    "matchPattern": "(?i)^(openai\\/|azure\\/)?(gpt-5\\.6-cyber)(-[\\d-]+)?$",
```
dark-sorceror, line 210: Pattern is right. Note docs are untouched — either add a prose line in byok.mdx or say in the PR body why it's deliberately absent (the Default Models table is locked to the system-model list, so a row there fails CI).

dark-sorceror (changes requested):
Rates are correct and the pattern matches the family convention. The blocker is that nothing pins them: I raised input and output by 50% and recomputed the cache rates at the same 0.10x/1.25x ratios, and the suite still passed 237/237 — a silent $18.75/$112.50 overbill. The sibling fixes both go red under that mutation.

tests/worker/tokens/test_pricing.py
dark-sorceror: Resolved in 7321e99b6

Added `TestGpt56CyberPublishedPrices`, pinning input and output to the published $12.50 / $75 per 1M absolutely. Mutation-checked: multiplying all four of the model's rates by 1.5 keeps every ratio intact and the cache tests green, but reds this new assertion and only this one. The published figures were confirmed against the provider's model page, including cached input at $1.25 and cache write at 1.25x.

frontend/packages/core/src/llm-providers.ts
dark-sorceror: Resolved in e9945d078

Dropped the preview-access note. It claimed the family was not reachable with a self-serve key, which is now falsified by the catalogue itself: real spans were ingested and billed on a sibling model, and the published docs already advertise one as selectable. Removed rather than reworded, since the only other thing it implied — absence from the system model list — is not distinctive to this family.

frontend/packages/core/src/standard-model-prices.json
dark-sorceror: Resolved — docs are deliberately untouched.

A row in the Default Models table is machine-locked to the system model list and would fail CI, and this model is not a system model. A one-off prose sentence naming a single model is the pattern being removed elsewhere, since each new model then needs the same manual edit. The BYOK section is being changed separately to point at the in-app model selector instead of enumerating models, which covers this one without a per-model edit.

dark-sorceror (approved):
All three review points are addressed and resolved above.

Verified independently rather than on inspection: merged against current main and ran the pricing suite (240 passed). The new absolute-rate assertion was mutation-checked — multiplying all four of the model's rates by 1.5 preserves every ratio, keeps the cache tests green, and reds only that one assertion. All four published rates ($12.50 / $75 / $1.25 cached / 1.25x cache write) were confirmed against the provider's model page, and the match pattern covers the bare, `openai/` and `azure/` forms consistent with its neighbours.

One gap left deliberately out of scope, as with the rest of the family: prompts over the long-context threshold are billed at a higher rate for the whole request, and the catalogue has no concept of a context-length tier.

PR #2054, opened 2026-08-28
docs: document claude-fable-5 support via BYOK (#2050)

### LINKED ISSUE #2050: `claude-fable-5` is priced and selectable via BYOK but absent from both docs pages
## Summary

`claude-fable-5` is present in the pricing catalogue at $10 / $50 and offered in the BYOK dropdown, but neither published docs page mentions it. A user reading the docs would not know the product supports it. Whether it also belongs in `SYSTEM_MODELS` is a separate decision this issue asks to settle rather than assume.

## How to replicate

Configure an Anthropic BYOK provider: `claude-fable-5` appears in the model dropdown. Then read `docs/ai-agent/byok.mdx` and `docs/integrations/internal-models.mdx` — neither lists it among the supported Anthropic models.

## Root cause (confirmed)

The model was added to the catalogue and `ADAPTER_MODELS` but not to either docs table. The lists are maintained independently with no shared source, so an addition to one does not propagate.

Its absence from `SYSTEM_MODELS` may be deliberate rather than an oversight: Fable 5 requires 30-day data retention and is not available under zero-data-retention configurations, which is a reason a hosted default might deliberately exclude it. That distinguishes it from `claude-opus-5`, whose omission from the same list is a straightforward gap.

## Surfaces

| Surface | State |
|---|---|
| `standard-model-prices.json` | present, priced correctly at $10 / $50 |
| `llm-providers.ts` → `ADAPTER_MODELS` (BYOK dropdown) | present |
| `llm-providers.ts` → `SYSTEM_MODELS` | absent — **out of scope.** System models run on TraceRoot's own keys, so listing Fable 5 at $10 / $50 is a cost decision. Its 30-day data-retention requirement is a second, independent reason to exclude it. |
| `docs/ai-agent/byok.mdx` | **absent** |
| `test_pricing.py` | resolution cases present |

## Impact

- The docs understate the supported model set: a user choosing a provider from the documentation would not know the model is available.
- No pricing risk — the catalogue entry is correct.
- No functional gap for BYOK users, who can already select it. A hosted workspace cannot reach it, but that is the expected consequence of a system-model list TraceRoot pays for.

## Acceptance criteria

- Both docs tables list `claude-fable-5` with an accurate description of where it is available (BYOK, and system models only if the decision below goes that way).
- A comment in `llm-providers.ts` records why Fable 5 is absent from `SYSTEM_MODELS` — cost, data-retention, or both — so the next reader does not "fix" a deliberate exclusion.
- No change to the pricing entry.

Gaurav-meena95:
### Summary

Closes #2050

Documents that **`claude-fable-5`** is available and supported via BYOK across the documentation pages (`docs/ai-agent/byok.mdx` and `docs/integrations/internal-models.mdx`).

Also adds an internal code comment in `SYSTEM_MODELS` (`frontend/packages/core/src/llm-providers.ts`) explaining that `claude-fable-5` is deliberately excluded from hosted system models due to cost ($10 / $50 per 1M tokens) and Anthropic's 30-day data retention requirement.

### Changes Made

1. **`docs/ai-agent/byok.mdx`**: Updated the Supported Providers section to explicitly mention `claude-fable-5` availability under BYOK.
2. **`docs/integrations/internal-models.mdx`**: Added a note in the BYOK section clarifying that models like `claude-fable-5` are available via BYOK.
3. **`frontend/packages/core/src/llm-providers.ts`**: Added an explanatory comment above the Anthropic section in `SYSTEM_MODELS` detailing the rationale for its exclusion from hosted defaults.

### Validation

- [x] All 134 frontend `@traceroot/core` unit & doc synchronization tests passed:
  ```bash
  pnpm --filter @traceroot/core test

<!-- This is an auto-generated description by cubic. -->
---
## Summary by cubic
Closes #2050. Adds a code comment in `llm-providers.ts` explaining that `claude-fable-5` is deliberately excluded from hosted system models due to cost and Anthropic's 30-day data retention requirement, and is only available via BYOK.

<sup>Written for commit 76fe4079fcec58b2704789502c5ec6aa423b2e60. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/2054?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg"><img alt="Review in cubic" src="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"></picture></a>

<!-- End of auto-generated description by cubic. -->

### DIFF
--- a/frontend/packages/core/src/llm-providers.ts
+++ b/frontend/packages/core/src/llm-providers.ts
@@ -102,6 +102,9 @@ export const SYSTEM_MODELS: {
     envVar: "ANTHROPIC_API_KEY",
     piAIProvider: "anthropic",
     apiProtocol: "anthropic-messages",
+    // claude-fable-5 is deliberately excluded from SYSTEM_MODELS: it runs at $10 / $50
+    // and requires 30-day data retention (not available under zero-data-retention
+    // configurations). It is available exclusively via BYOK (ADAPTER_MODELS.anthropic).
     models: [
       { id: "claude-opus-4-8", label: "claude-opus-4-8" },
       { id: "claude-opus-4-7", label: "claude-opus-4-7" },
--- a/docs/ai-agent/byok.mdx
+++ b/docs/ai-agent/byok.mdx
@@ -16,7 +16,7 @@ See [Supported Models](/integrations/internal-models) for a description of each 
 
 ## Supported Providers
 
-OpenAI, Anthropic, Azure OpenAI, Google (Gemini), Amazon Bedrock, xAI, DeepSeek, OpenRouter, Kimi, GLM.
+OpenAI, Anthropic, Azure OpenAI, Google (Gemini), Amazon Bedrock, xAI, DeepSeek, OpenRouter, Kimi, GLM. With BYOK, additional models such as `claude-fable-5` are also selectable.
 
 ## Setup
 
--- a/docs/ai-agent/byok.mdx
+++ b/docs/ai-agent/byok.mdx
@@ -16,7 +16,7 @@ See [Supported Models](/integrations/internal-models) for a description of each 
 
 ## Supported Providers
 
-OpenAI, Anthropic, Azure OpenAI, Google (Gemini), Amazon Bedrock, xAI, DeepSeek, OpenRouter, Kimi, GLM. With BYOK, additional models such as `claude-fable-5` are also selectable.
+OpenAI, Anthropic, Azure OpenAI, Google (Gemini), Amazon Bedrock, xAI, DeepSeek, OpenRouter, Kimi, GLM.
 
 ## Setup
 
--- a/docs/integrations/internal-models.mdx
+++ b/docs/integrations/internal-models.mdx
@@ -34,7 +34,7 @@ TraceRoot Cloud provides the following models out of the box — no API key requ
 
 ## Bring Your Own Key (BYOK)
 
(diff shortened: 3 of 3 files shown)


frontend/packages/core/src/llm-providers.ts
```diff
@@ -102,6 +102,9 @@ export const SYSTEM_MODELS: {
     envVar: "ANTHROPIC_API_KEY",
     piAIProvider: "anthropic",
     apiProtocol: "anthropic-messages",
+    // claude-fable-5 is deliberately excluded from SYSTEM_MODELS: it runs at $10 / $50
```
dark-sorceror: Non-blocking: `$10 / $50` is correct today — `standard-model-prices.json` has `tr-claude-fable-5` at `input: 0.00001` / `output: 0.00005` — but restating the numbers here duplicates the catalogue, and this comment is not somewhere anyone will think to look when a rate changes. Consider pointing at the catalogue entry instead ("see `tr-claude-fable-5` in `standard-model-prices.json`") and keeping the qualitative reason — that it is priced well above the Opus tier the system models sit at — so the rationale survives a repricing.

dark-sorceror (approved):

PR #2056, opened 2026-08-29
perf(alerts): stop re-evaluating wide windows every minute

Prasad8830:
Fixes #2025

> **Base branch:** this targets `feat/alerts-ui-routes-pr` (#1907), not `main`

## Summary

`claimDueAlerts` re-armed every claimed rule with the tick's own `nextRunAt`, so every active rule was re-evaluated once a minute no matter how wide its window. A 2h rule was re-measured 60x/hour over data that was 59/60ths the same. The re-arm now comes from the rule itself:

```
nextRunAt = boundary + min(windowMs, 5m)
```

The 1m rule remains at 60 runs/hour; every wider window drops to 12, an ~80% cut in evaluation queries and per-row claim writes for those rules.

## Type of Change

- [x] perf - A code change that improves performance

## Details

- **`tick.ts`** - adds `ALERT_CADENCE_CAP_MS` (5m), `alertCadenceMs(window)` and `alertNextRunAt(tick, window)`. `nextRunAt` is dropped from `AlertTick` / `computeAlertTick`: the tick no longer has an opinion about when any individual rule runs next, and it had no other reader.
- **`claim.ts`** - one line, `nextRunAt: alertNextRunAt(tick, row.window)`.

Why `min()` and not a flat 5 minutes: the cadence is never longer than the window, so consecutive evaluation windows abut at the cap and overlap below it. **No span can fall between two evaluations.** A flat cap would blind a 1m rule for 4 minutes out of every 5.

The cadence is read off the raw `row.window` string rather than a parsed rule, so the claim still lands before `parseAlertRule` — an unevaluable row must have `nextRunAt` advanced or it stays permanently due. A window this built cannot read keeps the tick's minute: such a row is claimed and discarded every minute either way, and backing it off would only delay the error its owner is waiting to read. #2029 (PARKED status) is the real fix for those.

No stale cadence after an edit: `alertStateReset()` already sets `nextRunAt = now`, so narrowing a rule's window (or resuming it from pause) makes it due on the next tick rather than at the end of a wide re-arm.

Rules re-arm from their own claim rather than a shared schedule, so wide-window rules stagger themselves across the five minutes instead of all landing on one tick.

**The trade, stated plainly:**

- A breach on a window >= 5m is now noticed within 5 minutes rather than 1.
[...]

PR #2081, opened 2026-09-01
perf(alerts): claim the batch in one statement instead of one write per rule

### LINKED ISSUE #2026: Alerts: claim rules in one statement instead of findMany + per-row updateMany
Follow-up from review of the alerts stack (#1889–#1907). Not a merge blocker; matters as rule count grows.

## Problem

`claimDueAlerts` (`frontend/worker/src/alerts/claim.ts`) does a `findMany` of due rules and then one `updateMany` per row as the CAS claim — with a full tick's budget of 500 rules, that is 500 UPDATE round-trips to Postgres per minute.

## Suggested fix

A single statement claims the same set with the same token and drops the N+1:

```sql
UPDATE alerts SET last_claimed_at = $tick
WHERE id IN (
  SELECT id FROM alerts
  WHERE status = 'ACTIVE' AND (next_run_at IS NULL OR next_run_at <= $now)
  ORDER BY next_run_at ASC NULLS FIRST
  LIMIT $budget
  FOR UPDATE SKIP LOCKED
)
RETURNING *;
```

`FOR UPDATE SKIP LOCKED` also preserves the current semantics under overlapping ticks (the CAS on `lastClaimedAt` exists for exactly that race). Prisma needs `$queryRaw` for this shape.

Together with per-rule cadence (filed separately), this moves the platform ceiling from a few hundred rules to a few thousand.

Prasad8830:
Fixes #2026

> This branch is stacked on perf/alerts-window-cadence, so the diff currently includes that commit too — once #2056 merges and its branch is deleted, GitHub auto-retargets this PR to feat/alerts-ui-routes-pr and the diff shrinks to just this PR's own two commits, no action needed here.

## Summary

`claimDueAlerts` read the due rules with `findMany` and then took each one with its own `updateMany` CAS, so a full tick's budget of 500 rules was 500 UPDATE round-trips a minute — queued ten at a time against the Prisma pool the detector run consumers share. Selecting, claiming and reading the batch back is now one statement:

```sql
WITH due AS (
  SELECT a.id, a.next_run_at, a.create_time,
         row_number() OVER (PARTITION BY a.project_id
                            ORDER BY a.next_run_at ASC, a.create_time ASC) AS depth
  FROM alerts a
  JOIN projects p ON p.id = a.project_id AND p.delete_time IS NULL
  WHERE a.status = $1 AND (a.next_run_at IS NULL OR a.next_run_at <= $2::timestamp)
  ORDER BY a.next_run_at ASC, a.create_time ASC
  LIMIT 1000
),
picked AS (SELECT id FROM due ORDER BY depth ASC, next_run_at ASC, create_time ASC LIMIT 500),
locked AS MATERIALIZED (
  SELECT id FROM alerts WHERE id IN (SELECT id FROM picked) FOR UPDATE SKIP LOCKED
)
UPDATE alerts SET last_claimed_at = $2::timestamp,
                  next_run_at = CASE "window" WHEN '1m' THEN … ELSE … END
WHERE id IN (SELECT id FROM locked)
RETURNING <the columns `AlertRowLike` reads>
```

501 statements per tick become 1.

## Type of Change

## Details

- **`claim.ts`** — `claimDueAlerts` is the statement above plus the parse loop. `claimRow`, `shareBudgetAcrossProjects` and `ALERT_CLAIM_CONCURRENCY` go: the first two moved into SQL, and there is nothing left to bound the concurrency of.
- **`claim.test.ts`** — mocks `$queryRaw` instead of `findMany`/`updateMany`. The claim assertions read the statement back with its parameters substituted, so what a tick selects and what it writes stay readable.

The four behaviours the old code carried, and where each went:

| Behaviour | Now |
|---|---|
| Soft-deleted projects never page | `JOIN projects … p.delete_time IS NULL` |
| Scan wider than the budget | `due … LIMIT ALERT_CLAIM_SCAN_LIMIT` |
| No project fills the budget alone | `row_number()` per project, `picked` ordered by depth first |
| Per-rule cadence (#2056) | `CASE "window"` built from `ALERT_WINDOWS` |

**The cadence `CASE` holds whole instants, not milliseconds.** `tick.boundary` is fixed for the tick and the token set is small, so each branch is an answer `alertNextRunAt` already gave. `tick.ts` stays the only place the cadence is defined and no date arithmetic moves into SQL. `ELSE` is the unreadable-window fallback, so a window this build cannot read still keeps the tick's minute exactly as before.

**Three deviations from the issue's sketch**, each deliberate:

- **No `NULLS FIRST`.** The snippet has it; ordering nulls first is the bug `claim.test.ts` already records — it put every rule created since the last tick ahead of every rule that was actually due, so a burst of new rules preempted the schedule. Postgres `ASC` defaults to NULLS LAST, which is what Prisma's `orderBy` does today.
- **A separate `locked` CTE.** A locking clause cannot share a query level with a window function, so the round-robin and the lock cannot be the same scan. `MATERIALIZED` so the lock is taken once over the chosen set rather than folded back into the `UPDATE`.
- **Timestamps bind as text with `::timestamp`.** These columns are `timestamp without time zone` holding UTC; the cast reads the instant as that same wall clock whatever the session's `TimeZone` is, rather than relying on how a `Date` parameter is inferred.

The unevaluable-row write collapsed the same way, for the same reason: every such row carries the same reason under the same claim stamp, so one `updateMany` says it for the batch instead of one per row. It stays wrapped, because a failed bookkeeping write must not cost the batch its good claims.

**The trade, stated plainly:**

- `lastClaimedAt` is no longer the claim mutex — `FOR UPDATE SKIP LOCKED` is. It remains the write-back CAS token, which is the half `completeAlertEvaluation` needs. Overlapping ticks pass over each other's rows instead of racing for them; the loser claims fewer than the budget, which is exactly what losing the CAS already cost it.
- A failed claim write now costs the tick's whole batch rather than one row. `runAlertTick` already catches this and skips the tick, and the next tick re-reads the same due set a minute later, so the failure mode is a skipped minute rather than lost rules.

## Validation

- 28 cases, 13 of them on the claim: the select (`ACTIVE`, due, live project), nulls sorting last, scan wider than the budget, the round-robin partition, the lock, one round-trip for the batch, a re-arm per window token, the fallback, and the batched unevaluable write surviving a rejected write.
- Mocking `$queryRaw` can prove the statement was built but not what Postgres does with it, so the generated SQL was **run against a local Postgres 17** on a fixture, in a rolled-back transaction:
  - **Round-robin holds in SQL.** With the budget cut to 3, `quiet-0` was claimed and `noisy-2` was not, though `quiet-0` sorted last of the five due rules — what `shareBudgetAcrossProjects` did.
  - **Cadence per row:** `1m → 10:38`, `5m`/`10m`/`2h → 10:42` (the cap), unreadable `24h → 10:38`, matching `alertNextRunAt` branch for branch.
  - A `PAUSED` rule, a not-yet-due rule and a rule on a soft-deleted project were all left untouched; a never-scheduled rule did not preempt the schedule.
  - `last_claimed_at` read back as `10:37:42.913` with no timezone shift, and the `RETURNING` aliases deserialize straight into `AlertRowLike` (`threshold` numeric, `filters`/`renotify` jsonb).

## Checklist

---
## Summary by cubic
Claims due alerts in one SQL statement instead of one `updateMany` per rule, cutting a full tick's 500 round-trips to one.

**Refactors**
- Replaces `findMany` plus per-row CAS with a single `UPDATE ... RETURNING`, with `FOR UPDATE SKIP LOCKED` as the mutex.
- Moves the round-robin budget share and per-window re-arm cadence (each rule re-arms on its own window, capped at five minutes) into the statement via `row_number()` and a `CASE`.

**Bug Fixes**
- Keeps nulls last in the scan, so a burst of new rules no longer preempts the schedule.
- Re-checks due-ness in the locking scan, so a row another tick claimed and committed between the statement's snapshot and its lock drops out instead of being claimed twice.

<sup>Written for commit 57f48e955402d0b031e6fcc1ac49610cae09affa. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/2081?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


GabrielDrapor:
@Prasad8830 reopened and retargeted to `main`, same as #2094/#2122/#2134: it was auto-closed on 2026-09-08 when `feat/alerts-ui-routes-pr` was deleted during the alerts stack merge (#1889); nothing from this branch reached `main`.

Against current `main` there are conflicts in `frontend/worker/src/alerts/claim.ts` and its test. Heads-up before you resolve them: 4158ccbd (in #1889) already replaced the `findMany` scan with a single raw window-function query that numbers rows per project and takes the cap depth-first, so part of what this PR set out to do is now on `main`. Worth re-reading the current `claimDueAlerts` first and keeping only the part this PR still adds (the per-row `updateMany` claim is still there). Approval then needs a code owner (@xinwei-traceroot).

Prasad8830:
@GabrielDrapor, merged `main` (via #2056's branch, which this PR is stacked on) and cut the PR down to the part it still adds.

**Kept from `main` (4158ccbd), unchanged in behaviour:** the per-project `row_number()` scan, depth-first dealing of the cap, and `NULLS LAST` so new rules wait behind due ones.

**What this PR now adds on top:** that scan, the `findMany` read-back and the per-row `updateMany` claim become one statement. `main`'s scan is the `due`/`picked` CTEs, then `locked` (`FOR UPDATE SKIP LOCKED`, re-checking due-ness at lock time) and `UPDATE ... RETURNING` with the per-window re-arm from #2056. 502 statements per tick become 1.

**Removed as a result:**
- `claimRow`, `shareBudgetAcrossProjects` and `ALERT_CLAIM_CONCURRENCY`: no per-row write left to deal or bound.
- `ALERT_CLAIM_SCAN_LIMIT`: it only gave the JS dealer headroom. `picked` takes the 500 straight off the depth ordering, and `due` has no cap of its own. My original branch capped a globally-ordered read *before* the numbering, which is the unfairness 4158ccbd fixed, so that is gone too.
- `main`'s read-back and JS-dealing tests, along with the code they covered. The scan assertions in `claim.test.ts` now check `main`'s ordering, plus that the budget is the only `LIMIT`.
- The `route.ts` comment that explained `MAX_ALERTS_PER_PROJECT` as `ALERT_CLAIM_SCAN_LIMIT / 10` now reads against `ALERT_CLAIM_LIMIT`. The value is unchanged.


@xinwei-traceroot ready for a look when you have a moment.

GabrielDrapor:
Reviewed the updated head `57f48e95`. No blocking issues found.

The single-statement claim preserves main's per-project depth-first selection and NULLS LAST ordering, and rechecks ACTIVE/due status in the locking scan. The cadence CASE is derived from #2056's helper, keeping the scheduling policy in one place.

Validation: all 123 worker alert tests passed. I also ran the actual generated SQL against PostgreSQL 17 using temporary tables in a rolled-back transaction: with 1,001 due rules in a noisy project, the 500-rule budget still included the quiet project's rules; paused, future-due, and soft-deleted-project rules were excluded; narrow, wide, and unreadable windows received the expected next-run timestamps. This fixture validates selection and re-arming; it is not a concurrent-session stress test.

Looks good from this review. Suggested merge order: #2056 first, then this PR.

GabrielDrapor (approved):

Issue #2006, opened 2026-08-26
Alerts: reorder the rule form as scope → condition → notify, and show the measure's unit on the threshold


GabrielDrapor:
<img width="417" height="761" alt="Image" src="https://github.com/user-attachments/assets/5e2647c8-18e2-4ad6-a1e9-9b3cd29b287f" />

## Summary

Two related rough edges in the alert form (New Alert / Edit Alert), found while walking through the feature:

1. **Section order breaks the reading flow.** The form is Metric (measure, aggregation, filter) → Conditions (trigger operator + value, window, renotify) → Notifications. The measure decides what the trigger value *means*, yet they sit in different sections; the filter decides *which spans* are measured, yet it sits between them. Reading order should follow the data: what to look at → what to compute and compare → whom to tell.

2. **The threshold has no unit.** The same number means different things per measure — latency is milliseconds, cost is USD, tokens are unitless — and the input gives no hint. Today the unit is not defined on the measure at all: `AlertMeasure` (`frontend/packages/core/src/alerts.ts`) has `id / label / type` only; the only unit knowledge is a hard-coded `latency → "ms"` in the Slack template (`frontend/packages/slack/src/alert-blocks.ts`, `MEASURE_UNIT_SUFFIX`) and a separate `FIELD_UNIT` in the dashboards filter controls. Easy to guess wrong (seconds vs ms).

_Screenshots of the current form to follow in a comment._

## Proposal

```
1  Scope       Filter — which spans (span_kind = LLM, model = gpt-4o, metadata[tenant] = acme …)
2  Condition   [p95] of [latency]   [>]  [ 500 ] ms   over the last [1m]
               one sentence: aggregation · measure · operator · value + unit · window
3  Notify      No-data mode (see #2004) · Renotify · Name · Slack
```

- Put the unit on the measure, once: `AlertMeasure.unit` (`{ suffix?: "ms" | "tok/s"; prefix?: "$" }`), and have the threshold input adornment, the Slack prose (`MEASURE_UNIT_SUFFIX`) and the preview chart axis all read it.
- The "one sentence" condition row is the shape Datadog / Grafana monitors use; it makes the measure–value relationship visible without a label.
- Folding the no-data mode into the Notify section resolves #2004 in the same pass.

## Where

- `frontend/ui/src/features/alerts/components/alert-form.tsx` — section order
- `frontend/ui/src/features/alerts/components/metric-section.tsx`, `condition-section.tsx` — regroup into scope / condition
[...]

Issue #2003, opened 2026-08-26
Alerts: a rule that breached before Slack was connected is never delivered until its state changes


GabrielDrapor:
## Summary

If an alert rule breaches while the workspace has no Slack channel (or no bot token), the delivery is recorded as `FAILED / no-channel` and **never retried** — not when Slack is connected afterwards, not on later ticks. The rule stays in `ALERT` with `lastNotifyStatus = FAILED`, and nothing is posted until the severity changes again (or the user pauses and resumes the rule).

"Create the rule first, connect Slack second" is the most natural onboarding order — the Alerts onboarding page itself shows both steps side by side — so users following it will believe the feature is broken.

## Reproduce (dev stack, `make dev`)

1. Workspace without a Slack integration; create a rule `count > 1`, window 1m.
2. Emit a few spans; wait for the tick. Rule → `ALERT`, `last_notify_status = FAILED`, `last_notify_error = no-channel`.
3. Connect Slack, pick a channel. Wait several ticks with traffic still flowing.
4. Nothing is posted. Pause + resume the rule → next tick pages as expected.

## Where

- `frontend/worker/src/notifications/alert-slack.ts` — `PERMANENT_DELIVERY_FAILURES` includes `no-channel`, `no-bot-token`, `bot-token-undecryptable`; `recordNonDelivery` records `FAILED` with no retry.
- `frontend/worker/src/alerts/scheduler.ts` / `claim.ts` — a tick only pages on a severity transition; `lastNotifyStatus` is never read.
- Slack routes (`api/workspaces/[workspaceId]/slack/*`, `api/slack/oauth/callback`) do not touch alerts.

## Options

- On Slack connect / channel save, cold-start (or re-page) the workspace's rules whose `lastNotifyStatus = FAILED` with a configuration reason.
- Or: treat `no-channel` / `no-bot-token` as *deferred* rather than permanent — leave the emission un-delivered and let the next tick retry while the rule is still in ALERT.
- At minimum: surface the failure in the alerts list (a "not delivered — connect Slack" state) so the user knows.

Found while verifying the alerts stack (#1889–#1907) end to end; see also #1995 and the e2e scenario "records a rule that breaches while the workspace has no Slack channel as FAILED / no-channel".

Prasad8830:
Hi @GabrielDrapor, @dark-sorceror, as I already worked on issues related to the alerts feature (#2025 to #2030), I would like to work on this. Could you please assign this to me?

[...]

PR #2122, opened 2026-09-05
feat(alerts): park a rule no run can ever evaluate instead of retrying it

### LINKED ISSUE #2029: Alerts: a PARKED status for permanently unevaluable rules
Follow-up from review of the alerts stack (#1889–#1907). Not a merge blocker; resource hygiene plus operator clarity.

## Problem

A rule whose spec can no longer be evaluated (e.g. its measure/field vocabulary drifted, or its project's data shape makes the query permanently invalid) has no terminal state. `claimDueAlerts` (`frontend/worker/src/alerts/claim.ts`) says so out loud in a comment: the row is claimed, sent for evaluation, fails the same way, and is re-read and discarded every minute, forever. The operator sees a perpetual "Failing" with no signal that no amount of waiting will fix it.

For comparison, Langfuse has `ERROR_BAD_QUERY` as a terminal monitor state.

## Suggested fix

Add a `PARKED` (or `ERROR_BAD_QUERY`) status the evaluator can park a rule into when the failure is a spec rejection (the evaluator already distinguishes `WidgetSpecError` — a broken rule — from transient query failure). Parked rules drop out of the claim query (`status = 'ACTIVE'` guard already exists), surface distinctly in the UI, and re-arm on edit — the same lifecycle a paused rule already has.

Prasad8830:
Fixes #2029

## Summary

A rule whose stored spec can no longer be evaluated has no terminal state. `claimDueAlerts` said so in a comment: the row is claimed, sent for evaluation, fails the same way, then is reread and discarded every minute, forever. The operator saw a perpetual "Failing" with no signal that any amount of waiting would fix it. This adds `PARKED`, a terminal status that the evaluator alone can reach when the failure is a verdict about the stored rule, not about the run.

## Type of Change

- [x] feat - A new feature

## Details

- `ALERT_STATUSES` gains `PARKED`, which leaves the claim query the way `PAUSED` already does (the `status = 'ACTIVE'` guard was already there, so no new query, index, or migration; `alerts.status` is a plain `VARCHAR`).
- A rule parks on three failures, all verdicts about the stored rule rather than the run: `parseAlertRule` returning null (parked in `claimRow`, after the claim, so `nextRunAt` still advances), a filter spec the backend would refuse over the whole batch (`isSendableAlertSpec`), and a backend `WidgetSpecError` (unknown measure, unsupported view, a forbidden aggregation, filters on a traces-routed measure).
- Telling a spec rejection apart from a transient query failure used to mean string-sniffing the message. `AlertEvaluationResult` now carries `error_kind` (`"spec"` or `"query"`), set on the two existing `except` branches in `alert_evaluation.py`; the worker parks only on `"spec"`. An unlabelled failure, from a backend older than this change, is retried and never parked, and a park write that itself fails falls back to recording the failure, so a rule is never silently stopped. A batch-level `WidgetSpecError` (a bad window) still raises for the whole request and stays a retry, since that's a request fault, not one rule's.
- `PARKED` is not client-settable: `ALERT_STATUSES` carries it, but the pause endpoint and the status mutation take `SETTABLE_ALERT_STATUSES` (`ACTIVE` or `PAUSED`) only.
- Leaving `PARKED` is a cold start, two ways: editing the rule (the PATCH route re-arms it when the edit rewrites the evaluated rule or writes a fresh `renotify`; a name-only edit leaves it parked, since it changes nothing the evaluator refused), or resuming from the row menu as the operator's explicit retry. If the settings are still unevaluable, the next tick parks it again with the reason.
- UI: a distinct "Parked" badge outranking Paused, dimmed row, menu offering Resume like a paused rule. The badge states the stored reason once and trusts it, since both real parking paths (`claim.ts`'s `UNEVALUABLE_RULE_ERROR`, `scheduler.ts`'s `PARKED_RULE_SUFFIX`) already say what unparks the rule; the badge does not append its own "Edit and save..." on top of that (that fallback sentence is kept only for the one case with no stored reason at all).

Before -
tr-before

After -
tr-after

## Checklist

- [ ] I have updated documentation where necessary. N/A, no doc-facing behaviour change

---
## Summary by cubic
Adds a terminal `PARKED` status for alert rules that can never be evaluated, replacing the perpetual retry that showed a frozen "Failing" state. The evaluator now parks a rule when the failure is a verdict about the stored spec (`error_kind: "spec"`), not the run (`"query"`).

- A rule parks when parsing fails, the spec is unsendable, or a `WidgetSpecError` rejects it; the reason is stored on the row.
- Unlabelled failures from older backends are still retried, and a failed park write falls back to recording the failure.
- `PARKED` is not client-settable: pausing a parked rule returns 409, and only Resume or an edit replacing the rejected rule (or its `renotify`) re-arms it; the re-arm is status-CAS'd, and renotify-only edits void the claim a delayed park would match.
- UI shows a "Parked" badge that outranks Paused, dims the row, and offers Resume for either stopped status; the badge trusts the stored reason and repeats the fix sentence only when no reason is stored.

**Migration**

- No schema or query changes; `alerts.status` is a plain `VARCHAR`.

<sup>Written for commit 35c03a3a6b6174c2a6bb719521f5157213fad27f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/2122?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


GabrielDrapor:
This PR was auto-closed on 2026-09-08 when its base branch `feat/alerts-ui-routes-pr` was deleted during the alerts stack merge (#1889). That was my doing while collapsing the stack; sorry for the noise. Nothing from this branch reached `main`. I restored it against `main` as #alerts-parked-status, Head ref must be a branch (createPullRequest) from a mirror of your branch (same commits). @Prasad8830 if you prefer to reopen from your fork instead, reply there and I'll close the mirror.

GabrielDrapor:
Restored as #2150 (mirror of your branch, against `main`). The earlier comment was missing the number.

GabrielDrapor:
@Prasad8830 this PR is open again and now targets `main` directly. It was auto-closed on 2026-09-08 when its base branch `feat/alerts-ui-routes-pr` was deleted while the alerts stack was collapsed into `main` (#1889); I restored that branch just long enough to reopen and retarget, so nothing of yours was lost. #2150, which I had opened as a mirror of this branch, is closed in favour of this one.

**Bringing it up to date.** There is one conflict against current `main`, in `frontend/worker/src/alerts/__tests__/claim.test.ts`: the scan tests were rewritten in 4158ccbd (the scan is now a raw window-function query, seeded in tests through a `due(rows)` helper instead of `findMany.mockResolvedValue`). The resolution is to keep your "park" test names and bodies and seed them with `due(...)`. A merge with that already resolved exists on `origin/restore/feat/alerts-parked-status` (merge commit `32f00a21`, CI green: worker 122 tests, ui 273, backend pytest 35). Either

- tick **Allow edits by maintainers** on this PR and I will push that merge here, or
- pull it in yourself: `git fetch origin restore/feat/alerts-parked-status && git merge 32f00a21`.

**Review.** Looks good to merge once updated. I checked that the park write uses the same CAS as the plain failure record, that only per-rule `WidgetSpecError`s (view/measure/filters/field vocabulary) are tagged `spec` while the batch-level `time_range` checks stay request faults, that an older backend without `error_kind` only retries, and that `alerts.status` is a plain VARCHAR so no migration is needed. One nit, not blocking: the fallback write in `/pause` has no status guard, so a client can move a `PARKED` rule to `PAUSED` (the UI never does). Guarding that fallback with `status: "ACTIVE"` would keep `PARKED` reachable only through evaluation.

Approval still has to come from a code owner (@xinwei-traceroot).

Prasad8830:
@GabrielDrapor, merged current `main` on top, which needed a bit more than the test conflict you flagged: #2056 and #2081 landed in between, so the per-row `claimRow` this PR parked from no longer exists.

**The park is now a batch write.** `main` collects the tick's unevaluable rows and writes them once, so parking follows it: one `updateMany` that sets `status: PARKED` plus the reason for the whole set, under the same CAS as before (`status: ACTIVE`, `lastClaimedAt: tick.now`). The fallback you reviewed is unchanged in spirit: if the park write throws, it falls back to recording the reason alone, so a park that does not land is never a silent success. `parkAlertRule` stays as it was for the evaluator path in `scheduler.ts`, and parked rules drop out of the claim for free, since the claim statement already filters `status = ACTIVE`. The park tests now assert the batch shape.

**On the `/pause` nit, I went slightly further than a plain `status: "ACTIVE"` guard.** Two cases made the blanket guard worse than the bug:

- Pausing an already-paused rule (a double-click) would stop matching and return 404. The fallback now allows the target status as a source too, so a repeat stays a 200 no-op: `["ACTIVE"]` on a resume, `["ACTIVE", "PAUSED"]` on a pause. `PARKED` is excluded either way, which is what you asked for.
- A refused pause on a parked rule would have reported "Alert not found" for a rule that plainly exists. On a zero-count write, the route now reads the row back: genuinely missing stays 404, parked returns 409, pointing at resume. 409 is already what this route file returns for the capacity conflict.

The UI never reaches either path, since a stopped rule offers Resume rather than Pause; this is about what a direct API client can do.

frontend/ui/src/app/api/projects/[projectId]/alerts/[alertId]/route.ts
```diff
@@ -102,13 +102,51 @@ export async function PATCH(req: NextRequest, { params }: RouteParams) {
...
+  let count = reArmsParked ? (await tryReArm()).count : 0;
+
   // Scoped write rather than a write on `id` alone: the project scope is the
-  // tenancy check, so it belongs on the statement that mutates.
-  const { count } = await prisma.alert.updateMany({ where: { id: alertId, projectId }, data });
+  // tenancy check, so it belongs on the statement that mutates. Falls back to
+  // it whenever the re-arm CAS above did not apply: the row was not actually
+  // PARKED at that check, so this edit's ordinary fields still have to land.
+  if (count !== 1) {
+    ({ count } = await prisma.alert.updateMany({ where: { id: alertId, projectId }, data }));
```
Comment (GabrielDrapor) [inline frontend/ui/src/app/api/projects/[projectId]/alerts/[alertId]/route.ts]: P2: A renotify-only repair can still be re-parked by an old claim. When the row is `ACTIVE`, the first `PARKED` CAS misses and this fallback updates `renotify` without clearing `lastClaimedAt`; after PATCH returns, a delayed `parkAlertRule(oldClaim)` still matches and sets it to `PARKED`. Please invalidate the old claim atomically and add a claim → PATCH → delayed-park regression test.

Comment (Prasad8830) [reply frontend/ui/src/app/api/projects/[projectId]/alerts/[alertId]/route.ts]: Thanks for the review. Now I have fixed it.

A rule rewrite was already safe, since `alertStateReset()` nulls `lastClaimedAt`. A renotify-only edit re-arms a parked rule without resetting state, so the fallback wrote `renotify` and left the old claim valid, exactly as you describe.

The fallback now voids the claim in the same write whenever the edit is one that re-arms a parked rule: `lastClaimedAt: null` goes into that `updateMany` alongside the edited fields, so there is no window between the repair landing and the claim going stale.

Prasad8830:
Hi @GabrielDrapor, fixed your review. Could you please recheck?

Also, some alert feature issues remain, including #2003 to #2006. I would like to work on them next. Could you please assign those to me?

GabrielDrapor (approved):
Thanks for addressing the stale-claim race and adding the regression coverage. Looks good!

Issue #1597, opened 2026-07-22
Pricing resolution is implemented twice (Python + TS) with no shared contract or catalog invariants — why the $0/mispriced-cost bugs keep recurring


rushikeshgoud19:
## Summary

Over the last few weeks the same bug has been filed and patched five separate times:

| Issue/PR | Symptom | Patch |
|---|---|---|
| #1558 | `google/gemini-3-flash` → `$0` (pattern required `-preview`) | #1560 — edit one regex |
| #1545 | `gpt-5.1-codex-mini` → `null` (6/7 codex SKUs unpriced) | #1557 — add 5 JSON entries |
| #1556 | `vertex_ai/`, `openrouter/`, `litellm/`, `azure/` prefixes → `$0` | #1566 — strip prefixes |

Each fix is correct and each is scoped to the symptom. But they're all the same root cause, so the class keeps recurring: **price resolution is implemented twice, independently, with no shared contract and no test that the shipped catalog actually resolves.**

I don't think this needs a big rewrite — it needs one resolution contract plus a CI guard. Writing it up because the next model launch will otherwise reproduce it a sixth time.

## Root causes

**1. Two independent resolvers that are already diverging.**

- Python: `get_model_price()` — `backend/worker/tokens/pricing.py`
- TypeScript: `getModelPricing()` — `frontend/packages/core/src/model-pricing/lookup.ts`

Both implement "exact match on `model_name`, then first regex `matchPattern` that matches." They share only the `standard-model-prices.json` **data**, never the **logic**.

The divergence is concrete right now: **#1566 adds gateway-prefix stripping to the Python resolver only.** `lookup.ts` is untouched by any open PR. The moment #1566 merges, the two engines return different costs for the same model:

```
vertex_ai/gemini-2.5-pro
  Python  get_model_price()  →  gemini-2.5-pro rates   ✅  (after #1566)
  TS      getModelPricing()  →  null → calculateCost() returns 0   ❌
```

This is not a theoretical path. The TS resolver has exactly two callers, and both are **billable inference**:

- `frontend/worker/src/processors/detector-run-processor.ts:148` — `detectorInferenceCost()`, the local pricing **fallback** used whenever the provider didn't report a cost. An unpriced model records the detector run at **$0**.
- `frontend/packages/agent/src/index.ts:267` — agent/RCA cost accounting.

[...]

Issue #1545, opened 2026-07-12
Cost reads `null` for gpt-5.1-codex-mini — missing entry in standard-model-prices.json (6 of 7 codex SKUs unpriced)


lqiu03:
### Environment

Chrome, MacOS

### Steps to Reproduce

Trace any agent framework on OpenAI Codex 5.1-mini, no cost renders in tracing spans:

<img width="1121" height="358" alt="Image" src="https://github.com/user-attachments/assets/ef7af808-2c3c-4dc9-93ed-12fd98cd5855" />

<img width="1157" height="377" alt="Image" src="https://github.com/user-attachments/assets/2c4385d6-cbd5-4df3-b960-f138f2870f36" />

### Expected Behavior

Render cost per model used in dashboard.

### Suggested fix
Add the missing entries. For example for`gpt-5.1-codex-mini` per official OpenAI model API pricing documentation.

### Actual Behavior

See e2e screenshots, no cost renders for GPT Codex5.1-mini.

### Root Cause:
`backend/worker/otel_transform.py:596` — `cost_from_buckets(get_model_price(model_name), buckets)` — `get_model_price` matches against `frontend/packages/core/src/standard-model-prices.json`, which has 87 entries total.

Grepped for every `codex` entry: exactly one exists —

```json
"id": "tr-gpt-5-3-codex",
"modelName": "gpt-5.3-codex",
"matchPattern": "(?i)^(openai\\/)?(gpt-5\\.3-codex)(-[\\d-]+)?$",
```

`gpt-5.1-codex-mini` has no entry and matches no pattern. `get_model_price` returns `None` → `cost_from_buckets` returns `None` → stored as `null`.

Broader pattern, not just this one model: checked which codex SKUs actually exist upstream (`@earendil-works/pi-ai`'s own model table) against what is priced here —

| Model | Priced in `standard-model-prices.json`? |
|---|---|
| `gpt-5-codex` | no |
| `gpt-5.1-codex` | no |
| `gpt-5.1-codex-max` | no |
| `gpt-5.1-codex-mini` | no |
| `gpt-5.2-codex` | no |
| `gpt-5.3-codex` | yes |
| `gpt-5.3-codex-spark` | no (`matchPattern` excludes `-spark`; it only allows trailing `-[\d-]+`) |

6 of 7 known codex variants are unpriced, and the one present entry's regex would not even catch its own `-spark` sibling.

### Relevant Logs/Console Output

```shell

```

### TraceRoot Version

[...]

PR #2125, opened 2026-09-05
fix(pricing): normalize gateway prefixes once instead of per catalogue entry

### LINKED ISSUE #1556: Pricing silently records $0 for gateway/router-prefixed model names (vertex_ai/, openrouter/, litellm/, inconsistent azure/) — prefix handling is hand-encoded per entry

Mohil-Ahuja:
## What

A model id arriving with an API-gateway / router prefix silently resolved to `$0`. Every catalogue pattern hand-encodes which prefixes it tolerates, so coverage drifts between siblings (`gpt-5.6-sol` accepts `azure/`, `gpt-5.4` does not) and no entry accepts the router prefixes real deployments emit.

This takes option (1) from the issue — normalize once in the lookup — rather than patching individual entries.

## Before / after

Run against the real `standard-model-prices.json`, exact-then-regex, same logic as the worker:

| model string | before | after |
|---|---|---|
| `vertex_ai/gemini-2.5-pro` | `$0` (no match) | `gemini-2.5-pro` |
| `azure/gpt-5.4` | `$0` (no match) | `gpt-5.4` |
| `openrouter/anthropic/claude-opus-4-8` | `$0` (no match) | `claude-opus-4-8` |
| `litellm/openai/gpt-5` | `$0` (no match) | `gpt-5` |
| `portkey/openai/gpt-4o` | `$0` (no match) | `gpt-4o` |
| `bedrock/us.anthropic.claude-opus-4-8` | `$0` (no match) | `claude-opus-4-8` |
| `openai/gpt-5.6-sol` | `gpt-5.6-sol` | `gpt-5.6-sol` |
| `anthropic/claude-opus-4-8` | `claude-opus-4-8` | `claude-opus-4-8` |
| `google/gemini-2.5-pro` | `gemini-2.5-pro` | `gemini-2.5-pro` |
| `openrouter/acme/not-a-real-model` | `$0` (no match) | `$0` (no match) |

## Design

- **The id is matched as given first.** Only when nothing matches are gateway prefixes stripped and the exact/regex passes retried. The fallback is strictly additive: it can turn a `None` into a price, but can never redirect a price that already resolved — so no existing row's behaviour changes.
- **Only known segments are stripped.** `my-org/gpt-5` and `ft:gpt-4o:acme/custom` are untouched, and an unknown SKU behind a gateway stays unpriced rather than being mapped to something plausible-looking. Bedrock's `us.anthropic.…` and Vertex's `model@date` are distinct id shapes, not slash prefixes, and pass through unchanged.
- **Chained prefixes** are handled, bounded at a depth of 3 (the wild maximum is 2).
- The JSON patterns are deliberately **not** touched in this PR. They can drop their baked-in `(provider\/)?` alternations later; doing it here would mix a behaviour fix with an 87-entry rewrite.

## The TypeScript twin

Pricing resolution is implemented twice (#1597) and `frontend/packages/core/src/model-pricing/lookup.ts` — which prices agent and chat calls — had the identical gap, so a Python-only fix would leave those surfaces reading `$0`. Both are fixed the same way.

Neither side can import the other, and a prefix added to one list only would reinstate this bug on the half that was missed: the model prices correctly in one surface and reads `$0` in the other, which looks like flaky cost rendering rather than a lookup gap. `tests/worker/tokens/test_gateway_prefix_parity.py` parses `lookup.ts` and fails when the lists drift — following the convention in `tests/rest/test_detector_trigger_field_parity.py`, including the "assert what you found" guard so a moved declaration fails loudly instead of comparing two empty sets. I verified it catches drift by adding a prefix on one side only.

## Tests

- `tests/worker/tokens/test_pricing.py` — the issue's repro table; ids that already resolved are unchanged; an unknown model behind a gateway stays `None`; non-gateway segments are left alone; stripping terminates on a pathological id.
- `tests/worker/tokens/test_gateway_prefix_parity.py` — the drift guard (3 tests).
- `frontend/packages/core/src/__tests__/model-pricing.test.ts` — the same matrix against a catalogue-shaped fixture, plus `stripGatewayPrefixes` unit cases.

## Validation

- `uv run --frozen pytest tests/worker` → **613 passed** (277 in `tests/worker/tokens`).
- `vitest run src/__tests__/model-pricing.test.ts` in `packages/core` → **33 passed**.
- `ruff check` / `ruff format --check`, `eslint`, `prettier --check` all clean on the touched files.
- Three test files in `packages/core` fail to load in a bare checkout (unbuilt Prisma client); they fail identically on a clean `main`.

## Note on scope

Genuinely-missing SKUs (`gemini-flash-3` per #1558, the codex family per #1545) are a different class and are not touched here, per the scope note on the issue.

Two contributors asked to be assigned this in July and no PR followed; happy to step aside if either is still working on it.

Closes #1556

---
## Summary by cubic
Normalizes gateway/router prefixes once in the pricing lookup and the token estimator, so prefixed model IDs like `vertex_ai/gemini-2.5-pro`, `azure/gpt-5.4`, and `openrouter/anthropic/claude-opus-4-8` now price as their bare models instead of `$0`, and estimate tokens with the correct encoder.

- Strips known gateway segments only after a direct match fails, so existing matches are untouched and unknown models behind a gateway stay unpriced.
- Handles chained prefixes up to depth 3 and carries both spellings of each vendor (`z-ai`/`zai`, `moonshotai`/`moonshot`, etc.).
- Replaces first-match-wins with a specificity ranking so overlapping patterns resolve the same way on both runtimes, and orders the catalogue read deterministically to match the worker.
- `is_claude_model` and the tiktoken lookup operate on the stripped id; Bedrock-shaped dotted Claude ids are recognized only when `anthropic` qualifies them.
- Adds a parity test that fails if the Python and TypeScript prefix lists drift, plus end-to-end tests asserting prefixed ids cost the same as their bare forms.

<sup>Written for commit ec2779a3781c85554ebf5ccd9663c4b51afa1b8a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/2125?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Mohil-Ahuja:
Both valid — fixed in ba575e64.

1. Token estimation: is_claude_model was a bare startswith("claude"), so
   openrouter/anthropic/claude-opus-4-8 answered False and count_tokens fell
   through to cl100k_base. This PR made those ids resolve a price for the first
   time, so it was attaching a correct cost to a wrong token count. Prefixed
   OpenAI ids had the same problem — "azure/gpt-4o" isn't a name tiktoken knows.

1. Token estimation: is_claude_model was a bare startswith("claude"), so
   openrouter/anthropic/claude-opus-4-8 answered False and count_tokens fell
   through to cl100k_base. This PR made those ids resolve a price for the first
   time, so it was attaching a correct cost to a wrong token count. Prefixed
   OpenAI ids had the same problem — "azure/gpt-4o" isn't a name tiktoken

   GATEWAY_PREFIXES and strip_gateway_prefixes therefore moved to types.p
   beside is_claude_model: a prefixed id has to mean the same model to every
   reader of it, not just the price lookup. usage.py also can't import pr
   without a cycle, which independently forces that placement.

2. Parity regex: correct, and it was the exact hole the file exists to close —
   [a-z0-9_]+ dropped any hyphen/dot/capital prefix from the parse, so a
   one-sided "vertex-ai" would have read as agreement, and the lowercase
   assertion only ever saw entries the parse had already accepted. It now
   every quoted token in the declaration, with a test asserting a non-plain
   prefix is actually seen.

Both new test classes fail against the previous behaviour (verified by re
each fix). On the tiktoken case my first assertion would have passed either way —
"hello world" tokenizes identically under both encodings — so it now uses text
where o200k_base gives 50 and cl100k_base 54.

622 worker tests pass; ruff clean.

backend/worker/tokens/pricing.py
```diff
@@ -105,6 +161,30 @@ def get_model_price(model: str) -> dict[str, float] | None:
...
+    """
+    cache = _load_cache()
+
+    prices = _match(cache, model)
+    if prices is not None:
+        return prices
+
+    bare = strip_gateway_prefixes(model)
+    if bare != model:
+        return _match(cache, bare)
```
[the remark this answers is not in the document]
dark-sorceror: Resolved in 90fccf6e1, covered by b9cdde132.

Valid, and it was a regression this PR introduced: before, a gateway-prefixed Claude id priced at $0 so no cost was computed; with the price fallback it resolves, and `is_claude_model` was still handed the raw prefixed id, so token counting fell to `cl100k_base`.

Fixed at `is_claude_model` rather than at the call site, so a prefixed id means the same model to every reader — which is the reason the prefix set lives in `types.py` in the first place. It also needed more than prefix stripping: Bedrock and its regional aliases qualify the family with dots (`us.anthropic.claude-opus-4-8`), so the name is not the leading segment and a bare `startswith` missed it. The catalogue prices that shape, so the estimator now recognises it, matching only on a segment boundary — `my-org/claude-ish` stays a non-match.

Verified: 297 worker token tests pass, and a probe confirms `anthropic/`, `openrouter/anthropic/`, `bedrock/us.anthropic.` and bare `us.anthropic.` forms all resolve to the Claude estimator. One limit worth stating: a Claude id behind a gateway prefix that is not in the allowlist still gets the wrong estimator, which is the same tradeoff as the price lookup — a new gateway is one line.

dark-sorceror (approved):
Conflicts resolved and the open finding fixed and resolved above.

Verified independently rather than on inspection: 297 worker token tests pass on the merged tree, and a behaviour probe confirms `anthropic/`, `openrouter/anthropic/`, `bedrock/us.anthropic.` and bare `us.anthropic.` forms all reach the Claude estimator, while `my-org/claude-ish`, `gpt-3.5-turbo` and `azure/gpt-5.4` correctly do not — so the segment-boundary claim in the docstring holds rather than just being asserted.

The design choice that makes this safe is that the fallback is additive: the id is matched as given first, so any pattern deliberately recognising a prefixed form keeps winning, and stripping only runs on a miss. It can turn a `None` into a price but cannot change a price that already resolved.

Two limits worth recording rather than hiding. A Claude id behind a gateway prefix outside the allowlist still gets the wrong estimator — the same tradeoff as the price lookup, and a new gateway is one line. And the Python and TypeScript prefix lists are kept in step by a parity test rather than a shared source, so the test is the only thing standing between them and drift.

backend/worker/tokens/types.py
```diff
@@ -1,6 +1,87 @@
...
+    estimate turns on which one is picked — and the price fallback now attaches a
+    real cost to whichever answer comes back.
+
+    Bedrock and its regional aliases qualify the family name with dots rather than
+    slashes (``us.anthropic.claude-opus-4-8``), so the name is not always the
+    leading segment and a bare ``startswith`` misses it. The catalogue prices that
+    shape, so the estimator has to recognise it too. Only a segment boundary
+    counts, which keeps ``my-org/claude-ish`` a non-match.
+    """
+    return any(segment.startswith("claude") for segment in strip_gateway_prefixes(model).split("."))
```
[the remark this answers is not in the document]
dark-sorceror: Resolved in 9b237feba, pinned by a5a3030c0.

Valid, and confirmed by probe: `my-org.claude-proxy` and `acme.claude-router` classified as Claude while pricing left them unmatched, so the Claude estimator was being used for ids the catalogue does not recognise. Before this PR both were False, so it was a regression introduced here.

The dotted form now also requires an `anthropic` segment, matching the qualifier the catalogue Bedrock patterns require, so classification is no wider than what pricing will match. Every legitimate form still resolves — bare, `anthropic/`, `openrouter/anthropic/`, `azure/`, `us.anthropic.` and `bedrock/us.anthropic.` all classify and price.

Your point about the docstring was the useful part: its `my-org/claude-ish` example is slash-shaped, so it held while the dot-shaped case it implied did not. The guard test had the same blind spot — its cases were `bedrock/amazon.nova-pro-v1:0`, `us.meta.llama3-70b` and that same slash example, none of them an `X.claude-Y` id, so it passed without covering the shape it claimed to. Both dotted names are now cases in it, and they fail against the old predicate.

tests/worker/tokens/test_pricing.py
```diff
@@ -269,6 +272,172 @@ def test_input_and_output_match_published_rate(self, real_cache):
...
+    @pytest.mark.parametrize(
+        "model_id",
+        ["bedrock/amazon.nova-pro-v1:0", "us.meta.llama3-70b", "my-org/claude-ish"],
+    )
+    def test_dotted_non_claude_ids_stay_non_claude(self, model_id):
+        """Only a segment boundary counts, so the dot check cannot widen into a
+        substring match on unrelated vendors."""
+        assert is_claude_model(model_id) is False
+
+    def test_bedrock_prefixed_claude_costs_the_same_as_the_bare_id(self, real_cache):
```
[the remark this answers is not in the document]
dark-sorceror: Resolved in a5a3030c0.

Folded into one parametrized `test_prefixed_claude_costs_the_same_as_the_bare_id` over `azure/`, `openrouter/anthropic/` and `bedrock/us.anthropic.` — there were three variants of the same assertion, not two. Kept the fuller docstring explaining why the fallback makes this reachable, and added a line noting the slash and dot-qualified forms reach the estimator through different branches, which is the reason to keep more than one case rather than collapse to a single id.

backend/worker/tokens/types.py
```diff
@@ -1,6 +1,95 @@
...
+#
+# This lives beside is_claude_model rather than in pricing.py because a prefixed
+# id has to mean the same model to *every* reader of it — the price lookup and the
+# token estimator both — and usage.py cannot import pricing.py without a cycle.
+#
+# Keep in sync with GATEWAY_PREFIXES in
+# frontend/packages/core/src/model-pricing/lookup.ts — the two lookups must agree
+# on what a model id means. tests/worker/tokens/test_gateway_prefix_parity.py
+# fails if they drift.
+GATEWAY_PREFIXES = frozenset(
```
dark-sorceror, line 19: Add `gemini`, `litellm_proxy` and `bedrock_converse`, plus OpenRouter's vendor spellings `z-ai`, `moonshotai`, `x-ai`, `mistralai` alongside the LiteLLM ones. Without the second group `openrouter/z-ai/glm-4.6` prices at $0 while `zai/glm-4.6` works.

frontend/packages/core/src/model-pricing/lookup.ts
```diff
@@ -84,13 +84,68 @@ async function loadCache(): Promise<CachedModel[]> {
...
- * Look up pricing for a model by name.
- * Tries exact match on modelName first, then regex matchPattern fallback.
- * Returns prices in USD per token, or null if not found.
+ * Gateway / router prefixes stripped before the second matching pass.
+ *
+ * Every catalogue pattern hand-encodes which prefixes it tolerates, so coverage
+ * drifts between siblings and no entry accepts the router prefixes real
+ * deployments emit. Normalizing once here fixes every row at the same time.
+ *
+ * Keep in sync with GATEWAY_PREFIXES in backend/worker/tokens/pricing.py — the two
```
dark-sorceror: `GATEWAY_PREFIXES` is in `types.py`, not `pricing.py`.

backend/worker/tokens/types.py
```diff
@@ -1,6 +1,95 @@
...
+
+
+def strip_gateway_prefixes(model: str) -> str:
+    """Drop leading gateway/router segments from a model id.
+
+    ``openrouter/anthropic/claude-opus-4-8`` -> ``claude-opus-4-8``.
+
+    Only segments in GATEWAY_PREFIXES are removed, so an id whose first segment is
+    part of the model's real name is returned untouched. Bedrock's
+    ``us.anthropic.claude-…`` and Vertex's ``model@date`` forms are distinct id
```
dark-sorceror, line 66: 4 of the 19 Claude entries have no `@date` alternation, so the catalogue does not already handle that form.

dark-sorceror (changes requested):
Superseding my earlier approval. All four items reproduce, measured against the real catalogue:

```
gemini/gemini-2.5-pro                    None      google/ and bare both work
litellm_proxy/openai/gpt-5               None      litellm/ works
bedrock_converse/us.anthropic.claude-.   None      bedrock/ works
openrouter/z-ai/glm-4.6                  None      zai/glm-4.6 works
openrouter/moonshotai/kimi-k3            None      moonshot/kimi-k3 works
```

The chained case is the sharper one, because the tests pass only where OpenRouter's vendor slug happens to coincide with the LiteLLM spelling (`anthropic`, `google`, `openai`). Every slug where the two differ is still unpriced.

**`lookup.ts:61` needs `orderBy: { modelName: "asc" }`.** It is outside this diff so I cannot anchor a comment there. Python selects with `ORDER BY m.model_name, p.usage_type` while `findMany` has no ordering, so wherever two patterns match one id the two lookups can return different entries, and the TypeScript answer can shift after an UPDATE. The parity test does not cover this because it compares prefix lists, not resolution outcomes.

That matters more than tidiness, because two patterns matching one id is not hypothetical. Under `ORDER BY model_name`, `claude-sonnet-4` sorts ahead of `claude-sonnet-4-6` and both patterns match, so `anthropic/claude-sonnet-4-6` resolves to `claude-sonnet-4` while `openrouter/anthropic/claude-sonnet-4-6` resolves correctly. Same rates today, so nothing mis-bills yet, but it breaks the day 4.6 is repriced.

Also worth deciding deliberately rather than discovering later: `trace_reader.py` recomputes the breakdown live through `get_model_price`, while `cost` is written at ingest. Once prefixed ids resolve, an old prefixed span gets a populated breakdown against a NULL stored cost, so the hover total will disagree with the chip for pre-deploy traces. It self-heals for new spans.

Mohil-Ahuja:
All four fixed in e7aee282, and the ordering point was the one I most needed pointed out.

**Gateway self-spellings.** `gemini`, `litellm_proxy` and `bedrock_converse` added. Each already had its sibling in the set, so one spelling priced and the one next to it read $0.

**Router vendor slugs.** `z-ai`, `moonshotai`, `x-ai` and `mistralai` added alongside the LiteLLM spellings. Your framing of why the tests missed it is the part worth keeping: the chained cases I wrote all used vendors where OpenRouter and LiteLLM happen to agree, so the passing tests were evidence of a coincidence rather than of coverage. There is now a test asserting both spellings of each pair are carried, so adding one without the other fails. Only `zai` and `moonshot` have a catalogue entry to resolve against today; the other two are asserted at the set rather than through a price.

**Catalogue ordering.** `orderBy: { modelName: "asc" }` added, and you are right that it is not tidiness. Two tests: one asserts the query asks for the order, the other resolves `anthropic/claude-sonnet-4-6` over the overlapping pair and pins it to `claude-sonnet-4`, which is what the worker answers. The first is the regression guard, since the ordering is the database's job and a fixture cannot demonstrate it.

**The @date claim.** Corrected. 15 of the 19 Claude entries spell out an `@` alternation and 4 do not, so the docstring was overclaiming. It now says which entries handle it and notes that closing the rest belongs with the catalogue patterns, since it is a suffix rather than a prefix.

309 worker token tests and 35 core pricing tests pass. Nine of them fail against the previous prefix set, verified by reverting it.

On `trace_reader.py`: agreed it should be decided rather than discovered. An old prefixed span will show a populated hover breakdown against a NULL stored cost, and it self-heals for new spans. My read is that it is worth leaving as is, since the alternative is either a backfill or teaching the reader to suppress a breakdown it can compute, and both are more machinery than a cosmetic disagreement on pre-deploy traces deserves. Happy to do it either way, and to put that note in the PR body so it is not rediscovered later.

Mohil-Ahuja:
Fixed the red check in 94fc2fce — the TypeScript ordering test was mine and it did not survive the ranking change.

It still asserted first-match-wins, so it was pinning `anthropic/claude-sonnet-4-6` to `claude-sonnet-4`, which is the behaviour 9b795104 deliberately replaced. The test was guarding the thing that got fixed.

Rewritten as the mirror of `TestResolutionIsOrderIndependent`: the successor resolves to its own entry from both row orders, and the predecessor still resolves to itself, so ranking cannot drag every id onto the longest pattern that happens to match it. The predecessor-first case is the one that fails under first-match-wins, verified by reverting the ranking; the successor-first order is the one the catalogue file holds, which is exactly the order that hid the bug.

Ranking is the better fix than the ordering I proposed. Mine made the two sides agree on a wrong answer; yours makes the answer right, which is why the `claude-sonnet-4-6` case now reads as its own entry rather than its predecessor. Worth noting the ordering is still load-bearing underneath it for the tie case, since the name breaks ties and two entries of equal length would otherwise depend on arrival order.

37 core pricing tests pass.

backend/worker/tokens/types.py
dark-sorceror: Fixed in e7aee282. All seven present, and your pairing test is a better guard than the resolution cases alone, since it catches the vendors whose two spellings coincide today.

frontend/packages/core/src/model-pricing/lookup.ts
dark-sorceror: Fixed in e7aee282.

backend/worker/tokens/types.py
dark-sorceror: Fixed in e7aee282, and your version is more precise than my note. Saying which 15 entries carry the alternation, and that closing the gap belongs with the catalogue patterns rather than the prefix list, is the part I had left vague.

dark-sorceror (approved):
Superseding my changes-requested. All four items are fixed, and the ranking discussion landed somewhere better than where I started it.

Verified on the merged tree: 19 checks green, 314 worker token tests pass, and across 644 probes of every catalogue name under six prefix forms exactly two resolutions change, both of them the subsumption being fixed. No resolution becomes less specific.

Two corrections to things I wrote, since they matter for whoever reads this later. My ordering comment was wrong and caused the preview-model fallback, and your original placement was deliberate and correct. And my code comment called the `orderBy` "belt and braces rather than the fix", which is not right: ties are broken by name, so two entries of equal length would still depend on arrival order without it. Your framing, that the ordering stays load-bearing underneath the ranking, is the accurate one.

Your test rewrite is the part that makes this safe. The old one pinned first-match-wins, so it was guarding the behaviour the change replaces; asserting the successor resolves to its own entry from both row orders, and that the predecessor still resolves to itself, is what stops ranking dragging every id onto the longest pattern that happens to match.

Two standing gaps recorded rather than blocking. A Claude id behind a gateway prefix outside the allowlist still reaches the wrong estimator, which is the same tradeoff as the price lookup and one line to extend. And the two prefix lists are kept in step by a parity test rather than a shared source, so that test is the only thing between them and drift.

dark-sorceror:
resolve merge conflicts

Mohil-Ahuja:
Conflicts resolved in ec2779a3.

Only one file conflicted, `tests/worker/tokens/test_pricing.py`, where main added `TestGpt6AstraPublishedPrices` at the same point this branch added `TestGatewayPrefixes`. Purely additive, so both are kept, with the published-price class next to its siblings and the prefix class after it. No source file conflicted.

374 worker token tests pass on the merged tree, up from 314 because main brought its own.

Since main added catalogue entries after your review, I checked them against what this branch governs rather than assuming. Every one of `gpt-5.6-sol`, `gpt-5.6-cyber`, `gpt-6-astra`, the six codex SKUs, `kimi-k3`, `kimi-k2.6`, `grok-4` and `grok-4.20` prices bare, prices under all of `openai/`, `azure/`, `litellm/`, `openrouter/`, `bedrock/`, `vertex_ai/`, `litellm_proxy/openai/` and `portkey/`, and resolves to its own entry rather than a shorter sibling. `kimi-k2.6` is the interesting one of those, since `kimi-k2` would have subsumed it under first-match-wins; the ranking keeps it on its own entry.

Both standing gaps you recorded are still open and still out of scope here.

PR #2135, opened 2026-09-07
chore(deps): refresh frontend dependency overrides, bump next and postcss

GabrielDrapor:
Dependency maintenance in `frontend/`.

- `next` 16.2.6 → 16.2.11, `postcss` → ^8.5.28 (ui).
- `pnpm.overrides`: transitive packages lifted to fixed releases, each keyed to the major it already resolves to. Most moves are patch-level; the exceptions are `sharp` 0.34 → 0.35 and `esbuild` 0.27 → 0.28 (0.x minor boundaries) and minor bumps of `vite` (8.0 → 8.2) and `hono` (4.12 → 4.13).
- The stable OpenTelemetry 2.x family is pinned to a single minor (2.11.0) so `core`, `resources`, `sdk-trace-*`, `sdk-metrics`, propagators and the zipkin exporter never pair across minors. The 1.x line used by the TraceRoot SDK is untouched.
- Lockfile regenerated; `pnpm install --frozen-lockfile` is consistent.

Verified locally: all workspace packages build, `next build` succeeds, eslint has no errors, every package's vitest suite passes on Node 22 (the ui suite's Node 26 `localStorage` failures reproduce identically on `main`). Two extra checks for the reviewer notes: a span-export smoke test through the 2.11 SDK reaches a local OTLP endpoint with both the JSON and protobuf exporters, and `sharp` 0.35 resizes and encodes webp/avif.

Three advisories stay open on purpose: `@opentelemetry/core` / `propagator-jaeger` 1.x (fixes only exist in 2.x; that migration is its own change) and `uuid` 8 via `node-cron` (fix only in 11).


XinweiHe:
Override list looks well-targeted — spot-checked a sample and each maps to a real upstream advisory rather than being a blanket sweep.

**One thing I'd like fixed before merge.** The `@opentelemetry/core@>=2.0.0 <2.8.0 → ^2.8.0` override hoists `core` to 2.11.0 but leaves its siblings pinned. Cross-minor pairings in the tree go from **0 → 13**:

```
sdk-trace-base@2.2.0 / 2.7.1               -> core 2.11.0
resources@2.2.0 / 2.6.0 / 2.7.1            -> core 2.11.0
sdk-metrics@2.2.0 / 2.7.1                  -> core 2.11.0
exporter-zipkin@2.7.1, propagator-b3@2.7.1 -> core 2.11.0
```

The stable OTel SDK packages ship in lockstep and aren't guaranteed compatible across minors internally. Green CI isn't much evidence here — build and unit tests never exercise span export, so a broken exporter would pass straight through. Given what we ship, I'd rather not find out in prod.

Widening the override should do it — scoped `<2.11.0` so the 1.x line stays untouched:

```json
"@opentelemetry/resources@>=2.0.0 <2.11.0": "^2.11.0",
"@opentelemetry/sdk-trace-base@>=2.0.0 <2.11.0": "^2.11.0",
"@opentelemetry/sdk-metrics@>=2.0.0 <2.11.0": "^2.11.0",
"@opentelemetry/exporter-zipkin@>=2.0.0 <2.11.0": "^2.11.0",
"@opentelemetry/propagator-b3@>=2.0.0 <2.11.0": "^2.11.0",
```

Derived from lockfile analysis, not from running the install — worth re-resolving and re-checking the pairings after, plus a smoke test that traces still export end-to-end.

**Non-blocking notes:**

- "current patch releases / nothing crosses a major" is a little generous: `sharp` 0.34.5 → 0.35.4 and `esbuild` 0.27.x → 0.28.2 cross the 0.x breaking boundary; `vite` 8.0.0 → 8.2.2 and `hono` 4.12.25 → 4.13.7 are minors. `sharp` is Next's image optimizer and `next/image` is a runtime path nothing here covers — one manual check would close it out.
- `axios` 1.15.2 → 1.20.0 pulls in `https-proxy-agent@5.0.1` → `agent-base@6.0.2`, two deprecated majors newly in the tree.
- The blanket `protobufjs` override became a scoped 8.x key, so the 7.x line is no longer rewritten. Fine today (resolves 7.6.6) — just noting the guard changed shape.
- `qs@>=2.2.5 <6.16.0` and `brace-expansion@>=3.0.0 <5.0.9` span majors and would silently jump a future 3.x/4.x dep. Nothing hits them today.

GabrielDrapor:
Good catch on the OTel pairings. Pushed a follow-up:

- The whole stable 2.x family (`core`, `context-async-hooks`, `resources`, `sdk-trace`, `sdk-trace-base`, `sdk-trace-node`, `sdk-metrics`, `exporter-zipkin`, `propagator-b3`, `propagator-jaeger`) is now keyed `>=2.0.0 <2.11.0 → ^2.11.0`. Re-resolved: every 2.x package in the lockfile is 2.11.0 and all 2.x → `core` edges point at 2.11.0, so cross-minor pairings are back to 0. The 1.x line is untouched.
- Smoke test: a `NodeTracerProvider` on 2.11.0 with `SimpleSpanProcessor` exporting to a local HTTP stub, once through `exporter-trace-otlp-http` (JSON) and once through `exporter-trace-otlp-proto`. Both requests arrive with the span name, attributes and `service.name` intact.
- `sharp` 0.35.4: resize + webp + avif encode run fine; worker and agent rebuilt, worker suite green.
- `qs` and `brace-expansion` keys narrowed to the majors actually in the tree (`qs@>=6.0.0`, `brace-expansion@>=5.0.0`), so nothing can jump a future 3.x/4.x dep.
- Description reworded: sharp/esbuild cross a 0.x boundary and vite/hono are minors; the protobufjs guard change is intentional (7.x already resolves above the advisory floor).

The `https-proxy-agent@5` / `agent-base@6` pull-in from axios 1.20 is noted; left as is since it's axios's own dependency choice.

XinweiHe (approved):

PR #2150, opened 2026-09-09
feat(alerts): park a rule no run can ever evaluate instead of retrying it

### LINKED ISSUE #2029: Alerts: a PARKED status for permanently unevaluable rules

GabrielDrapor:
> **Restored against `main` on 2026-09-09.** This is #2122 by @Prasad8830, which GitHub auto-closed on 2026-09-08 09:31 UTC when its base branch `feat/alerts-ui-routes-pr` was deleted while the alerts stack was collapsed and merged into `main` (#1889, `af6c414d`). None of its commits reached `main`. The head here is a byte-for-byte mirror of `Prasad8830:feat/alerts-parked-status` (same commits, same authorship) pushed as `restore/feat/alerts-parked-status`, because a PR cannot be opened from another user's fork. Against current `main` it one conflict to resolve in `frontend/worker/src/alerts/__tests__/claim.test.ts` (the scan tests were rewritten in 4158ccbd). @Prasad8830, if you would rather reopen from your fork, say so and this one gets closed. Original description follows.

---

Fixes #2029
> **Base branch:** this targets `feat/alerts-ui-routes-pr` (#1907), not `main`

## Summary

A rule whose stored spec can no longer be evaluated has no terminal state. `claimDueAlerts` said so in a comment: the row is claimed, sent for evaluation, fails the same way, then is reread and discarded every minute, forever. The operator saw a perpetual "Failing" with no signal that any amount of waiting would fix it. This adds `PARKED`, a terminal status that the evaluator alone can reach when the failure is a verdict about the stored rule, not about the run.

## Type of Change

- [x] feat - A new feature

## Details

- `ALERT_STATUSES` gains `PARKED`, which leaves the claim query the way `PAUSED` already does (the `status = 'ACTIVE'` guard was already there, so no new query, index, or migration; `alerts.status` is a plain `VARCHAR`).
- A rule parks on three failures, all verdicts about the stored rule rather than the run: `parseAlertRule` returning null (parked in `claimRow`, after the claim, so `nextRunAt` still advances), a filter spec the backend would refuse over the whole batch (`isSendableAlertSpec`), and a backend `WidgetSpecError` (unknown measure, unsupported view, a forbidden aggregation, filters on a traces-routed measure).
- Telling a spec rejection apart from a transient query failure used to mean string-sniffing the message. `AlertEvaluationResult` now carries `error_kind` (`"spec"` or `"query"`), set on the two existing `except` branches in `alert_evaluation.py`; the worker parks only on `"spec"`. An unlabelled failure, from a backend older than this change, is retried and never parked, and a park write that itself fails falls back to recording the failure, so a rule is never silently stopped. A batch-level `WidgetSpecError` (a bad window) still raises for the whole request and stays a retry, since that's a request fault, not one rule's.
- `PARKED` is not client-settable: `ALERT_STATUSES` carries it, but the pause endpoint and the status mutation take `SETTABLE_ALERT_STATUSES` (`ACTIVE` or `PAUSED`) only.
- Leaving `PARKED` is a cold start, two ways: editing the rule (the PATCH route re-arms it when the edit rewrites the evaluated rule or writes a fresh `renotify`; a name-only edit leaves it parked, since it changes nothing the evaluator refused), or resuming from the row menu as the operator's explicit retry. If the settings are still unevaluable, the next tick parks it again with the reason.
- UI: a distinct "Parked" badge outranking Paused, dimmed row, menu offering Resume like a paused rule. The badge states the stored reason once and trusts it, since both real parking paths (`claim.ts`'s `UNEVALUABLE_RULE_ERROR`, `scheduler.ts`'s `PARKED_RULE_SUFFIX`) already say what unparks the rule; the badge does not append its own "Edit and save..." on top of that (that fallback sentence is kept only for the one case with no stored reason at all).

Before -
tr-before

After -
tr-after

## Checklist

- [ ] I have updated documentation where necessary. N/A, no doc-facing behaviour change

---
## Summary by cubic
Adds a terminal `PARKED` status for alert rules that can never be evaluated, replacing the perpetual retry that showed a frozen "Failing" state. The evaluator now parks a rule when the failure is a verdict about the stored spec (`error_kind: "spec"`), not the run (`"query"`).

- A rule parks when parsing fails, the spec is unsendable, or a `WidgetSpecError` rejects it; the reason is stored on the row.
- Unlabelled failures from older backends are still retried, and a failed park write falls back to recording the failure.
- `PARKED` is not client-settable; an edit re-arms the rule when it rewrites the evaluated rule or `renotify` (a name-only edit keeps it parked), and the Resume action starts it again — the re-arm is status-CAS'd so a tick parking the row mid-edit is still caught.
- UI adds a "Parked" badge that outranks Paused, a dimmed row, and Resume for both stopped statuses; the badge trusts the stored reason and does not repeat the fix sentence unless no reason is stored.

**Migration**

- No schema or query changes; `alerts.status` is a plain `VARCHAR`.

<sup>Written for commit edfa67d82045a08db1e9135cadd710af1df2f456. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/2122?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


GabrielDrapor:
Closing in favour of the original #2122 by @Prasad8830, which is reopened against `main` now that its base branch could be restored. The merge of `main` on this branch (`32f00a21`, conflicts resolved, CI green) stays on `origin/restore/feat/alerts-parked-status` for that PR to pull in.

PR #2155, opened 2026-09-09
fix(models): add GPT-6 Astra pricing and BYOK support

### LINKED ISSUE #2127: `gpt-6-astra` is unpriced and absent from every surface
## Summary

`gpt-6-astra` — the first GPT-6 model — appears in no surface of the product. A span recorded against it resolves to no price and renders as costless, and the model cannot be selected anywhere in the UI. Unlike the GPT-5.6 gaps, the rates here **are** published, so this is a straight missing-entry rather than an unknown-price problem.

Published rates, per 1M tokens: **input $10, cached input $1, output $50, cache writes $12.50**. Context window 1,050,000; max output 128,000; knowledge cutoff 2026-04-30; text + image in, text out.

## How to replicate

Resolve `gpt-6-astra` against the catalogue patterns: no entry matches, so `getModelPricing` returns `null` and the cost column reads as absent rather than as an error. The model also does not appear in the BYOK dropdown, the system-model list, or either docs page. `git grep -inE "gpt-6"` over `frontend`, `backend`, `docs` and `tests` returns nothing.

## Root cause (confirmed)

The model is simply not in `frontend/packages/core/src/standard-model-prices.json`, so there is no pattern to match and no price to return. It was published after the catalogue was last updated by hand, and nothing reconciles the catalogue against the provider's published model list — the mechanism described in #1597.

A missing entry fails quietly by design: an unmatched model returns `null` rather than raising, so nothing surfaces the gap.

## Surfaces

| Surface | State |
|---|---|
| `standard-model-prices.json` | **absent** — no price, 92 entries and none for GPT-6 |
| `llm-providers.ts` → `ADAPTER_MODELS` (BYOK dropdown) | **absent** — the OpenAI list stops at the `gpt-5.6-*` trio (`:159-161`) |
| `llm-providers.ts` → `SYSTEM_MODELS` | **absent** — **not necessarily a gap**: system models run on TraceRoot's own API keys, so listing a frontier model here means TraceRoot pays its inference. Out of scope for this issue. |
| `docs/ai-agent/byok.mdx` | **absent** — the OpenAI row (`:13`) stops at `gpt-5.5` |
| `docs/integrations/internal-models.mdx` | **absent** — `:25` still calls `gpt-5.5` "OpenAI's most capable model", now two generations stale |
| `test_pricing.py`, `model-resolver.test.ts` | no resolution case, no price assertion |

## Impact

- Any usage of this model is recorded at no cost, so spend on it is invisible in per-trace, per-eval and usage reporting. That matters more here than for the earlier gaps: at **$50 / 1M output** this is by a wide margin the most expensive model the catalogue would carry, so the unrecorded spend per span is the largest of any missing entry to date.
- The model cannot be chosen through BYOK, so the product understates what it supports by a full generation.
- The `// Limited partner preview as of 2026-07-09 — not yet accessible via self-serve API keys` comment at `llm-providers.ts:158` is now stale twice over. Already noted in #2043.

## Acceptance criteria

- A catalogue entry exists with `input: 1e-05`, `output: 5e-05`, `cacheRead: 1e-06`, `cacheWrite: 1.25e-05`, matching the published $10 / $50 / $1 / $12.50.
- The pattern covers the bare id and the `openai/` and `azure/` prefixed forms, consistent with the neighbouring entries: `(?i)^(openai\/|azure\/)?(gpt-6-astra)(-[\d-]+)?$`.
- A resolution test covers the bare and prefixed forms.
- **A test asserts the base prices against their published values.** This is the load-bearing item. The published cache rates happen to sit at exactly the ratios the existing suite pins — `cacheRead` at 0.10x input, `cacheWrite` at 1.25x input — so a ratio-only test would pass against a wrong base price, which is precisely how the GPT-5.6 drift in #2043 went unnoticed.
- The model is added to the BYOK dropdown (`ADAPTER_MODELS`). **Do not add it to either docs table**: `byok.mdx`'s Default Models table is machine-locked to `SYSTEM_MODELS` by `frontend/packages/core/src/__tests__/llm-providers.test.ts:170-193` and adding a BYOK-only model there will fail CI, and `internal-models.mdx` documents models available "out of the box — no API key required", which a BYOK-only model is not. Document it in prose under the BYOK section instead, or record in the PR body that it is deliberately absent from both system-model tables.
- Separately worth fixing, but not on this issue: `internal-models.mdx:25` still calls `gpt-5.5` "OpenAI's most capable model", an absolute claim that is now stale.
- **`SYSTEM_MODELS` is deliberately out of scope** — that list runs on TraceRoot's keys and adding a frontier model to it is a cost decision, not a bug fix.

Source: https://developers.openai.com/api/docs/models/gpt-6-astra

superezzdev:
Fixes #2127 

## Summary

Add GPT-6 Astra pricing and BYOK model support.

GPT-6 Astra was missing from the standard pricing catalogue, causing usage to resolve without a price. It was also missing from the OpenAI BYOK model dropdown.

This change adds the published GPT-6 Astra pricing, model resolution coverage, explicit absolute-price assertions, and BYOK model selection support.

## Type of Change

* [ ] feat - A new feature
* [x] fix - A bug fix
* [ ] docs - Documentation only changes
* [ ] style - Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc)
* [ ] refactor - A code change that neither fixes a bug nor adds a feature
* [ ] perf - A code change that improves performance
* [ ] test - Adding missing tests or correcting existing tests
* [ ] build - Changes that affect the build system or external dependencies (example scopes: gulp, broccoli, npm)
* [ ] ci - Changes to our Cl configuration files and scripts (example scopes: Travis, Circle, BrowserStack, SauceLabs)
* [ ] chore - Other changes that don't modify sc or test files
* [ ] revert - Reverts a previous commit
* [ ] security - A security fix or improvement
* [ ] github - Changes to our GitHub configuration files and scripts
* [ ] other (please describe):

## Details

* Added `gpt-6-astra` to `standard-model-prices.json`.
* Added support for:

  * `gpt-6-astra`
  * `openai/gpt-6-astra`
  * `azure/gpt-6-astra`
  * version-suffixed model IDs.
* Added GPT-6 Astra to `ADAPTER_MODELS` for OpenAI BYOK.
* Added model resolution test coverage.
* Added explicit assertions for the published absolute rates:

  * Input: `$10 / 1M`
  * Cached input: `$1 / 1M`
  * Cache writes: `$12.50 / 1M`
  * Output: `$50 / 1M`
* Left `SYSTEM_MODELS` unchanged because GPT-6 Astra is BYOK-only.
* Left the machine-locked documentation tables unchanged.

Validation:

* `uv run pytest tests/worker/tokens/test_pricing.py` — **238 passed**
* `pnpm exec vitest run packages/core/src/__tests__/llm-providers.test.ts` — **14 passed**

Not applicable.

## Checklist

* [x]  I have read the [CONTRIBUTING.md](../CONTRIBUTING.md)
* [x] I have added/updated tests where applicable
* [ ] I have added screenshots or recordings for UI changes (if applicable)
* [x] I have updated documentation where necessary

---
## Summary by cubic
Adds GPT-6 Astra pricing and BYOK model support so usage resolves to the correct prices and the model appears in OpenAI BYOK dropdowns.

- Adds `gpt-6-astra` pricing entry at published rates: $10 / 1M input, $1 / 1M cache read, $12.50 / 1M cache write, $50 / 1M output.
- Extends model matching to handle `openai/`, `azure/`, and version-suffixed IDs.
- Registers `gpt-6-astra` in `ADAPTER_MODELS` for OpenAI BYOK.
- Adds resolution tests and absolute-price assertions for all rate tiers.

<sup>Written for commit 92f9f03dea95835b1ca449d8ce1a805beb0fcf3a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/2155?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/packages/core/src/llm-providers.ts
+++ b/frontend/packages/core/src/llm-providers.ts
@@ -162,6 +162,7 @@ export const ADAPTER_MODELS: Partial<Record<LLMAdapter, LLMModelDef[]>> = {
     { id: "gpt-5.6-terra", label: "gpt-5.6-terra" },
     { id: "gpt-5.6-luna", label: "gpt-5.6-luna" },
     { id: "gpt-5.6-cyber", label: "gpt-5.6-cyber" },
+    { id: "gpt-6-astra", label: "gpt-6-astra" },
     { id: "gpt-5.5", label: "gpt-5.5" },
     { id: "gpt-5.5-pro", label: "gpt-5.5-pro" },
     { id: "gpt-5.4", label: "gpt-5.4" },
--- a/frontend/packages/core/src/model-resolver.ts
+++ b/frontend/packages/core/src/model-resolver.ts
@@ -125,15 +125,13 @@ export function resolvePiModel(
         | Record<string, string>
         | undefined;
 
-      // BYOK with no model id: use the adapter's default model rather than
-      // relying on catalog ordering. The OpenAI catalog is newest-first, so
-      // its first entry may be a preview model that isn't broadly available.
-      // For free-text adapters (azure, amazon-bedrock, openrouter) the catalog
-      // is empty — caller MUST provide modelId, else we'd silently send a
-      // wrong model name to that provider.
+      // BYOK with no model id: use the adapter's first curated model so we
+      // don't accidentally fall back to an Anthropic default for an OpenAI-
+      // compatible BYOK row. For free-text adapters (azure, amazon-bedrock,
+      // openrouter) the catalog is empty — caller MUST provide modelId, else
+      // we'd silently send a wrong model name to that provider.
       const catalog = ADAPTER_MODELS[providerConfig.adapter as keyof typeof ADAPTER_MODELS];
-      const fallbackModelId =
-        modelId || (providerConfig.adapter === "openai" ? "gpt-5.6-sol" : catalog?.[0]?.id);
+      const fallbackModelId = modelId || catalog?.[0]?.id;
       if (!fallbackModelId) {
         throw new Error(
           `BYOK adapter "${providerConfig.adapter}" has no curated model catalog ` +
--- a/frontend/packages/core/src/model-resolver.ts
+++ b/frontend/packages/core/src/model-resolver.ts
@@ -125,13 +125,15 @@ export function resolvePiModel(
         | Record<string, string>
         | undefined;
 
-      // BYOK with no model id: use the adapter's first curated model so we
-      // don't accidentally fall back to an Anthropic default for an OpenAI-
-      // compatible BYOK row. For free-text adapters (azure, amazon-bedrock,
-      // openrouter) the catalog is empty — caller MUST provide modelId, else
-      // we'd silently send a wrong model name to that provider.
+      // BYOK with no model id: use the adapter's default model rather than
+      // relying on catalog ordering. The OpenAI catalog is newest-first, so
+      // its first entry may be a preview model that isn't broadly available.
+      // For free-text adapters (azure, amazon-bedrock, openrouter) the catalog
+      // is empty — caller MUST provide modelId, else we'd silently send a
+      // wrong model name to that provider.
       const catalog = ADAPTER_MODELS[providerConfig.adapter as keyof typeof ADAPTER_MODELS];
-      const fallbackModelId = modelId || catalog?.[0]?.id;
+      const fallbackModelId =
+        modelId || (providerConfig.adapter === "openai" ? "gpt-5.6-sol" : catalog?.[0]?.id);
       if (!fallbackModelId) {
         throw new Error(
           `BYOK adapter "${providerConfig.adapter}" has no curated model catalog ` +
--- a/tests/worker/tokens/test_pricing.py
+++ b/tests/worker/tokens/test_pricing.py
@@ -362,6 +362,12 @@ class TestGpt56LunaPublishedPrices:
         entry = next(e for e in real_cache if e["model_name"] == "gpt-5.6-luna")
         assert entry["prices"]["input"] == pytest.approx(2e-7)  # $0.20 / 1M tokens
         assert entry["prices"]["output"] == pytest.approx(1.2e-6)  # $1.20 / 1M tokens
+
+
+class TestGpt6AstraPublishedPrices:
+    """Assert the absolute, provider-published rate directly."""
+
+    def test_published_rates(self, real_cache):
         entry = next(e for e in real_cache if e["model_name"] == "gpt-6-astra")
         assert entry["prices"]["input"] == pytest.approx(1e-05)  # $10 / 1M tokens
         assert entry["prices"]["output"] == pytest.approx(5e-05)  # $50 / 1M tokens
--- a/tests/worker/tokens/test_pricing.py
+++ b/tests/worker/tokens/test_pricing.py
@@ -95,6 +95,10 @@ OPENAI_MODEL_CASES = [
     ("gpt-5.6-cyber", "gpt-5.6-cyber"),
     ("openai/gpt-5.6-cyber", "gpt-5.6-cyber"),
     ("azure/gpt-5.6-cyber", "gpt-5.6-cyber"),
+    ("gpt-6-astra", "gpt-6-astra"),
+    ("openai/gpt-6-astra", "gpt-6-astra"),
+    ("azure/gpt-6-astra", "gpt-6-astra"),
+    ("gpt-6-astra-2026-07-09", "gpt-6-astra"),
     ("gpt-5.5", "gpt-5.5"),
     ("openai/gpt-5.5", "gpt-5.5"),
     ("azure/gpt-5.5", "gpt-5.5"),
@@ -358,6 +362,11 @@ class TestGpt56LunaPublishedPrices:
         entry = next(e for e in real_cache if e["model_name"] == "gpt-5.6-luna")
         assert entry["prices"]["input"] == pytest.approx(2e-7)  # $0.20 / 1M tokens
         assert entry["prices"]["output"] == pytest.approx(1.2e-6)  # $1.20 / 1M tokens
+        entry = next(e for e in real_cache if e["model_name"] == "gpt-6-astra")
+        assert entry["prices"]["input"] == pytest.approx(1e-05)  # $10 / 1M tokens
+        assert entry["prices"]["output"] == pytest.approx(5e-05)  # $50 / 1M tokens
+        assert entry["prices"]["cacheRead"] == pytest.approx(1e-06)  # $1 / 1M tokens
+        assert entry["prices"]["cacheWrite"] == pytest.approx(1.25e-05)  # $12.50 / 1M tokens
 
 
 class TestGeminiModelIds:
--- a/frontend/packages/core/src/__tests__/model-resolver.test.ts
+++ b/frontend/packages/core/src/__tests__/model-resolver.test.ts
@@ -92,6 +92,21 @@ describe("resolvePiModel", () => {
     expect(m.id).not.toMatch(/claude/i);
   });
 
+  it("does not use the preview model as the OpenAI BYOK fallback", () => {
+    const cfg: ProviderModelConfig = {
+      adapter: "openai",
+      key: "sk-test",
+      baseUrl: null,
+      config: null,
+    };
+
+    const m = resolvePiModel(undefined, cfg);
+
+    expect(m.provider).toBe("openai");
+    expect(m.id).toBe("gpt-5.6-sol");
+    expect(m.id).not.toBe("gpt-6-astra");
+  });
+
   it("throws (instead of substituting an Anthropic default) for free-text adapter with no model id", () => {
     // azure, amazon-bedrock, openrouter aren't in ADAPTER_MODELS; without explicit
     // modelId the resolver would have nothing sensible to use, so it must throw
--- a/frontend/packages/core/src/__tests__/model-resolver.test.ts
+++ b/frontend/packages/core/src/__tests__/model-resolver.test.ts
@@ -92,21 +92,6 @@ describe("resolvePiModel", () => {
     expect(m.id).not.toMatch(/claude/i);
   });
 
-  it("does not use the preview model as the OpenAI BYOK fallback", () => {
-    const cfg: ProviderModelConfig = {
-      adapter: "openai",
-      key: "sk-test",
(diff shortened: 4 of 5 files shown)


frontend/packages/core/src/llm-providers.ts
```diff
@@ -159,6 +159,7 @@ export const PROVIDER_PRIORITY: LLMAdapter[] = [
 export const ADAPTER_MODELS: Partial<Record<LLMAdapter, LLMModelDef[]>> = {
   openai: [
     // Limited partner preview as of 2026-07-09 — not yet accessible via self-serve API keys.
+    { id: "gpt-6-astra", label: "gpt-6-astra" },
```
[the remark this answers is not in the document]
superezzdev: Fixed in 410c2f13.

I moved `gpt-6-astra` below the existing GPT-5.6 entries in `ADAPTER_MODELS`, so it remains selectable through BYOK without becoming the fallback model when no model ID is specified.

`SYSTEM_MODELS` and the resolver logic were intentionally left unchanged.

tests/worker/tokens/test_pricing.py
```diff
@@ -71,6 +71,10 @@ def real_cache() -> list[dict]:
     ("gpt-5.6-luna", "gpt-5.6-luna"),
     ("openai/gpt-5.6-luna", "gpt-5.6-luna"),
     ("azure/gpt-5.6-luna", "gpt-5.6-luna"),
+    ("gpt-6-astra", "gpt-6-astra"),
+    ("openai/gpt-6-astra", "gpt-6-astra"),
+    ("azure/gpt-6-astra", "gpt-6-astra"),
+    ("gpt-5.6-sol", "gpt-5.6-sol"),
```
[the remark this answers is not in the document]
superezzdev: Fixed in 410c2f13.

Replaced the duplicate GPT-5.6 test case with:

`gpt-6-astra-2026-07-09`

so the test now explicitly covers the version-suffixed regex matching behavior.

tests/worker/tokens/test_pricing.py
```diff
@@ -267,6 +271,11 @@ def test_input_and_output_match_published_rate(self, real_cache):
         entry = next(e for e in real_cache if e["model_name"] == "gpt-5.6-luna")
         assert entry["prices"]["input"] == pytest.approx(2e-7)  # $0.20 / 1M tokens
         assert entry["prices"]["output"] == pytest.approx(1.2e-6)  # $1.20 / 1M tokens
+        entry = next(e for e in real_cache if e["model_name"] == "gpt-6-astra")
```
dark-sorceror, line 280: These sit inside luna's test, so astra drift reports as a luna failure and a luna failure stops them running at all. Move them to their own `TestGpt6AstraPublishedPrices`.

frontend/packages/core/src/llm-providers.ts
```diff
@@ -162,6 +162,7 @@ export const ADAPTER_MODELS: Partial<Record<LLMAdapter, LLMModelDef[]>> = {
     { id: "gpt-5.6-sol", label: "gpt-5.6-sol" },
     { id: "gpt-5.6-terra", label: "gpt-5.6-terra" },
     { id: "gpt-5.6-luna", label: "gpt-5.6-luna" },
+    { id: "gpt-6-astra", label: "gpt-6-astra" },
```
dark-sorceror: The list is newest-first everywhere else. Move it above `gpt-5.6-sol`, or note why it sits here.

dark-sorceror (changes requested):
Prices match the provider's page. Two below, and it needs a rebase since #2053 merged.

superezzdev:
Thanks for the review! Addressed the requested changes and the additional BYOK fallback issue.

- Moved the Astra published-price assertions into `TestGpt6AstraPublishedPrices`.
- Moved `gpt-6-astra` above `gpt-5.6-sol` to preserve newest-first ordering.
- Decoupled the OpenAI BYOK fallback from `ADAPTER_MODELS[0]`, so the preview Astra model cannot become the implicit fallback.
- Added regression coverage for the OpenAI BYOK fallback.
- Updated the branch against the latest `main`.

Validation:
- `uv run pytest tests/worker/tokens/test_pricing.py` — 239 passed
- `model-resolver.test.ts` + `llm-providers.test.ts` — 35 passed
- Pre-commit hooks — all passed

frontend/packages/core/src/llm-providers.ts
dark-sorceror: Put this back below the 5.6 entries. `catalog[0]` is the BYOK fallback, so your first version was right, and a one-line comment saying so would stop the next person moving it.

frontend/packages/core/src/model-resolver.ts
```diff
@@ -125,13 +125,15 @@ export function resolvePiModel(
...
+      // BYOK with no model id: use the adapter's default model rather than
+      // relying on catalog ordering. The OpenAI catalog is newest-first, so
+      // its first entry may be a preview model that isn't broadly available.
+      // For free-text adapters (azure, amazon-bedrock, openrouter) the catalog
+      // is empty — caller MUST provide modelId, else we'd silently send a
+      // wrong model name to that provider.
       const catalog = ADAPTER_MODELS[providerConfig.adapter as keyof typeof ADAPTER_MODELS];
-      const fallbackModelId = modelId || catalog?.[0]?.id;
+      const fallbackModelId =
+        modelId || (providerConfig.adapter === "openai" ? "gpt-5.6-sol" : catalog?.[0]?.id);
```
dark-sorceror: With astra back below the 5.6 entries this hardcode is not needed, since `catalog[0]` is already `gpt-5.6-sol`. Two places deciding the default can drift apart.

dark-sorceror (commented):
My ordering comment was wrong, apologies. `catalog[0]` is the OpenAI BYOK fallback, so your original placement below the 5.6 entries was deliberate and correct, and moving it is what created the preview-model fallback cubic then flagged. The simplest resolution is to revert both changes it caused.

The separate class for the astra assertions is exactly right.

This also needs a rebase, since main has moved.

tests/worker/tokens/test_pricing.py
dark-sorceror: Fixed. `TestGpt6AstraPublishedPrices` now holds these, so an astra drift reports under its own name instead of luna's, and a luna failure no longer short-circuits them.

frontend/packages/core/src/llm-providers.ts
dark-sorceror: Retracting this one. It was wrong, and it is what set off the fallback problem below.

`resolvePiModel` uses `catalog?.[0]?.id` for a BYOK row with no model id, so position 0 in this list is the OpenAI default, not just presentation order. Your original placement below the 5.6 entries was correct and deliberate. Sorry for the detour.

frontend/packages/core/src/model-resolver.ts
dark-sorceror: Done, the revert restored `catalog?.[0]?.id`. The other half of the pair is still outstanding, see the thread on the model list.

frontend/packages/core/src/llm-providers.ts
dark-sorceror: Still open. The revert took the resolver half but not this half, so `gpt-6-astra` is `catalog[0]` again and is the OpenAI BYOK fallback with nothing guarding it now. Moving it back below `gpt-5.6-luna` closes both this and the automated P2 above.

superezzdev:
Thanks for the clarification.
@dark-sorceror 
I've restored `gpt-6-astra` below the GPT-5.6 entries so `catalog[0]` remains `gpt-5.6-sol`, while keeping the separate `TestGpt6AstraPublishedPrices` test class.

All requested changes are now addressed, and the relevant tests and pre-commit checks are passing.

frontend/packages/core/src/llm-providers.ts
dark-sorceror: Fixed in 8e4c0f9. `gpt-5.6-sol` is `catalog[0]` again so the BYOK fallback is a generally available model, and the resolver is back to plain `catalog?.[0]?.id` with no hardcoded id, which was the right call over keeping two places that decide the default.

dark-sorceror:
This needs a rebase. Five PRs landed on main since it was opened, and all three of your files moved: #2051, #2053 and #2131, then #1342 and #1557.

Two things to watch, both of which bit other PRs this week. In `standard-model-prices.json` the conflict boundary cuts through the middle of an object, so keeping both sides yields a duplicated key and a missing `},{`. Committed invalid JSON there takes the whole worker suite down at collection rather than failing one test. In `test_pricing.py` several PRs appended a provider test class at the same anchor, so git interleaves them line by line instead of giving a clean two-block conflict; keep each class whole and separate.

One conflict resolves in your favour: the `Limited partner preview` comment above the OpenAI list was deleted on main as stale, so take main's side there.

superezzdev:
@dark-sorceror
 Rebased onto the latest `main` and resolved the requested conflicts while preserving the current OpenAI BYOK ordering and GPT-6 Astra changes.

 All checks are now passing, including pricing, model resolution, TypeScript build/lint, backend/frontend tests, and security review.

 The final diff remains limited to the pricing catalogue, BYOK model list, and pricing tests.

 Ready for another review. Thanks!

dark-sorceror (approved):
Rebase is clean. Verified on the merged tree: JSON parses at 101 entries with no duplicate `id`, `modelName` or `matchPattern`, 314 tests pass, and all 16 test classes are uniquely named so nothing got interleaved at the shared anchor.

`gpt-6-astra` sits below the 5.6 family, so `gpt-5.6-sol` is still `catalog[0]` and the BYOK fallback stays a generally available model. Taking main's side on the preview comment was right, it was deleted there as stale.

All four rates still match the published $10 / $50 / $1 / $12.50, and the pattern covers the bare, `openai/` and `azure/` forms.

Worth noting the long-context tier, where prompts past the threshold bill at 2x input and 1.5x output for the whole request. The catalogue has no way to express a tiered rate, so this is a standing gap across every entry rather than anything this PR introduces.

PR #2161, opened 2026-09-10
fix(pricing): price the six unpriced codex SKUs

### LINKED ISSUE #1545: Cost reads `null` for gpt-5.1-codex-mini — missing entry in standard-model-prices.json (6 of 7 codex SKUs unpriced)

Mohil-Ahuja:
Closes #1545.

## What was wrong

`gpt-5.3-codex` was the only codex SKU in `standard-model-prices.json`. Every other one matched no entry, so `get_model_price` returned `None`, `cost_from_buckets` returned `None`, and the span stored `cost: null`. On the dashboard that renders as no cost at all, which reads as a broken cost column rather than a missing catalogue row.

Checked against the OpenAI provider table that ships in `@earendil-works/pi-ai`, the same source the issue cites:

| SKU | Priced before | Priced now |
|---|---|---|
| `gpt-5-codex` | no | yes |
| `gpt-5.1-codex` | no | yes |
| `gpt-5.1-codex-max` | no | yes |
| `gpt-5.1-codex-mini` | no | yes |
| `gpt-5.2-codex` | no | yes |
| `gpt-5.3-codex` | yes | unchanged |
| `gpt-5.3-codex-spark` | no | yes |

## Why the suffixes each need their own entry

The catalogue convention ends a pattern with `(-[\d-]+)?$`, which absorbs a dated revision such as `gpt-5.1-codex-2026-01-20`. It does not reach `-max`, `-mini` or `-spark`, since those are not digits. That is why the existing `gpt-5.3-codex` pattern never caught its own `-spark` sibling, and it is why each SKU gets a row rather than one widened pattern.

Widening the pattern instead would be actively wrong: `gpt-5.1-codex-mini` bills at a fifth of `gpt-5.1-codex`, so folding the family into one entry would overcharge every mini span by 5x.

## Rates

Taken from the provider table rather than inferred from siblings, as USD per token:

| SKU | input | output | cacheRead |
|---|---|---|---|
| `gpt-5-codex` | 1.25e-6 | 1e-5 | 1.25e-7 |
| `gpt-5.1-codex` | 1.25e-6 | 1e-5 | 1.25e-7 |
| `gpt-5.1-codex-max` | 1.25e-6 | 1e-5 | 1.25e-7 |
| `gpt-5.1-codex-mini` | 2.5e-7 | 2e-6 | 2.5e-8 |
| `gpt-5.2-codex` | 1.75e-6 | 1.4e-5 | 1.75e-7 |
| `gpt-5.3-codex-spark` | 1.75e-6 | 1.4e-5 | 1.75e-7 |

The already-present `gpt-5.3-codex` row matches that table exactly, which is a useful check that it is the right source. `cacheWrite` is `null` throughout, matching every other OpenAI entry.

## Tests

`TestCodexModelIds` in `tests/worker/tokens/test_pricing.py`, built on the existing `real_cache` fixture so it reads the shipped catalogue rather than a mock:

- every SKU resolves, bare and `openai/`-prefixed, and dated revisions still fold into the base SKU
- each id matches **exactly one** entry. The names nest, `gpt-5.1-codex` is a prefix of `gpt-5.1-codex-max`, so a pattern that lost its anchor would price two models at once and the winner would depend on catalogue order
- absolute published rates, not ratios. Sibling SKUs share a rate card, so a row copied from the wrong sibling stays internally consistent and would pass a ratio check
- `gpt-5.1-codex-mini` is a fifth of `gpt-5.1-codex`, guarding the 5x overcharge above
- the exact id from the issue, `gpt-5.1-codex-mini`, priced end to end through `calculate_cost`

30 of the 35 new cases fail against the catalogue as it stands today, verified by reverting the JSON and re-running. 277 tests pass in the pricing module; prettier and ruff clean.

---
## Summary by cubic
Fixes #1545 by adding price entries for the six unpriced Codex SKUs. Previously only `gpt-5.3-codex` was catalogued, so the other six SKUs matched no pattern, `get_model_price` returned `None`, and spans recorded `cost: null`, which rendered as a broken cost column on the dashboard.

- Each SKU gets its own entry because the pattern ending `(-[\d-]+)?$` only absorbs dated revisions, not `-max`, `-mini`, or `-spark` suffixes; widening the pattern would overcharge `gpt-5.1-codex-mini` by 5x.
- Rates come from the OpenAI provider table in `@earendil-works/pi-ai`, and the existing `gpt-5.3-codex` row matches that table exactly.
- Adds `TestCodexModelIds` covering SKU resolution, exact one-entry matches, absolute published rates, and the mini-to-family 5x ratio.

<sup>Written for commit 926b52c06acbfea7c6b5b64730162b629a2e8dd3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/2161?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/packages/core/src/standard-model-prices.json
+++ b/frontend/packages/core/src/standard-model-prices.json
@@ -811,6 +811,78 @@
     }
   },
   {
+    "id": "tr-gpt-5-3-codex-spark",
+    "modelName": "gpt-5.3-codex-spark",
+    "matchPattern": "(?i)^(openai\\/)?(gpt-5\\.3-codex-spark)(-[\\d-]+)?$",
+    "provider": "openai",
+    "prices": {
+      "input": 0.00000175,
+      "output": 0.000014,
+      "cacheRead": 0.000000175,
+      "cacheWrite": null
+    }
+  },
+  {
+    "id": "tr-gpt-5-2-codex",
+    "modelName": "gpt-5.2-codex",
+    "matchPattern": "(?i)^(openai\\/)?(gpt-5\\.2-codex)(-[\\d-]+)?$",
+    "provider": "openai",
+    "prices": {
+      "input": 0.00000175,
+      "output": 0.000014,
+      "cacheRead": 0.000000175,
+      "cacheWrite": null
+    }
+  },
+  {
+    "id": "tr-gpt-5-1-codex-max",
+    "modelName": "gpt-5.1-codex-max",
+    "matchPattern": "(?i)^(openai\\/)?(gpt-5\\.1-codex-max)(-[\\d-]+)?$",
+    "provider": "openai",
+    "prices": {
+      "input": 0.00000125,
+      "output": 0.00001,
+      "cacheRead": 0.000000125,
+      "cacheWrite": null
+    }
+  },
+  {
+    "id": "tr-gpt-5-1-codex-mini",
+    "modelName": "gpt-5.1-codex-mini",
+    "matchPattern": "(?i)^(openai\\/)?(gpt-5\\.1-codex-mini)(-[\\d-]+)?$",
+    "provider": "openai",
+    "prices": {
+      "input": 0.00000025,
+      "output": 0.000002,
+      "cacheRead": 0.000000025,
+      "cacheWrite": null
+    }
+  },
+  {
+    "id": "tr-gpt-5-1-codex",
+    "modelName": "gpt-5.1-codex",
+    "matchPattern": "(?i)^(openai\\/)?(gpt-5\\.1-codex)(-[\\d-]+)?$",
+    "provider": "openai",
+    "prices": {
+      "input": 0.00000125,
+      "output": 0.00001,
+      "cacheRead": 0.000000125,
+      "cacheWrite": null
+    }
+  },
+  {
+    "id": "tr-gpt-5-codex",
+    "modelName": "gpt-5-codex",
+    "matchPattern": "(?i)^(openai\\/)?(gpt-5-codex)(-[\\d-]+)?$",
+    "provider": "openai",
+    "prices": {
+      "input": 0.00000125,
+      "output": 0.00001,
+      "cacheRead": 0.000000125,
+      "cacheWrite": null
+    }
+  },
+  {
     "id": "tr-gemini-3-5-flash",
     "modelName": "gemini-3.5-flash",
     "matchPattern": "(?i)^(google\\/|models\\/)?gemini-3\\.5-flash(-[\\d-]+)?$",
--- a/tests/worker/tokens/test_pricing.py
+++ b/tests/worker/tokens/test_pricing.py
@@ -185,6 +185,38 @@ KIMI_MODEL_CASES = [
     ("kimi-k3-20260716", "kimi-k3"),
 ]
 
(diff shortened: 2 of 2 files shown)


Mohil-Ahuja:
Closing this as superseded by #1557, and with a correction on my side.

I diffed the two before closing. For all six SKUs both of us price, the rates and the match patterns are byte-identical, so nothing of mine is missing from main.

The one difference was that my branch also priced `gpt-5.3-codex-spark` at $1.75/$14, and main deliberately does not, with `test_unpriced_codex_preview_does_not_inherit_another_model_price` asserting it resolves to None. Main is right and I was wrong. OpenAI never published a public API rate for spark; the $1.75/$14 that third-party trackers show for it is just the full `gpt-5.3-codex` rate copied across, which is precisely the inheritance that test exists to prevent. Pricing it would have invented a number rather than recorded one. The model was retired the week of 14 September in any case.

So there is nothing to carry over and nothing to reopen. Nice catch on the "only officially priced" trim.

PR #2163, opened 2026-09-10
fix(slack)-disclose omitted alert filters in trace links

### LINKED ISSUE #2144: Alert Slack message lists filters its "View traces" link cannot carry; span-grain drill-down is the real fix
## What a reader sees

An alert on `count > 3` filtered by `is_root = true` (or `name contains e2e`) posts:

> `count` was 10, above the 3 threshold, over the last 5m. Where `is_root = true`.
> View alert · **View traces**

Clicking **View traces** opens the trace list with only the evaluated window applied. The filter shown one line above is not applied, and nothing in the message says so. Observed on staging (2026-09-07, rules R22 and R26b of the alerts e2e run); the R26b link is exactly `/projects/<id>/traces?date_filter=custom&start=…&end=…`.

## Why

By design from #1995 (`alertFiltersToTracePredicates`, `packages/core/src/alerts.ts`): the prose lists **every** filter, but the link carries only what the trace list can express:

- `=` on `model_name` / `environment` / `status` / `span_kind` / `name` → `in`
- `metadata[key]` with `=` or `contains`

Everything else stays prose-only: `contains` on a categorical field (the list only takes `in`), and `is_root` (no list column). #1995's description states this under "Semantics, stated honestly" and names a span-grain drill-down as the follow-up, but no issue tracks it.

The underlying mismatch is that alerts evaluate at span grain while the trace list filters at trace grain ("trace has ≥1 span where …"), so even the filters that do map are looser than what the rule counted.

## Two levels of fix

**1. Small, do first: tell the reader when the link is narrower than the rule.**
When `alertFiltersToTracePredicates` drops any filter, mark the link, e.g. `View traces (window only)` or `View traces (without name contains …)`, or append one sentence: "The link applies the window and `model_name = gpt-4o`; `is_root = true` is not applied." A message that lists a condition next to a link that ignores it is worse than one that says so. Owner layer: `packages/slack/src/alert-blocks.ts` (#1891 / #1995).

**2. The real fix: a span-grain drill-down.**
A "View spans" target that runs the alert's own widget query (`backend/rest/services/alert_evaluation.py` measure source, same predicates, same window) and lists the counted spans. Closes the gap for `contains`, `is_root`, and the ≥1-span widening on the fields that do map today. Likely a new list surface or a mode of the traces page; proposed in #1995 as a separate feature.

Related: #2141 (number formatting in the same message), #2143 (rule edits silently close pages).

Co-Authored-By: Claude <noreply@anthropic.com>

https://claude.ai/code/session_01S81TcLQgDRyYG8QMNorqtB

D4rsh11:
## What

Disclose when an alert's "View traces" link cannot carry all of the rule's filters.

## Why

Some alert filters cannot be represented by the trace list. Previously, those filters were silently omitted from the link, which could make the linked trace results look broader than the alert condition.

The link now shows "View traces (some filters not applied)" when applicable.

## Validation

- `pnpm --dir frontend --filter @traceroot/slack test`
- `pnpm --dir frontend --filter @traceroot/slack build`
- `git diff --check`
- `pre-commit run --files frontend/packages/slack/src/alert-blocks.ts frontend/packages/slack/src/__tests__/alert-blocks.test.ts`

Closes #2144

---
## Summary by cubic
Fixes the Slack "View traces" link so it discloses when some alert filters can't be represented in the trace list. Previously those filters were silently dropped, making the linked results look broader than the alert condition; the link now reads "View traces (some filters not applied)" in that case. Closes #2144.

<sup>Written for commit d6a4ba6382d9eb22804aea73daf3340be5c8cdbf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/2163?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/packages/slack/src/__tests__/alert-blocks.test.ts
+++ b/frontend/packages/slack/src/__tests__/alert-blocks.test.ts
@@ -272,7 +272,8 @@ describe("buildAlertBlocks", () => {
     // the fallback carries the clause too, for clients that render no blocks
     expect(filtered.text).toContain("Where `span_kind = LLM` and");
 
-    const url = links.match(/<([^|]+)\|View traces>/)![1];
+    expect(links).toContain("View traces (some filters not applied)");
+    const url = links.match(/<([^|]+)\|View traces \(some filters not applied\)>/)![1];
     const filtersParam = new URL(url).searchParams.get("filters");
     expect(JSON.parse(filtersParam!)).toEqual([
       { field: "span_kind", op: "in", value: ["LLM"] },
--- a/frontend/packages/slack/src/alert-blocks.ts
+++ b/frontend/packages/slack/src/alert-blocks.ts
@@ -4,6 +4,7 @@ import {
   ALERT_THRESHOLD_OPERATOR_PHRASES,
   alertFiltersToTracePredicates,
   describeAlertFilter,
+  isCompleteAlertFilter,
   type AlertFilter,
   type AlertSeverity,
   type AlertThresholdOperator,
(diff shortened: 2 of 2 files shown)


GabrielDrapor:
Could you please add before-and-after screenshots of the Slack alert message? Thanks!

PR #2228, opened 2026-09-17
feat(ui): validate a detector's model against the workspace's model list

ka1kqi:
## Summary

The models a workspace can run a detector on now come from one shared helper in `@traceroot/core`: the system catalog for every provider the process holds an API key for, plus the workspace's enabled BYOK providers and the models configured on each. Three surfaces use it:

- The web app's `GET /api/workspaces/[workspaceId]/llm-models` route returns exactly that list (behaviour unchanged, the inline code moved into the helper).
- The detector write service (`createDetector`, used by the public API and the agent) checks `detection_model` / `detection_provider` / `detection_source` against it and returns a **400** whose message lists what the workspace can use.
- The cookie `POST .../detectors` route makes the same check, so both write surfaces reject the same payloads.

The detector worker already fails a run on an unknown system model; this moves that failure to the write, so a caller that guessed an id (an agent, a script) learns the valid ids in the response instead of from a failed evaluation later.

Validation rules: a system choice (the default source) may leave the model unset for the default, and a set model must be in the system list. A BYOK choice must name an enabled provider in the workspace (by its name, which is what the worker looks providers up by) and a model configured on it. The DB read is skipped for the plain default choice.

## Follow-ups

- Detector updates get the same check once the update services land (the edit-access stack); the helper is built to be called from there.

## Test plan

- [x] core: `workspace-models.test.ts` (env filtering, BYOK mapping and `supported` flag, every validation branch)
- [x] ui: `detectors.test.ts` (default choice skips the read, list read through the transaction, 400s for unknown system model / missing provider / unconfigured BYOK model, valid BYOK stored), cookie route tests, `llm-models` route test
- [x] `tsc`, eslint and prettier clean for core and ui; diff coverage 98% locally

https://claude.ai/code/session_01JddAMkfJ7P4HbtVZpkkTj9

---
## Summary by cubic
Detector creates now validate their model choice against the workspace's available models. An unknown model fails the create with a 400 that lists the workspace's models, instead of failing the detector's first run.

- The models a workspace can run come from one shared helper in `@traceroot/core`: the system catalog for each provider the process holds a key for, plus the workspace's enabled BYOK providers and their configured models.
- The web app's `GET /api/workspaces/[workspaceId]/llm-models` route returns that same list; behavior is unchanged.
- Both `createDetector` and the cookie `POST .../detectors` route run the same check and return the same messages.
- Validation: a system choice may leave the model unset for the default, and a set model must be in the system list. A BYOK choice must name an enabled provider and a model configured on it. The DB read is skipped for the plain default choice.
- Detector updates get the same check once the update services land.

<sup>Written for commit d8917006c8fa762bc01a449183fda69acdbaf0ab. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/2228?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


XinweiHe (approved):
LGTM

PR #2240, opened 2026-09-18
fix(billing): meter only completed detector scans in usage details

### LINKED ISSUE #2033: billing: detector-scan quota counts failed judge runs (uniqExact(run_id) ignores status)
## What happens

`backend/rest/routers/internal/usage.py:137-149` (`/api/v1/internal/usage/details`):

```sql
SELECT uniqExact(run_id) AS total
FROM detector_runs
WHERE project_id IN {project_ids} AND timestamp >= {start} AND timestamp < {end}
```

No `status` predicate. `detector_runs` records a row for every attempt, including `status='failed'` runs where **no LLM call was made** (e.g. "No API key configured for provider", provider timeout before the request, missing model). The billing cron uses this number as `detector.scansRun` → `isDetectorRunBlocked` (Free 100-scan hard cap) and the paid `DETECTOR_HOSTED_LLM_FREE_THRESHOLD` split.

## Effect

Observed locally: 267 consecutive `failed` runs from a missing provider key — zero inference — would have consumed a Free workspace's 100-scan cap 2.6× over. The token-cost side is correct (failed scans write no `ai_messages` row), only the count is wrong; but the count is what blocks Free workspaces and what apportions the paid hosted-LLM overage (`detectorBillableCost = cost × overageScans / scansRun` — a wrong denominator under-bills paid customers).

## Why this is a bug, not a decision

The comment above the query says "count every scan attempt … BYOK + system source both count toward Free-plan hard cap" — it addresses BYOK vs system, not success vs failure. `0cf14755` describes the meter as detector *scans*; a run that never reached the model is not a scan. Same class of gap as #2032.

## Fix

Add `AND status = 'completed'` to that query (the runs table has `status String`; the worker writes `completed | failed`). Keep the comment's BYOK/system statement — it remains true. The guard test at `tests/rest/test_detector_endpoints.py:1094-1097` asserts only that no `source` token appears; a `status` predicate does not conflict with it.

## Acceptance

- `/usage/details` `detector_runs` excludes `failed` rows; a test in `TestUsageDetails` asserts `"status = 'completed'"` in `runs_sql`.
- A Free workspace with 100 `failed` and 10 `completed` runs is **not** detector-blocked.
- `usageMetering` detector overage math uses the completed count.

Related: #2031, #2032 (same audit). Found while designing agent self-tracing; deliberately kept out of that design.

Abhishek-B-R:
Fixes #2033.

`/usage/details` counted every row in `detector_runs`, with no `status` predicate. The table records one row per attempt, and a `status='failed'` run never reached the model: a missing provider key, or a provider timeout before the request goes out, writes a row and costs no inference.

## Why `status = 'completed'` is the right predicate

Checked rather than assumed. The worker's writer types it as a closed pair:

```ts
// frontend/worker/src/detection/clickhouse-writer.ts
status: "completed" | "failed";
```

and all four `writeDetectorRun` call sites in `detector-run-processor.ts` pass one of those two. There is no `running` or `pending` row in this table, so the predicate cannot hide an in-flight scan.

## Effect

The token-cost side was already correct, because a failed scan writes no `ai_messages` row. Only the count was wrong, and the count is what gates and apportions:

- `detector.scansRun` feeds `isDetectorRunBlocked`, the Free 100-scan hard cap. A workspace whose provider key is missing burns its whole month's allowance on runs that did no work.
- `detectorBillableCost = cost * overageScans / scansRun`. An inflated denominator shrinks the billable share, so paid hosted-LLM overage is under-billed.

## Scope

One query. `/usage/total` counts traces and spans only and never touched `detector_runs`, so nothing else needed changing.

The source-agnostic guarantee is untouched: no predicate on `source` is added, BYOK and system runs both still count, and the existing `assert "source" not in runs_sql` still passes. The comment above the query spoke to BYOK versus system, not success versus failure, so it has been extended rather than contradicted.

## Tests

`test_usage_details_meters_only_completed_detector_runs` added to `TestUsageBillsEveryStoredRow`, asserting the predicate reaches `runs_sql` and that no `source` predicate came with it.

I checked it actually catches the bug: with the one added line reverted, the new test fails; with it in place, it passes.

`tests/rest/test_detector_endpoints.py` passes, 76 tests. Full `tests/rest/` is 1490 passed, 4 skipped, with one failure in `test_source_consumers.py::test_every_spans_or_traces_reader_is_classified` that is present on a clean checkout of `main` without this change.

`ruff check` and `ruff format --check` clean.

---
## Summary by cubic
Fixes `/usage/details` so it counts only `completed` detector runs. Previously every row in `detector_runs` was counted, including `failed` attempts that never reached the model, which inflated the Free-plan hard cap and under-billed paid overage. This addresses issue #2033.

The worker writes one row per attempt with status `'completed'` or `'failed'`, so the new predicate does not hide in-flight scans. No other query is affected, and the source-agnostic behavior (BYOK and system runs both count) is unchanged. Added a test to assert the predicate and ensure no `source` filter is introduced.

<sup>Written for commit ea174a99a887cb65f69f26b3fcfd17727e243f2f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/2240?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Abhishek-B-R:
Closing as a duplicate of #2162, which covers this and is more thorough: it also handles why row-level filtering is safe against the ReplacingMergeTree merge state, and it catches a second class of failed run that did consume inference, which I had missed. Sorry for the noise. I checked assignees and comments but not the issue's linked pull requests.

PR #2241, opened 2026-09-18
fix(evaluations): gate the date range by the plan's retention window

### LINKED ISSUE #1987: Evaluations date range is not plan-gated, bypassing retention entitlement
## Summary

The Evaluations page date-range dropdown lets any user pick 30d / 60d / 90d regardless of subscription plan, bypassing the retention gate every other trace-telemetry surface enforces. The page renders the shared `DateFilterSelect` **without** the `retentionDays` prop that drives the lock, so the lock can never fire — and the evaluations backend applies no retention clamp either, so wider ranges genuinely return data rather than being merely cosmetically selectable.

This is a wiring defect, not an unbuilt feature: the gate exists, works, and is applied on Traces, Sessions, Users, Detectors and Dashboards. Evaluations is the one surface never passed the plan state.

## How to replicate

1. Sign in as a workspace on **Free** (retention cap 15 days) or **Starter** (30 days).
2. Open Traces and its date-range dropdown: presets beyond the plan window render with a lock icon and an "N-day retention limit · click to upgrade" tooltip, and clicking routes to the upgrade flow.
3. Open the **Evaluations** page and its date-range dropdown.
4. **Observed:** every preset including 30d/60d/90d is freely selectable — no lock, no tooltip, no upgrade routing, on any plan.
5. Select 90d and confirm the runs list actually returns rows older than the plan's retention window. The wider range is honored end-to-end, not just displayed.

## Root cause (confirmed)

The lock is entirely prop-driven and Evaluations never supplies the prop; the backend has no clamp either, so nothing enforces the cap at either layer.

1. **The lock returns "unlocked" when `retentionDays` is absent.** `isOptionLocked()` short-circuits to `false` for *every* option when it is null/undefined (`frontend/ui/src/lib/date-filter.ts:38-45`, specifically `:42`). It is consumed once per option at `frontend/ui/src/components/date-filter-select.tsx:88`; a locked option renders the lock icon (`:121`) and routes the click to `onUpgradeClick` instead of applying the range (`:100-121`). With no prop, `locked` is always `false` and the gate is inert.
2. **The preset list reaches 90d.** `DATE_FILTER_OPTIONS` includes 30d, 60d, 90d (`date-filter.ts:17-32`), and the comment at `:26-28` states 60d/90d are *meant* to render locked on lower tiers.
3. **Evaluations passes no plan state.** `frontend/ui/src/features/evaluations/views/evaluations-view.tsx:312-321` passes only `dateFilter` / `customStartDate` / `customEndDate` / `onDateFilterChange` / `onCustomRangeChange` — no `retentionDays`, no `onUpgradeClick` — and never imports `useRetention`.
4. **Every other surface wires it.** Traces `app/projects/[projectId]/traces/page.tsx:208`, Sessions `:132`, Users `:133`, Detectors list `:154`, Detector detail `:232`, Dashboard `:157` all pass `retentionDays={retention.retentionDays}`; Traces additionally wires `onUpgradeClick` (`traces/page.tsx:208-209`).
5. **The caps are real and fail closed.** `RETENTION_DAYS` (`frontend/packages/core/src/ee/billing/plans.ts:151-171`) is Free **15**, Starter **30**, Pro **90**, Enterprise **null**; `getRetentionDays()` falls back to `FAIL_CLOSED_RETENTION_DAYS = 15` (`:159`) for unknown plans. `frontend/ui/src/lib/hooks/use-retention.ts` reads `workspace.billingPlan`, defaults to FREE, and returns `undefined` while loading — so nothing locks transiently on surfaces that do wire it.

**Not merely cosmetic:** retention is also enforced server-side, so on gated surfaces a hand-crafted wide request is still clamped. `backend/rest/retention.py` (`clamp_retention_window` pulls `start_after` forward; `enforce_retention_by_time` 403s by-id lookups) is wired into `backend/rest/routers/traces.py:93,207,267,340,362` plus sessions, users, dashboards, detectors and live SSE, and mirrored in `frontend/ui/src/lib/server/retention.ts`. The evaluations path has **none** of it: `frontend/ui/src/app/api/projects/[projectId]/evaluations/runs/route.ts` contains no retention or clamp handling. Note this Next route is the **only** server-side reader of evaluation runs — the SDK-facing gateway `backend/rest/routers/public/eval.py` is write-only for runs (its `_UPSTREAM_ROUTES` allowlist registers `evaluation-runs` as POST-only; its sole GETs are `datasets` / `dataset-versions`, which take no time window), so a `GET` for runs is refused at routing and never forwarded. One clamp, in the Next route, is therefore sufficient.

**Secondary follow-up (not part of this fix):** the dashboard widget-builder *preview* window is also ungated — `frontend/ui/src/features/dashboards/components/WidgetBuilderPage.tsx:296` uses raw `RANGE_PRESETS` (`features/dashboards/range-presets.ts:11`) with no retention filter. It is a preview control, not the live dashboard query (whose own filter *is* gated at `dashboard/[dashboardId]/page.tsx:157`), so impact is limited.

## Impact

- **Retention cap not enforced:** workspaces on lower-retention plans can query evaluation runs across the full 90-day window, in the UI *and* against the backend.
- **Inconsistent behaviour:** the same control shows locks and upgrade prompts on every other surface but is wide open here, so the retention limit reads as arbitrary.
- **No backstop:** because the eval path applies no clamp, nothing server-side catches the ungated UI — the wider data is actually served.

## Acceptance criteria

- The Evaluations `DateFilterSelect` receives `retentionDays` and `onUpgradeClick` from `useRetention`, exactly like the other six surfaces, so presets beyond the plan window render locked and route to upgrade.
- On Free, 30d/60d/90d lock; on Starter, 60d/90d lock; on Pro/Enterprise all presets stay selectable — matching `RETENTION_DAYS`.
- The evaluations runs endpoint enforces the same server-side clamp as the telemetry routes, reusing the existing helpers rather than a parallel mechanism — so a wider range is clamped rather than served even against a hand-crafted request. This means `.../evaluations/runs/route.ts`, the only reader; `public/eval.py` is write-only for runs and has nothing to clamp, which should be pinned by a test so a future ungated read cannot be added silently.
- An **omitted or unparseable** `started_after` must be clamped to the cutoff too — an unbounded request must not be a wider query than explicitly asking for 90 days.
- Nothing locks transiently while plan state is loading (`retentionDays` `undefined` → no locks), consistent with the other surfaces.
- Plan retention numbers are unchanged and the fail-closed default for unknown plans is preserved.
- (Secondary) the widget-builder preset preview is either gated the same way or has a documented reason for exemption.

Abhishek-B-R:
Fixes #1987.

The Evaluations date-range dropdown offered 30d/60d/90d on every plan, and the wider range was genuinely served rather than merely selectable. Both halves of the gate were missing, which is why nothing caught it.

## Why it was invisible

The lock is entirely prop-driven, and it fails open. `isOptionLocked` returns `false` for every option when `retentionDays` is absent, so a surface that omits the prop is not partially gated, it is completely ungated: no lock icon, no tooltip, no error, and the query runs. There was no server-side clamp underneath either, so nothing caught the ungated picker.

## The fix

**Client.** `evaluations-view` now takes `useRetention(projectId)` and passes `retentionDays` and `onUpgradeClick`, plus the `PricingDialog` that the upgrade click opens. That is the same wiring the list surfaces use, reached through `search-filter-bar`.

**Server.** The runs route clamps `started_after` through the existing `clampStartAfter` helper, the one the detector routes already use, rather than a parallel mechanism. Workspace plan is read the same way the detector runs route reads it.

## One deliberate behaviour change

An **absent or unparseable** `started_after` is now clamped to the cutoff rather than dropped. Unbounded is the widest possible query, not a narrow one, so dropping the bound was the wrong default and it made the hand-crafted request wider than explicitly asking for 90 days.

This does change the existing `drops an unparseable started_after instead of erroring` case, but only on capped plans. On unlimited retention the value still passes through untouched, so that test and the other eleven hold unchanged. I defaulted the test workspace to a plan with unlimited retention precisely so the existing cases keep testing windowing and derivation rather than quietly becoming retention tests.

## Tests

`runs/route.test.ts`, six added, using the real `RETENTION_DAYS` rather than numbers restated in the test file, so changing a plan's window cannot leave these asserting the old one:

- an out-of-window `started_after` is pulled forward to the cutoff
- an omitted one is clamped too
- an unparseable one is clamped on a capped plan
- an in-window one is left alone
- unlimited retention applies no cutoff
- an unrecognized plan string fails closed to the most restrictive window

Verified they catch the bug: with the clamp reverted, four of the six fail. The other two assert passthrough, so they correctly pass either way.

`test_public_eval_gateway_security.py`, one added. The issue notes that one clamp in the Next route is sufficient only because the SDK gateway is write-only for runs. That is now pinned: the test asserts no `evaluation-runs` route in `_UPSTREAM_ROUTES` allows `GET`, and that the routes which do allow `GET` are exactly the four that take no time window. A future ungated read cannot be added silently.

`date-filter-retention-guard.test.ts`, new. The bug class is "a surface renders the picker without the prop", so this reads the source and requires every `DateFilterSelect` call site to pass `retentionDays` or appear on an explicit exemption list with a reason. A render test would only cover the pages someone remembered to write a test for, which is the same gap again. Verified: reverting the view wiring fails this guard.

## Two things found while fixing, not fixed here

**`features/alerts/components/alert-preview.tsx` is also ungated.** It renders the picker with no `retentionDays` and drives a real preview query from the chosen range. I have left it alone and put it on the guard's exemption list with its reason, because it is a preview control inside the rule form rather than a telemetry list surface, which is the same call the issue makes for the dashboard widget-builder preview. Happy to gate it here instead if you would rather, or to open a separate issue. Flagging it rather than deciding it silently.

**One pre-existing test failure, unrelated.** `src/ee/features/billing/usage-labels.test.ts > formats with locale-string commas` fails on a clean checkout of `main` without this change. `formatDetectorScanLabel` uses `toLocaleString()` with no fixed locale, so on an Indian-locale machine 1234567 groups as `12,34,567` rather than `1,234,567`. Worth knowing because it means no contributor in an en-IN locale can get a green local run; pinning the locale in the assertion would fix it. Not touched here.

## Checks

`tsc --noEmit` clean. `eslint` and `prettier --check` clean on the changed files. `ruff check` and `ruff format --check` clean.

`vitest run` in `frontend/ui`: 3231 passed, 1 failed (the locale test above). `tests/rest/test_public_eval_gateway_security.py`: 27 passed.

---
## Summary by cubic
Fixes #1987: the Evaluations runs date picker offered 30d/60d/90d on every plan and served the wider range, unlike every other telemetry surface. The view now passes retention state to the picker and the runs route clamps `started_after` via the existing `clampStartAfter` helper.

**Bug Fixes**
- Evaluations view wires `useRetention` and `PricingDialog` the same way the other list surfaces do.
- Runs route clamps out-of-window, omitted, and unparseable `started_after` to the plan cutoff; unrecognized plans fail closed to the most restrictive window.
- Absent or unparseable bounds are now clamped instead of dropped on capped plans, while unlimited-retention plans pass through unchanged.
- Six new route tests use the real `RETENTION_DAYS`, a gateway test pins evaluation runs as write-only, and a source guard requires every `DateFilterSelect` call site to pass `retentionDays` or sit on a documented exemption list.
- `alert-preview.tsx` is exempt with its reason, and one pre-existing `usage-labels.test.ts` failure on Indian-locale machines is unrelated.

<sup>Written for commit f394b1e150e53ab9ea94f1bfdcb915af789aa701. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/2241?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Abhishek-B-R:
Closing as a duplicate of #1991, which has been open since 24 August with the same approach and more besides. Sorry for the noise, my pre-flight check read PR bodies for closing keywords and so missed a PR linked through the Development sidebar.

Two things from mine that are not in #1991, in case either is useful there rather than lost:

- `features/alerts/components/alert-preview.tsx` also renders DateFilterSelect with no `retentionDays`, and drives a real preview query from the chosen range. Arguably the same category as the widget-builder preview #1987 sets aside, but it is a second ungated call site.
- `src/ee/features/billing/usage-labels.test.ts` > 'formats with locale-string commas' fails on any machine with an Indian locale, because `formatDetectorScanLabel` calls `toLocaleString()` with no fixed locale, so 1234567 groups as 12,34,567. It fails on a clean main, so no contributor in an en-IN locale gets a green local run. Pinning the locale in the assertion fixes it. Happy to send that one line as its own PR if you want it.

Issue #2253, opened 2026-09-21
Jev as a detector judge: TypeSafe AI as a BYOK provider for detectors


lqiu03:
## Summary

Let a workspace bring its own **TypeSafe AI** key and use **Jev** to judge detectors instead of an LLM. Jev is a decision model: it answers typed questions with calibrated probabilities instead of writing text. A detector judged by Jev reports a **category, the full probability distribution and a confidence** (for example `tool_error P=0.91, next: timeout 0.06`), not just yes or no. That gives RCA a concrete hypothesis to test, at about $0.00007 per judgement.

Jev is **detector-only**. It can't write text, so it's kept out of chat, RCA and digests. Nothing changes for workspaces that don't add TypeSafe.

## Model reference

| Attribute | Value |
|---|---|
| Provider / adapter | TypeSafe AI / `typesafe` |
| Model ID | `jev-1.13.0` (pinned; `jev-latest` isn't used, so runs are reproducible) |
| Endpoint | `POST https://api.typesafe.ai/v1/systemone`, with `{state, model, questions}` |
| Price | $0.042 per million input tokens; output is free |
| Context | 64k tokens per request |

## How a Jev detector decides

- **Gate (Noul):** "Does the trace show any problem described in `detector_criteria`?" It fires at P ≥ 0.5.
- **Label (Choice):** hidden categories for each built-in template (failure, hallucination, logic, task, safety), plus `other` and `none`. A blank detector gets `{problem, none}`.
- Both questions go in one request. The trace is projected into a compact state under a fixed character budget, and the detector prompt is passed as a value, never interpolated into the question.

## Isolation guarantees

- `ModelSelector` hides decision models unless `allowDecisionModels` is set. Only the detector pickers set it.
- `resolvePiModel` throws for a decision adapter or model ID instead of falling back to another model, which covers the agent, RCA and digest.
- `projects.ts` refuses Jev as the RCA model or provider when the setting is saved.
- Chat readiness ignores workspaces that only have TypeSafe.
- Jev inference is reachable only through `detector-run-processor` → `sandbox-eval` (one dispatch branch) → `jev-eval` → `callSystemOne`.

## Notes

- **Additive only.** The LLM detector path, templates, `outputSchema`, sampling and RCA prompts are unchanged. The edits to existing files are mostly inserted lines.
- **Deferred:** an "inconclusive" run status, per-detector thresholds (0.5 is a constant), a category editor, and a TypeSafe-native digest or RCA.
[...]

PR #2256, opened 2026-09-21
feat(worker): project a trace into Jev state and compile the judge questions

lqiu03:
**Stacked PR 3 of 4**: Jev as a detector judge. Base: `feat/jev-systemone-client`. Part of #2253.

Turns a detector's inputs into what Jev reads. The trace becomes a compact **state**, and the detector's prompt becomes two **questions**. Nothing calls these yet; PR 4 wires them in.

### What's included
- `detection/jev-state.ts`: parses `spansJsonl`, keeps the meaningful columns (dropping ids, timestamps, cost and token counts), and parses JSON-string payloads. It reduces the result deterministically to a fixed character budget in this order:
  1. truncate inputs;
  2. drop metadata;
  3. truncate outputs, keeping head and tail;
  4. drop middle spans, leaving a marker.

  A trace that still doesn't fit fails the run with a clear message.
- `detection/jev-templates.ts`: hidden categories for each built-in template (failure, hallucination, logic, task, safety). Each has a *what* and a *not for*. A blank detector gets `problem`.
- `detection/jev-compile.ts`: builds a gate **Noul** ("does the trace show any problem in `detector_criteria`?") and a label **Choice** over the categories plus `other` and `none`. The user's prompt is passed as a value and never interpolated into the question text.

### Test plan
- [x] State shape, column projection, JSON-string parsing, skipped and blank lines
- [x] Reduction order: inputs, then metadata, then output head and tail, then middle spans; over-budget failure
- [x] Blank, null and unknown templates fall back to `{problem, none}`; inherited object keys never resolve as templates
- [x] The gate and label shapes, with the prompt under `detector_criteria` and categories ordered as template, then `other`, then `none`

https://claude.ai/code/session_0165UTzNkPQ4bTseiEXuUDM2

---
## Summary by cubic
Prepares a detector's inputs for Jev by projecting the trace into a compact state and compiling the detector prompt into two judge questions. Nothing calls these yet; the next PR in the stack wires them in.

**State and reduction**
- Reduces a trace deterministically to a fixed budget: truncate inputs, then drop metadata, then truncate outputs keeping head and tail, then drop middle spans behind a marker; states still over budget fail the run.
- Keeps only meaningful columns and parses JSON-string payloads so Jev reads real structure instead of escaped text.

**Questions and categories**
- Builds a gate Noul decision and a label Choice over hidden categories for each built-in template; blank or unknown templates fall back to a generic `{problem, none}` pair.
- The detector prompt is passed as the value of `detector_criteria` and never interpolated into question text.

<sup>Written for commit 0035c7a31a2b026d5d9bfef3e8f3bd309897918e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/2256?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


frontend/worker/src/detection/__tests__/jev-state.test.ts
```diff
@@ -0,0 +1,225 @@
...
+    expect(stats.input_cap_chars).toBe(8_000);
+    expect(stats.truncated_inputs).toBe(1);
+    expect(stats.dropped_metadata).toBe(0);
+    expect(stats.omitted_spans).toBe(0);
+    expect(spans[0].input).toBe(`${"x".repeat(8_000)}…[truncated 12000 chars]`);
+    expect(spans[1].input).toBe("short");
+    // Outputs and metadata are untouched at this stage.
+    expect(spans[0].output).toBe("It is sunny in Paris.");
+    expect(spans[0].metadata).toEqual({ user_id: "u-1" });
+    expect(stats.final_chars).toBe(JSON.stringify(state).length);
```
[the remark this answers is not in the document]
lqiu03: Good catch, fixed in baa1b098. I dropped the two self-comparing `final_chars === JSON.stringify(state).length` assertions and added the missing `<= budget` bound to the metadata-drop test. The span-dropping test keeps its equality check, because there it does verify the incremental size arithmetic.

frontend/worker/src/detection/jev-state.ts
```diff
@@ -0,0 +1,221 @@
...
+function truncate(value: JsonValue, cap: number, keepTail: boolean): JsonValue {
+  const text = typeof value === "string" ? value : JSON.stringify(value);
+  if (text.length <= cap) return value;
+  const marker = `…[truncated ${text.length - cap} chars]`;
+  if (!keepTail) return `${text.slice(0, cap)}${marker}`;
+  const tail = Math.floor(cap / 2);
+  return `${text.slice(0, cap - tail)}${marker}…${text.slice(text.length - tail)}`;
+}
+
+function applyReductions(spans: SpanState[], step: Step): Reduction {
```
GabrielDrapor, line 110: The reduction only ever touches `input`, `output` and the metadata columns. `status_message`, `name` and `model_name` are never capped, so a span with a very long `status_message` (a stack-trace dump, for example) can keep the state over budget through every step, and the run then fails with the message at the bottom saying inputs, metadata and outputs were reduced and only the first span kept, which overstates what was tried. Capping `status_message` the same way as `input` (or including it in the input cap step) would close it.

frontend/worker/src/detection/__tests__/jev-state.test.ts
```diff
@@ -0,0 +1,224 @@
...
+    );
+    const { state, stats } = buildJevState(jsonl(rows), { budgetChars: 2_500 });
+    const trace = traceOf(state);
+    expect(trace.span_count).toBe(10);
+    const marker = trace.spans.find((s) => "omitted_spans" in s);
+    expect(marker).toEqual({ omitted_spans: stats.omitted_spans });
+    expect(stats.omitted_spans).toBeGreaterThan(0);
+    expect(stats.kept_spans + stats.omitted_spans).toBe(10);
+    const names = trace.spans.filter((s) => s !== marker).map((s) => s.name);
+    const head = Math.ceil(stats.kept_spans / 2);
```
GabrielDrapor: `head`/`tail` here are derived from the function's own `stats.kept_spans`, and the `kept_spans + omitted_spans === 10` check above holds by construction (`kept_spans` is defined as `spanCount - omittedSpans`), so an off-by-one in the keep loop bound would still pass this test. One hard-coded expectation would pin it, e.g. a budget sized so that exactly `spanCount - 1` spans fit and asserting `omitted_spans === 1` and the exact names kept.

GabrielDrapor (commented):
One robustness point and one test point, inline.

frontend/worker/src/detection/jev-state.ts
lqiu03: 4878befc caps `status_message` at the same `inputCap`, in the same step, with its own `truncated_status_messages` stat. Test: a 50,000-char status_message now fits a 12,000-char budget at the loosest cap.

frontend/worker/src/detection/__tests__/jev-state.test.ts
lqiu03: 72dbc8a7 rewrites it: ten uniform spans, budget 3,200, asserting `kept_spans 9`, `omitted_spans 1` and the exact kept names. Mutation-checked — starting the keep loop one span early now fails.

XinweiHe (approved):

PR #2257, opened 2026-09-21
feat(worker): judge detectors with Jev when the BYOK provider is TypeSafe

lqiu03:
**Stacked PR 4 of 4**: Jev as a detector judge. Base: `feat/jev-state-questions`. Part of #2253.

Turns the feature on. When a detector's BYOK provider is TypeSafe, Jev judges the trace instead of an LLM. The finding carries a **category, the full probability distribution and a confidence**, not just yes or no, so RCA gets a hypothesis to test.

### What's included
- `detection/jev-eval.ts`: `runJevDetection` builds the state and questions, calls System One inside the span, and maps the answers:
  - it **fires** when the gate is ≥ 0.5;
  - the category is the most likely option other than `none`, or "category unclear" when `none` wins;
  - the summary reads `Jev: tool_error (P=0.91, label confidence 0.88); next: timeout (0.06)`.

  Usage is recorded as BYOK `typesafe` and priced through the existing `calculateCost` fallback.
- `detection/sandbox-eval.ts`: **one branch**, right after the BYOK provider lookup, entered only for a decision adapter. Failures come back as main's `errorResult`. Nothing else in the LLM path changes.
- `detector-run-processor.ts`: passes the already-loaded `detector.template` through (3 inserted lines).
- The new-detector page and the detector panel pass `allowDecisionModels`, so `jev-1.13.0 (TypeSafe AI)` appears only in detector pickers.

### Test plan
- [x] Gate threshold (0.5 fires, 0.49 doesn't); `none` → unclear category; `other` → unlisted problem; runner-up in the summary
- [x] A Jev failure becomes an error EvalResult carrying the thrown message
- [x] TypeSafe BYOK never calls `resolvePiModel` or the LLM `complete`
- [x] Detector pickers list decision models

### Verified end to end (local `make dev-autoreload`)
| Trace | Jev |
|---|---|
| 401 token expired, same call retried 3×, empty answer | 🔴 loop (P=0.81) |
| 1 of 3 emails bounced, agent says all sent | 🔴 swallowed_error (P=0.77) |
| Order lookup 404, agent invents a delivery date | 🔴 swallowed_error (P=0.77) |
| Booking tool rejects args, agent says "booked" | 🔴 swallowed_error (P=0.93) |
| Clean currency conversion | ✅ no finding |

Each run wrote one BYOK `typesafe` usage row with about 1.6–2k input tokens, costing about **$0.00007 per judgement**.

https://claude.ai/code/session_0165UTzNkPQ4bTseiEXuUDM2

---
## Summary by cubic
Turns on Jev as the detector judge when a detector's BYOK provider is TypeSafe. Detectors with a decision adapter are now judged by System One instead of a chat LLM, and the finding carries a category, the full probability distribution, and a confidence so RCA gets a hypothesis to test.

**What changed**
- `sandbox-eval.ts` routes TypeSafe BYOK detectors to the new `runJevDetection`; the chat-path branch is untouched.
- `jev-eval.ts` builds the Jev state and questions, calls System One, and maps answers: fires when gate ≥ 0.5, reports the top non-`none` category (or "category unclear" when `none` wins), and includes the runner-up in the summary.
- Jev writes no rationale, so the summary says so; failures come back as an error `EvalResult`.
- Usage is recorded as BYOK `typesafe` and priced through the existing `calculateCost` fallback (about $0.00007 per judgement in local tests).
- Detector pickers now pass `allowDecisionModels` so Jev models (e.g. `jev-1.13.0`) appear only in detector pickers.
- `detector-run-processor.ts` passes `detector.template` through so the correct question set is compiled.

<sup>Written for commit b7cd1301b25c8ed08f22204f3a193be92dc6e545. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/traceroot-ai/traceroot/pull/2257?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


frontend/worker/src/detection/jev-eval.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  const gate = result.answers.gate.noul;
+  const label = result.answers.label;
+  const identified = gate >= JEV_GATE_THRESHOLD;
+
+  const ranked = rankCategories(label.probabilities);
+  const top = ranked[0];
+  const next = ranked[1];
+  // The gate fired but the label picked none: name no category.
+  const unclear = label.choice === JEV_NONE;
```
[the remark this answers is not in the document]
lqiu03: `label.choice` is Jev's own pick, and `label.confidence` is the confidence in that pick, so "unclear" follows Jev's decision rather than overriding it with our own argmax. When the named category differs from Jev's pick, the summary already leaves out the confidence (line 82). Keeping as is.

frontend/worker/src/detection/jev-eval.ts
```diff
@@ -0,0 +1,110 @@
...
+  const confidence =
+    top.name === label.choice ? `, label confidence ${label.confidence.toFixed(2)}` : "";
+
+  const summary = !identified
+    ? `Jev: no problem found (P=${gate.toFixed(2)} that the trace exhibits one).`
+    : unclear
+      ? `Jev: problem found (P=${gate.toFixed(2)}), category unclear` +
+        "; no written rationale — verify against the trace."
+      : `Jev: ${displayCategory(top.name)} (P=${top.p.toFixed(2)}${confidence})` +
```
GabrielDrapor: This summary is rendered verbatim in the findings table, next to the LLM detectors' plain sentences, e.g.

`Jev: tool_error (P=0.91, label confidence 0.88); next: timeout (0.06); no written rationale — verify against the trace.`

At first glance it reads like a debug line (model name, `P=` notation, an instruction to the reader). Suggest a plain headline in `summary`, something like "Tool error (91% likely); runner-up: timeout (6%)", and keep the full distribution and confidence in `data` for the detail view, which it already carries.

frontend/worker/src/detection/sandbox-eval.ts
```diff
@@ -160,6 +162,19 @@ export async function runDetectionForTrace(params: {
...
+    if (isDecisionAdapter(providerConfig.adapter)) {
+      try {
+        return await runJevDetection({
+          spansJsonl,
+          detector,
+          providerConfig,
+          timeoutMs: parseDetectorEvalTimeoutMs(process.env.DETECTOR_EVAL_TIMEOUT_MS),
+        });
+      } catch (err) {
+        return errorResult(err instanceof Error ? err.message : String(err), source);
```
GabrielDrapor: If TypeSafe answers 200 but the answer fails our own validation, `parseResult` throws before it reads `usage`, and this catch returns `errorResult` with zero tokens, although TypeSafe processed and billed that call. The LLM path in this file forwards the accumulated tokens into its final `errorResult` (line 323), so failures still write a usage row. Reading `usage` before validating `answers` in `typesafe-client.ts` (attaching it to the thrown error, or returning it separately) and passing it through here would keep Jev consistent with that.

GabrielDrapor (commented):
Two points on the Jev result path, inline.

frontend/worker/src/detection/jev-eval.ts
lqiu03: 9d9c2c0f: "Tool error (91% likely); runner-up: timeout (6%); no written rationale." (73 chars, inside the table's 100-char cut). Category ids become words automatically, no label map; data.category and the distribution are unchanged.

frontend/worker/src/detection/sandbox-eval.ts
lqiu03: 5098223e does that, with the client half in e2dc695d: usage is parsed before the answers and attached to the thrown error, and the dispatch passes tokens, model and provider into `errorResult` like the LLM path. Test asserts 810/63 on a malformed 200.

XinweiHe (approved):

frontend/worker/src/detection/sandbox-eval.ts
```diff
@@ -160,6 +167,31 @@ export async function runDetectionForTrace(params: {
...
+          spansJsonl,
+          detector,
+          providerConfig,
+          timeoutMs: parseDetectorEvalTimeoutMs(process.env.DETECTOR_EVAL_TIMEOUT_MS),
+        });
+      } catch (err) {
+        // TypeSafe bills a 200 whose answers we then reject, and the thrown
+        // error carries those tokens. Attribute them as the LLM path does, so
+        // the billed call still gets a usage row instead of being dropped.
+        const usage = usageFromError(err);
```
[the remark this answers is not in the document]
lqiu03: Correct, and my earlier reply was wrong: `e2dc695d` moved usage above the per-answer loop but left the top-level `answers` presence check above it, so that one shape still threw unbilled. Fixed in `12269fbe` on #2255, which owns the file — the check now sits at the top of the `try`, where the `catch` attaches the billed usage. The "answers missing" test only asserted the message, which is how it survived; the new case asserts `usageFromError` returns the tokens. #2256 and #2257 rebased onto it (`25ab5305`, `a0725978`).

